PR #17, opened 2026-07-18
[fix](execution) Preserve ExecutionBatch payloads across retries

### LINKED ISSUE #14: [P0][Bug] ExecutionBatch loses payload when a blocking sink is retried
## Severity

**P0 / release blocker**

## Summary

The new `ExecutionBatch` execution path can silently truncate local streaming query results when a result sink returns `SinkResultType::BLOCKED`.

The default `PhysicalOperator::SinkBatch` fallback destructively moves `batch.materialized` into a temporary `DataChunk` before the sink has accepted it. If `PhysicalBufferedCollector::Sink` finds the streaming buffer full, it returns `BLOCKED` before appending the chunk. The temporary is then destroyed, while `ExecutionBatch::rows` remains non-zero. On retry, the batch has logical rows but no payload, so an empty chunk reaches the stream consumer and is interpreted as EOF.

This violates the asynchronous sink ownership invariant: **a call that returns `BLOCKED` must not consume its input**.

Observed with DuckDB submodule revision:

```text
398033a962719ac09868f4484ec4f97353bb0325
```

## Minimal reproduction

After a Release build:

```python
import duckdb

con = duckdb.connect()
rows = con.execute("SELECT range AS a FROM range(1000000)").fetchall()
print(len(rows))
```

Expected:

```text
1000000
```

Observed with the default streaming buffer (`976.5 KiB`):

```text
126976
```

The underlying query still computes all rows:

```sql
SELECT count(*) FROM range(1000000);
-- 1000000
```

This shows that rows are lost at the result sink/consumer boundary rather than at the source or during query computation.

## Buffer-size correlation

| `streaming_buffer_size` | `fetchall()` rows |
| --- | ---: |
| Default / 1 MB | 126,976 |
| 2 MB | 251,904 |
| 64 MB | 1,000,000 |

The following result APIs are affected:

- `fetchall()`
- `fetchmany()`
- `to_arrow_table()`
- `to_arrow_reader()`

The issue reproduces with both one and multiple DuckDB execution threads. It does not require Ray, a UDF, networking, or an extension.

## Root cause

Relevant code paths:

- `external/duckdb/src/parallel/pipeline_executor.cpp`
- `external/duckdb/src/execution/physical_operator.cpp`
- `external/duckdb/src/execution/operator/helper/physical_buffered_collector.cpp`
- `external/duckdb/src/main/stream_query_result.cpp`

Failure sequence:

1. `PipelineExecutor` submits an `ExecutionBatch` to the result sink.
2. The default `PhysicalOperator::SinkBatch` calls `TakeMaterializedExecutionBatch`.
3. The helper moves `batch.materialized` into a local `unique_ptr<DataChunk>`.
4. `PhysicalBufferedCollector::Sink` sees that the buffer is full and returns `BLOCKED` before appending the chunk.
5. The local chunk is destroyed as the call returns.
6. `PipelineExecutor` retries the same logical batch; `batch.rows` is still non-zero, but `batch.materialized` is empty.
7. The fallback submits an empty chunk, which the stream result path interprets as normal EOF and closes the query early.

The operation succeeds without an exception, so callers can unknowingly consume incomplete data.

## Impact

- Normal local queries can trigger the bug with the default approximately 1 MiB buffer.
- Queries return success rather than an error.
- Exports, ETL pipelines, model inputs, statistics, and downstream applications can silently use incomplete results.
- Increasing the buffer only delays the failure and is not a correctness fix.

## Separate ordering regression

Arrow dictionary and multi-batch scans also expose an independent batch-ordering regression (F-01B) that is not caused by the payload move described here.

kaka11chen:
## Summary

  Fixes #14.

  Preserve materialized `ExecutionBatch` payloads in the owning batch and pass a
  borrowed `DataChunk &` to the legacy `Sink` and `Execute` fallbacks instead of
  moving the payload into call-local storage.

  This preserves input ownership when:

  - a sink returns `BLOCKED` and must retry the same batch;
  - an operator returns `HAVE_MORE_OUTPUT` and must continue producing output
    from the same input.

  The change prevents both streaming-result truncation and silent under-counting
  from multi-output operators such as `PhysicalCrossProduct`. It also raises an
  internal consistency error when a batch reports logical rows without a payload.

  Regression coverage includes:

  - a deterministic sink that returns `BLOCKED` twice before accepting its input;
  - a deterministic operator that returns `HAVE_MORE_OUTPUT` twice and verifies
    the same payload address and contents on every call;
  - small streaming buffers across `fetchall`, `fetchmany`, Arrow tables, and
    Arrow readers;
  - strings, NULLs, dictionary input, ordering, and single-/multi-threaded
    execution;
  - CROSS JOIN detail multisets, row counts, SQL and Relation API aggregation,
    NULLs, both input orientations, and vector/multi-chunk boundaries.

  ## Validation

  - Native `[execution_batch]`: 3 cases, 120,120 assertions passed
  - `pytest tests/fast/api/test_streaming_result.py`: 20 passed
  - `pytest tests/fast/relational_api/test_joins.py`: 31 passed
  - `pytest tests/fast/test_udf_process.py`: 4 passed
  - `scripts/run_release_tests.sh`: 262 passed, 4 skipped
    - skips require an external writable MinIO/S3 endpoint
  - Release incremental build completed successfully

  ## Checklist

  - [x] Security implications have been considered.
  - [x] Native changes were compiled; relevant tests passed.

### DIFF
--- a/external/duckdb/test/execution/test_execution_batch.cpp
+++ b/external/duckdb/test/execution/test_execution_batch.cpp
@@ -1,0 +1,231 @@
+// SPDX-FileCopyrightText: 2026 Vane contributors
+// SPDX-License-Identifier: MIT
+
+#include "catch.hpp"
+
+#include "duckdb/execution/physical_operator.hpp"
+#include "duckdb/execution/physical_plan_generator.hpp"
+#include "duckdb/main/connection.hpp"
+#include "duckdb/main/database.hpp"
+#include "duckdb/parallel/interrupt.hpp"
+#include "duckdb/parallel/thread_context.hpp"
+
+using namespace duckdb;
+
+namespace {
+
+class RetryBlockingSink : public PhysicalOperator {
+public:
+	explicit RetryBlockingSink(PhysicalPlan &physical_plan)
+	    : PhysicalOperator(physical_plan, PhysicalOperatorType::RESULT_COLLECTOR, {}, 3) {
+	}
+
+	SinkResultType Sink(ExecutionContext &, DataChunk &chunk, OperatorSinkInput &) const override {
+		sink_calls++;
+		if (sink_calls <= 2) {
+			return SinkResultType::BLOCKED;
+		}
+
+		accepted_calls++;
+		accepted_rows = chunk.size();
+		accepted_values.clear();
+		for (idx_t row = 0; row < chunk.size(); row++) {
+			accepted_values.push_back(chunk.GetValue(0, row).GetValue<int64_t>());
+		}
+		return SinkResultType::NEED_MORE_INPUT;
+	}
+
+	mutable idx_t sink_calls = 0;
+	mutable idx_t accepted_calls = 0;
+	mutable idx_t accepted_rows = 0;
+	mutable vector<int64_t> accepted_values;
+};
+
+class MultiOutputOperator : public PhysicalOperator {
+public:
+	explicit MultiOutputOperator(PhysicalPlan &physical_plan)
+	    : PhysicalOperator(physical_plan, PhysicalOperatorType::PROJECTION, {LogicalType::BIGINT}, 3) {
+	}
+
+	OperatorResultType Execute(ExecutionContext &, DataChunk &input, DataChunk &output, GlobalOperatorState &,
+	                           OperatorState &) const override {
+		execute_calls++;
+		observed_payloads.push_back(&input);
+		observed_values.emplace_back();
+		for (idx_t row = 0; row < input.size(); row++) {
+			observed_values.back().push_back(input.GetValue(0, row).GetValue<int64_t>());
+		}
+		output.Reference(input);
+
+		if (execute_calls <= 2) {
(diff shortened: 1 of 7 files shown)


external/duckdb/test/execution/test_execution_batch.cpp
QuakeWang, line 1: This inherited DuckDB file now contains Vane changes but lacks the MIT AND Apache-2.0 SPDX header and both copyright notices required by the repository’s source provenance policy.

kaka11chen: This file is newly added relative to the recorded DuckDB baseline, rather than an inherited DuckDB file. Per CONTRIBUTING.md and SOURCE_PROVENANCE.md, new and modified source under external/duckdb remains MIT-licensed. The repository’s header checker also classifies new files there as MIT with the Vane copyright only, so the existing header matches the current provenance policy.

hubgeter:
LGTM

QuakeWang (approved):
+1

PR #21, opened 2026-07-19
[fix](build) Make native test gate reproducible

kaka11chen:
## Summary

- Add scripts/run_native_tests.sh as the shared development and CI entry point, using the pinned vcpkg Arrow and Flight packages, Release mode, C++20, a fresh CMake configuration, and a memory-safe two-job default.
- Propagate the imported Arrow target and the C++20 requirement to the distributed exchange test objects and final unittest executable, fixing clean standalone builds that could not find arrow/io/api.h.
- Align the root native build, DEVELOPMENT.md, and CI on the same configuration before release artifacts are built.

The exchange-disabled DuckDB target keeps its C++11 minimum declaration. Enabling Arrow Flight exchange raises the standalone build to the C++20 requirement imposed by the pinned Arrow 24 package.

## Validation

- Native distributed suite: 89 test cases and 3564 assertions passed.
- Release wheel rebuilt and reinstalled with uv pip install . --no-build-isolation.
- Release gate: 262 passed and 4 skipped; the skips require an available MinIO endpoint.
- Pre-commit checks passed for all changed files.


QuakeWang (approved):
LGTM

PR #29, opened 2026-07-20
fix: restore insertion order preservation by default

### LINKED ISSUE #20: [P1][Bug] Default preserve_insertion_order=false breaks Arrow order contracts
## Summary

On the latest `main` (`74f8d91976c69d8861262944eb61f4b8a05abd42`), a fresh connection reports `preserve_insertion_order=false`. Parallel scans of an insertion-order-preserving Arrow source can therefore return complete rows in worker-dependent order.

The ordered ExecutionBatch/collector path itself appears healthy: explicitly setting `preserve_insertion_order=true` preserves the full sequence across thread counts, Arrow RecordBatch sizes, dictionary arrays, and the Parquet roundtrip case.

This should be treated as a default/API contract regression, not as evidence that the ExecutionBatch batch-index integration is broken.

## Environment

- Vane `main`: `74f8d91976c69d8861262944eb61f4b8a05abd42`
- DuckDB engine: `v1.6.0-dev2`
- Python: `3.12.3`
- PyArrow: `25.0.0`
- Local engine runner: `VANE_RUNNER=local-fast`

## Reproduction

```python
import os

os.environ["VANE_RUNNER"] = "local-fast"

import duckdb
import pyarrow as pa

n = 70_000
table = pa.Table.from_batches(
    pa.table({"a": range(n)}).to_batches(10)
)

for threads in (1, 2, 4, 8):
    con = duckdb.connect(config={"threads": threads})
    setting = con.execute(
        "select current_setting('preserve_insertion_order')"
    ).fetchone()[0]
    actual = [row[0] for row in con.from_arrow(table).execute().fetchall()]
    mismatch = next(
        (i for i, value in enumerate(actual) if i != value),
        None,
    )
    print(
        threads,
        setting,
        len(actual),
        len(set(actual)),
        mismatch,
    )
```

One run produced:

| threads | setting | rows | unique rows | first mismatch |
| ---: | :---: | ---: | ---: | ---: |
| 1 | false | 70,000 | 70,000 | none |
| 2 | false | 70,000 | 70,000 | 20 |
| 4 | false | 70,000 | 70,000 | 0 |
| 8 | false | 70,000 | 70,000 | 0 |

The precise mismatch index is scheduling-dependent. The output remains complete; this is reordering rather than row loss.

## Control: explicit preservation works

Adding this connection option makes all tested thread counts preserve the exact sequence:

```python
con = duckdb.connect(
    config={
        "threads": threads,
        "preserve_insertion_order": True,
    }
)
```

Verified controls on the same `main` revision:

- unique integers, threads `1/2/4/8`: exact sequence passes;
- Arrow RecordBatch sizes `1/10/2048/4097`: exact sequence passes for every thread count;
- dictionary batches with `PRAGMA verify_parallelism`: passes;
- `duckdb.arrow(...).project(...).to_arrow_table()` Parquet roundtrip with batch sizes `7/51/99/100/101/500/1000/2000`: all pass.

## Existing test failures

The following tests still fail under the current default:

```text
tests/fast/arrow/test_dictionary_arrow.py::TestArrowDictionary::test_dictionary_batches
tests/fast/arrow/test_dictionary_arrow.py::TestArrowDictionary::test_dictionary_batches_parallel
tests/fast/arrow/test_integration.py::TestArrowIntegration::test_parquet_roundtrip
```

Focused result: `3 failed in 1.53s`.

## Relevant code

- `external/duckdb/src/include/duckdb/main/settings.hpp`: `PreserveInsertionOrderSetting::DefaultValue` is `"false"`.
- `external/duckdb/src/execution/physical_plan/plan_insert.cpp`: the setting controls whether insertion order is preserved.
- `external/duckdb/src/execution/operator/helper/physical_result_collector.cpp`: when preservation is disabled, the plan selects the parallel unordered collector; when enabled and batch indexes are supported, it selects the batch collector.
- `external/duckdb/src/function/table/arrow.cpp`: Arrow assigns increasing RecordBatch indexes and exposes them through `GetPartitionData`.

liwuhen:
## Summary

Fresh connections currently default `preserve_insertion_order` to `false`.
Parallel scans of insertion-order-preserving Arrow sources can therefore return
complete rows in worker-dependent order.

This restores `preserve_insertion_order=true` as the default while retaining
`false` as an explicit opt-out for users who prefer unordered parallelism.

Regression coverage validates:

- default and explicit-`true` connections preserve exact Arrow row order;
- thread counts `1`, `2`, `4`, and `8`;
- RecordBatch sizes `1`, `10`, `2048`, and `4097`;
- explicit `false` preserves the row count and multiset without requiring order.

Fixes #20

## Validation

- Issue-focused regression tests and the previously failing dictionary/Parquet
  cases: `15 passed`.
- Complete Arrow fast suite: `761 passed, 17 skipped, 1 xfailed, 1 xpassed`.
- `python3 scripts/sync_duckdb_source_id.py --check`
- Native extension compiled using the incremental Release build directory.
- The full release suite was not completed under the local Flight-disabled
  validation build; it stalled in `test_local_runner_write_parquet_e2e`.

## Checklist

- [x] Native changes were compiled; Python-only changes passed relevant tests.

### DIFF
--- a/external/duckdb/src/common/settings.json
+++ b/external/duckdb/src/common/settings.json
@@ -931,7 +931,7 @@
         "description": "Whether or not to preserve insertion order. If set to false the system is allowed to re-order any results \"\n\t    \"that do not contain ORDER BY clauses.",
         "type": "BOOLEAN",
         "default_scope": "global",
-        "default_value": "false"
+        "default_value": "true"
     },
     {
         "name": "produce_arrow_string_view",
--- a/external/duckdb/src/include/duckdb/main/settings.hpp
+++ b/external/duckdb/src/include/duckdb/main/settings.hpp
@@ -1367,7 +1367,7 @@ struct PreserveInsertionOrderSetting {
 	    "Whether or not to preserve insertion order. If set to false the system is allowed to re-order any results "
 	    "that do not contain ORDER BY clauses.";
 	static constexpr const char *InputType = "BOOLEAN";
-	static constexpr const char *DefaultValue = "false";
+	static constexpr const char *DefaultValue = "true";
 	static constexpr SettingScopeTarget Scope = SettingScopeTarget::GLOBAL_DEFAULT;
 	static constexpr idx_t SettingIndex = 82;
 };
--- a/tests/fast/arrow/test_insertion_order.py
+++ b/tests/fast/arrow/test_insertion_order.py
@@ -1,0 +1,53 @@
+# SPDX-FileCopyrightText: 2026 Vane contributors
+# SPDX-License-Identifier: Apache-2.0
+
+from collections import Counter
+
+import pytest
+
+import duckdb
+
+pa = pytest.importorskip("pyarrow")
+
+ROW_COUNT = 70_000
+ORDER_CASES = (
+    pytest.param(1, 1, id="threads-1-batch-1"),
+    pytest.param(2, 10, id="threads-2-batch-10"),
+    pytest.param(4, 2048, id="threads-4-batch-2048"),
+    pytest.param(8, 4097, id="threads-8-batch-4097"),
+)
+
+
+def make_batched_table(batch_size):
+    table = pa.table({"value": range(ROW_COUNT)})
+    return pa.Table.from_batches(table.to_batches(max_chunksize=batch_size))
+
+
+@pytest.mark.parametrize("threads,batch_size", ORDER_CASES)
+@pytest.mark.parametrize(
+    "preserve_insertion_order",
+    [None, True],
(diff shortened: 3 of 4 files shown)


kaka11chen (commented):
This section of `settings.hpp` is autogenerated, but the source configuration in `src/common/settings.json` still sets `preserve_insertion_order` to `false`. Running `generate_settings.py` will therefore overwrite this change and silently restore the
incorrect default. Please change the JSON value to true and regenerate the header.

kaka11chen (approved):
LGTM

Issue #50, opened 2026-07-20
vLLM: Fix execution races and resource-lifecycle issues


kaka11chen:
## Problem

Vane's vLLM execution path spans native `PhysicalVLLM`, Python executors,
shared Ray actors, and distributed named pools.

The basic single-engine path works with valid configuration, but several
correctness and resource-lifecycle problems appear under prefix routing,
concurrent Ray execution, and distributed execution:

- Invalid vLLM options can reach native execution instead of failing early.
- Common-prefix calculation can truncate a multi-byte UTF-8 character.
- Parallel producers and blocked finalizers can miss a completion or scheduler
  wakeup transition, causing queries to stall.
- Executors sharing Ray actors can interfere with each other's request, result,
  error, and terminal state.
- Terminal operations can overtake pending submissions, and one-shot waits can
  leave buffered results unconsumed.
- Named vLLM pools can be reused with incompatible configuration or remain alive
  after the owning distributed plan terminates.

## Impact

Depending on the execution mode, these problems can cause:

- queries that stall instead of completing or failing;
- incorrect or unstable prefix routing;
- results or terminal state leaking between concurrent executors;
- named Ray actors and GPU resources remaining allocated after query completion;
- one distributed plan incorrectly affecting a pool shared with another plan.

## Expected behavior

- Invalid configuration fails before execution begins.
- Prefix routing preserves complete UTF-8 characters.
- Concurrent executors remain isolated and continue making progress.
- Submission, completion, and finalization transitions cannot be lost or
  reordered incorrectly.
- Named pools validate compatible reuse and are released only after their final
  owner exits.
- Query-owned Ray processes and GPU resources are released on every plan
  termination path.

kaka11chen:
Latest-main audit found one remaining deterministic RemoteVLLMExecutor race after #52.

`_drain_ready_actor` still increments `_results_per_actor` before entering the result condition and appending to `_result_buffer`. `all_tasks_finished()` can run between those steps, observe an empty buffer and completed counts, and report completion before the last result is visible.

A barrier-controlled fake actor reproduced:

```text
reported_finished_before_append = True
buffer_after_drain = 1
```

The local/native invalid-state and teardown regression files pass (21 tests), so #52 is effective for its intended scope. Please add atomic count+buffer publication, or define completion from one condition-protected state, to this issue's remaining work.

hubgeter:
Closing this umbrella tracker as completed after moving the remaining focused
work into dedicated follow-up issues.

[...]

PR #52, opened 2026-07-20
[fix](vllm)Protect PhysicalVLLM queries from invalid state and concurrent teardown

hubgeter:
## Summary
issue #50 

PhysicalVLLM must reject unsafe normalized options before native execution, preserve complete UTF-8 prefix boundaries, and coordinate the exact producer count. Finalizers support an optional one-shot scheduler wakeup while executors without that protocol continue through the blocking fallback.

Blocking waits must remain outside executor_call_mutex so producers can submit work that wakes the waiter. Another finalizer can drain the last result and shut the Python bridge down during that window, so WaitForResult snapshots the inner Python object under the GIL and treats an already-cleared handle as terminal. Shutdown checks and resets the same handle under the GIL.

## Validation

- Built and reinstalled the native extension in Release mode.
  - pytest -q tests/fast/test_vllm_native_regressions.py tests/fast/test_vllm_runtime_fixes.py: 21 passed.
  - pytest -q tests/fast/test_udf_executor_lifecycle.py -k 'vllm and not actor_wait_for_result_finishes_per_executor_before_global_finish': 14 passed.
  - Repeated test_parallel_legacy_finalizer_wait_survives_shutdown 10 times: 10/10 passed.
  - Ruff, clang-format, pre-commit, and git diff --check passed for the changed fix files.
  - CPU-only validation with a synthetic 300,000-row dataset and PRAGMA threads=2; no performance claims. CUDA and real-model inference were not tested.

## Checklist

- [x] Native changes were compiled; Python-only changes passed relevant tests.

### DIFF
--- a/duckdb/execution/vllm.py
+++ b/duckdb/execution/vllm.py
@@ -131,9 +131,9 @@ class LocalVLLMExecutor(VLLMExecutor):
 
         self.model = model
         self.engine_args = dict(engine_args)
-        self.llm = None
+        self.llm: Any = None
         self.engine_ready = threading.Event()
-        self.engine_error_message = None
+        self.engine_error_message: str | None = None
         self.engine_init_timeout_s = _vllm_engine_init_timeout_s(engine_init_timeout_s)
 
         sampling_params = generate_args.pop("sampling_params", None)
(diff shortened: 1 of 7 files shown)


hubgeter:
@codex  review

kaka11chen (approved):
LGTM

PR #124, opened 2026-07-21
[fix](udf) Restore socket timeout unconditionally after the startup handshake

### LINKED ISSUE #65: [P1][Bug] UDF startup handshake does not restore blocking socket timeout
## What happens

`_recv_expected` temporarily sets a socket timeout and attempts to restore it only when the previous timeout is not `None`. A normal blocking socket uses `None`, so the temporary startup timeout leaks into normal task communication.

A fake-socket probe starting from blocking mode observes a 2-second timeout after the handshake.

dchaudhari7177:
## What

Fixes #65. `_recv_expected` used `restore_timeout = None` both as "we never overrode the timeout" and as the captured previous value — but `None` is exactly what a normal blocking socket's `gettimeout()` returns, so after a handshake that started from blocking mode the temporary timeout was never restored and leaked into all subsequent task communication.

## Fix

A separate `timeout_overridden` flag tracks whether we changed the timeout; the `finally` block restores `previous_timeout` unconditionally when set (`None` is a valid `settimeout` argument), on both the success and failure paths.

## Tests

`tests/fast/test_udf_socket_timeout_restore.py`, parameterized over the three starting modes from the acceptance criteria — blocking (`None`), an existing finite timeout (`5.0`), and non-blocking (`0.0`) — for both a successful and a failing handshake, asserting the original timeout is back and that the temporary timeout was genuinely applied in between. Plus one case pinning that a call without `timeout_s` never touches the socket timeout.

Local note: the native `_duckdb` module isn't buildable on this machine, so I verified the exact scenarios with the module loaded standalone (all 7 pass — blocking mode restores `[2.0 → None]` where the old code left `2.0` in place) and ran `ruff check`/`format --check`. The pytest file uses the normal `tests/fast` imports for CI.

Validated against `02389fd`.

### DIFF
--- a/duckdb/execution/udf_subprocess.py
+++ b/duckdb/execution/udf_subprocess.py
@@ -388,17 +388,22 @@ class _SingleSubprocessExecutor(BaseUDFExecutor):
 
     def _recv_expected(self, expected: tuple[int, ...], *, timeout_s: float | None = None) -> tuple[int, bytes]:
         sock = self._require_socket()
-        restore_timeout = None
+        # None is a valid previous timeout (blocking mode), so a separate flag
+        # tracks whether we overrode it — restoring must be unconditional or
+        # the temporary startup timeout leaks into normal task communication.
+        timeout_overridden = False
+        restore_timeout: float | None = None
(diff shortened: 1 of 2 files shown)


kaka11chen:

PR #125, opened 2026-07-21
[fix](video) Reject zero/negative VANE_MAX_CONCURRENT_DECODES instead of deadlocking

### LINKED ISSUE #68: [P1][Bug] Video decoder concurrency accepts zero and blocks every decode
## What happens

`VANE_MAX_CONCURRENT_DECODES` is converted with `int()` without a positive lower-bound check. Setting it to zero creates `Semaphore(0)`, so every decoder waits forever.

dchaudhari7177:
## What

Fixes #68. `VANE_MAX_CONCURRENT_DECODES` was converted with a bare `int()` — setting it to `0` builds `Semaphore(0)`, so every decode blocks forever with no hint why.

## How

The conversion moves into `_max_concurrent_decodes_from_env()`, which rejects zero, negative, and non-integer values at configuration load with an error naming the variable and the accepted range:

```
ValueError: VANE_MAX_CONCURRENT_DECODES must be a positive integer (>= 1), got '0'
```

`1` and ordinary positive integers are accepted; unset still defaults to `1`.

## Tests

`tests/fast/test_video_decode_concurrency_env.py` — parameterized accept (`1`/`2`/`64`), unset default, non-positive rejects (`0`/`-1`/`-64`), and invalid rejects (empty/word/float/nan), all via `monkeypatch.setenv`/`delenv` so process state is restored per the acceptance criteria.

Local note: the native `_duckdb` build isn't available on this machine, so I verified the helper standalone (8/8 scenarios) and ran `ruff check`/`format`; the pytest file uses the standard `tests/fast` imports for CI.

Validated against `02389fd`.

### DIFF
--- a/duckdb/datasource/video_reader.py
+++ b/duckdb/datasource/video_reader.py
@@ -168,8 +168,25 @@ def _emit_reader_timing(
     )
 
 
+def _max_concurrent_decodes_from_env() -> int:
+    """Positive decode concurrency from VANE_MAX_CONCURRENT_DECODES.
+
+    Zero would build Semaphore(0) and block every decode forever, and a
+    negative or non-integer value is equally meaningless, so all of those are
+    rejected loudly at configuration load instead.
+    """
+    raw = os.environ.get("VANE_MAX_CONCURRENT_DECODES", "1")
+    try:
+        value = int(raw)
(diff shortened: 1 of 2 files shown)


kaka11chen:

PR #126, opened 2026-07-21
[fix](env) Timeout env parsers reject NaN and infinity like the vllm reference

### LINKED ISSUE #69: [P2][Bug] Timeout environment parsers accept NaN and infinity
## What happens

Several positive-float environment helpers only compare the parsed value with zero. IEEE NaN bypasses those comparisons, and infinity can also be accepted.

Affected current paths include subprocess UDF, Ray actor pool, and Ray safe-get helpers. The vLLM copy was already corrected and can serve as the consistency reference.

dchaudhari7177:
## What

Fixes #69. Three `_positive_float_env` helpers — `duckdb/execution/udf_subprocess.py`, `duckdb/execution/udf_ray_actor_pool.py`, and `duckdb/runners/ray/safe_get.py` — only checked `value < 0.0`. `float("nan")` fails every ordering comparison and `float("inf")` passes it, so both were accepted as timeouts.

## How

Each helper now applies exactly the guard the corrected `vllm.py` copy uses, as the issue's consistency reference:

```python
if not math.isfinite(value) or value < 0.0:
    raise ValueError(f"{name} must be finite and non-negative")
```

- rejects NaN, ±infinity, and negatives; invalid strings keep raising from `float()` itself
- unset-variable semantics preserved: the two default-taking helpers return their default, safe-get returns `None`
- zero stays accepted, matching the vllm reference implementation (it rejects only negatives) — happy to tighten to strictly-positive everywhere if that's the intended reading of the criteria

## Tests

`tests/fast/test_positive_float_env_finiteness.py` parameterizes **every consumer** (per the criteria) over accept (`0`/`2.5`/`30`), reject (`nan`/`NaN`/`inf`/`Infinity`/`-inf`/`-1`/`-0.5`), invalid strings, and unset semantics — 33 cases.

Local note: no native `_duckdb` build on this machine, so the helpers were verified standalone via AST extraction (21/21 scenarios) plus `ruff check`/`format`; the pytest file imports normally for CI.

Validated against `02389fd`.

### DIFF
--- a/duckdb/execution/udf_ray_actor_pool.py
+++ b/duckdb/execution/udf_ray_actor_pool.py
@@ -3,6 +3,7 @@
 
 from __future__ import annotations
 
+import math
 import os
 from typing import TYPE_CHECKING, Any
 
@@ -212,8 +213,10 @@ def _positive_float_env(name: str, default: float | None = None) -> float | None
     if not raw:
         return default
     value = float(raw)
-    if value < 0.0:
-        raise ValueError(f"{name} must be non-negative")
+    # NaN compares false against 0.0 and infinity compares true, so both
+    # slipped through a plain sign check (matches the vllm.py reference).
+    if not math.isfinite(value) or value < 0.0:
+        raise ValueError(f"{name} must be finite and non-negative")
     return value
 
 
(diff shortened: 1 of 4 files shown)


kaka11chen:

PR #127, opened 2026-07-21
[fix](ai) chunk_text boundary validation and zero-weight aggregation guard

### LINKED ISSUE #70: [P2][Bug] AI text chunking accepts invalid size and overlap values
## What happens

The AI text chunk helper does not consistently validate `max_chars` and overlap/weight boundaries. Zero or invalid combinations can create empty chunks or zero-weight aggregation and NaN results.

dchaudhari7177:
## What

Fixes #70, in `vane/ai/functions.py`.

**`chunk_text`** accepted degenerate parameters silently:
- `max_chars <= 0` → now `ValueError: max_chars must be a positive integer`
- `overlap_chars` outside `[0, max_chars)` → `ValueError` naming both values (overlap == max_chars previously still worked only because of the `max(1, …)` step clamp, but produced maximal re-reading; it's rejected as the criteria require)
- non-integer / non-finite inputs (`2.5`, `nan`, strings, `None`, and `bool` — an `int` subclass — for both parameters) → `ValueError`
- `overlap_chars=0` remains valid; the docstring documents the raises

**`_weighted_average_embeddings`** divided by `w.sum()` unconditionally, so an all-zero weight vector produced NaN embeddings. When the total is zero or non-finite it now falls back to the unweighted mean (then normalizes as before); the weighted path is unchanged.

## Tests

`tests/ai/test_chunk_text_validation.py`: a normal multi-chunk test that verifies chunk sizes, the exact overlap, and loss-free reassembly; boundary rejects for both parameters (13 parameterized cases); `overlap=0` accept; and both aggregation paths (zero weights stay finite and equal the normalized unweighted mean; weighted path unchanged).

Local note: no native `_duckdb` build here, so both functions were verified standalone via AST extraction (13/13 scenarios) and `ruff check`/`format` pass; the pytest file imports normally for CI.

Validated against `02389fd`.

### DIFF
--- a/tests/ai/test_chunk_text_validation.py
+++ b/tests/ai/test_chunk_text_validation.py
@@ -1,0 +1,75 @@
+# SPDX-FileCopyrightText: 2026 Vane contributors
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
+
+import numpy as np
+import pytest
+
+from vane.ai.functions import _weighted_average_embeddings, chunk_text
+
+
+def test_multi_chunk_content_with_overlap():
+    text = "abcdefghij" * 10  # 100 chars
+    chunks = chunk_text(text, max_chars=40, overlap_chars=10)
+    assert len(chunks) > 1
+    assert all(chunks), "no empty chunks"
+    assert all(len(c) <= 40 for c in chunks)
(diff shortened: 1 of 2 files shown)


kaka11chen:

PR #128, opened 2026-07-21
[fix](ai) Token metrics: caller-owned snapshots, lock-consistent summary, derived totals

### LINKED ISSUE #67: [P2][Bug] AI metrics getters expose mutable collector state
## What happens

The metrics collector returns references to its internal mutable counter objects. Mutating an object returned by `get` changes later `summary` output.

The current snapshot also copies object references under the lock and reads fields later, so it is not a consistent point-in-time snapshot.

dchaudhari7177:
## What

Fixes #67, in `vane/ai/metrics.py`, per the three acceptance criteria plus the definition question:

1. **Immutable-to-callers snapshots** — `get_token_metrics()` returns `dataclasses.replace` copies, so mutating a returned entry no longer changes the collector's counters or later summaries. (Entries stay internally mutable for the `+=` accumulation; freezing the dataclass would have broken recording.)
2. **Point-in-time summary** — `get_token_metrics_summary()` previously copied entry *references* under the lock and read fields after releasing it; a concurrent `record` could tear totals. All field reads now happen while holding the lock. `by_provider` dicts were already fresh per call and are asserted caller-owned.
3. **`total_tokens` defined** — when a provider omits it, it's derived as `(input or 0) + (output or 0)`; when reported, the reported value is used as-is (never double-counted). Documented in the recording path.

## Tests

`tests/ai/test_metrics_snapshots.py`: snapshot-mutation isolation (entry and summary), derivation + no-double-count, a 4-writer concurrent test probing the summary 200× for the invariants `total == in + out` and `in == requests` (which only hold if the snapshot is torn-free, since writers omit the total and rely on derivation), and `by_provider` copy ownership.

Local note: no native `_duckdb` build here — the module is pure Python, so I ran the identical scenarios against the real `metrics.py` loaded standalone (6/6, including the 300-probe concurrency check) and `ruff check`/`format` pass. The pytest file imports normally for CI.

Validated against `02389fd`.

### DIFF
--- a/tests/ai/test_metrics_snapshots.py
+++ b/tests/ai/test_metrics_snapshots.py
@@ -1,0 +1,80 @@
+# SPDX-FileCopyrightText: 2026 Vane contributors
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
+
+import threading
+
+import pytest
+
+from vane.ai.metrics import (
+    get_token_metrics,
+    get_token_metrics_summary,
+    record_token_metrics,
+    reset_token_metrics,
+)
+
+
+@pytest.fixture(autouse=True)
+def _clean_metrics():
+    reset_token_metrics()
+    yield
+    reset_token_metrics()
+
+
+def test_mutating_snapshot_does_not_change_internal_state():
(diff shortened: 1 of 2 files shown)


kaka11chen:

PR #160, opened 2026-07-22
[fix](ai) Concatenate Anthropic text blocks instead of reading content[0]

### LINKED ISSUE #140: [P1][AI] Anthropic prompter text extraction breaks on leading thinking blocks
## Problem

`AnthropicPrompter.prompt` extracts plain-text responses via `response.content[0].text` (`vane/ai/providers/anthropic.py:316-318`), assuming the first content block is a text block. Extended-thinking models (and Anthropic-compatible endpoints with always-on thinking) return a leading `thinking` block, which has `.thinking` and no `.text`.

## Failure scenario

Any prompt against a thinking-enabled model → `AttributeError: 'ThinkingBlock' object has no attribute 'text'`, or silently empty output on compatible servers exposing an empty `text` field. This was actually observed in practice against an Anthropic-compatible endpoint whose served model always emits a thinking block.

The structured-output path just above (`anthropic.py:307-313`) already iterates `response.content` with `block.type` checks and is robust — only the plain-text path is broken.

## Suggested fix

```python
return next((b.text for b in response.content if getattr(b, "type", None) == "text"), None)
```

Reviewed commit: c43bd706.

caomaocao:
## Summary

Extended-thinking Anthropic models return a leading `thinking` block, so indexing `response.content[0].text` raised `AttributeError` (or yielded empty text on compatible servers). `prompt()` now filters content blocks by `type == "text"` and concatenates all text blocks — mirroring the Anthropic SDK's own `get_final_text()` helper, so responses that split text across multiple blocks are never silently truncated — and returns `None` when no text block exists. Existing Anthropic response mocks got realistic `.type`/`.usage` attributes so they keep exercising the production path when the `anthropic` extra is installed.

## Related issue

Closes #140

## Validation

- `tests/fast/test_anthropic_prompter_content.py` (5 cases: [thinking, text] → text; [text] → text; [thinking, text("a"), text("b")] → "ab"; [thinking only] → None; empty → None), red-run verified: on unfixed code the cases reproduce the issue's exact `AttributeError`.
- 210 tests green locally under the no-duckdb import harness (16 pre-existing environment-only failures, unrelated; engine-dependent suites run in CI). `ruff check`/`format` clean; pre-commit hooks passed.

## Checklist

- [x] Public behavior and compatibility impact are documented in code/docstrings; CHANGELOG intentionally untouched per maintainer guidance.
- [x] New dependencies, copied code, model assets, and datasets have compatible licenses and are recorded where required (none added).
- [x] Security implications of UDFs, serialization, remote code, network access, and untrusted input have been considered (no new surface).
- [x] Native changes were compiled; Python-only changes passed relevant tests (Python-only).

https://claude.ai/code/session_014az5eSZxzP2SBDyivvivCk

### DIFF
--- a/vane/ai/providers/anthropic.py
+++ b/vane/ai/providers/anthropic.py
@@ -317,8 +317,6 @@ class AnthropicPrompter:
             return None
 
         if response.content:
-            return next(
-                (block.text for block in response.content if getattr(block, "type", None) == "text"),
-                None,
-            )
+            text_blocks = [block.text for block in response.content if getattr(block, "type", None) == "text"]
+            return "".join(text_blocks) if text_blocks else None
         return None
--- a/vane/ai/providers/anthropic.py
+++ b/vane/ai/providers/anthropic.py
@@ -317,5 +317,8 @@ class AnthropicPrompter:
             return None
 
         if response.content:
-            return response.content[0].text
+            return next(
+                (block.text for block in response.content if getattr(block, "type", None) == "text"),
+                None,
+            )
         return None
--- a/tests/ai/test_ai_functions.py
+++ b/tests/ai/test_ai_functions.py
@@ -1227,8 +1227,13 @@ class TestAnthropicStructuredOutput:
         tool_block.type = "tool_use"
         tool_block.input = {"name": "Alice", "age": 30}
 
+        mock_usage = MagicMock()
+        mock_usage.input_tokens = 12
+        mock_usage.output_tokens = 7
+
         mock_response = MagicMock()
         mock_response.content = [tool_block]
+        mock_response.usage = mock_usage
 
         async def mock_create(**kwargs):
             assert "tools" in kwargs
@@ -1252,9 +1257,16 @@ class TestAnthropicStructuredOutput:
         )
 
         text_block = MagicMock()
+        text_block.type = "text"
         text_block.text = "Hello world"
+
+        mock_usage = MagicMock()
+        mock_usage.input_tokens = 8
+        mock_usage.output_tokens = 3
+
         mock_response = MagicMock()
         mock_response.content = [text_block]
+        mock_response.usage = mock_usage
 
         async def mock_create(**kwargs):
             assert "tools" not in kwargs
@@ -1356,9 +1368,16 @@ class TestAnthropicMultimodal:
 
         captured_messages = []
         text_block = MagicMock()
+        text_block.type = "text"
         text_block.text = "I see an image"
+
+        mock_usage = MagicMock()
+        mock_usage.input_tokens = 25
+        mock_usage.output_tokens = 6
+
         mock_response = MagicMock()
         mock_response.content = [text_block]
+        mock_response.usage = mock_usage
 
         async def mock_create(**kwargs):
             captured_messages.append(kwargs["messages"])
@@ -1389,9 +1408,16 @@ class TestAnthropicMultimodal:
 
         captured_messages = []
         text_block = MagicMock()
+        text_block.type = "text"
         text_block.text = "ok"
+
+        mock_usage = MagicMock()
+        mock_usage.input_tokens = 15
+        mock_usage.output_tokens = 2
+
         mock_response = MagicMock()
         mock_response.content = [text_block]
+        mock_response.usage = mock_usage
 
         async def mock_create(**kwargs):
             captured_messages.append(kwargs["messages"])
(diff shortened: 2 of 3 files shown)


kaka11chen (changes requested):
The production fix is correct, but this change leaves an existing Anthropic test failing when the optional anthropic dependency is installed.

In tests/ai/test_ai_functions.py::TestAnthropicStructuredOutput::test_plain_text_response, text_block is a MagicMock with only .text set. With the new type filter, its .type attribute is another MagicMock rather than "text", so prompt() returns None and the existing assertion fails. Please set text_block.type = "text" (and ideally make the other Anthropic response mocks realistic too), then run python -m pytest tests/ai/test_ai_functions.py -k anthropic with the Anthropic extra installed.

The new regression coverage itself looks good.

kaka11chen (commented):
Do not change `CHANGELOG.md`

caomaocao:
Fixed in 1967561d.

`test_plain_text_response` now sets `text_block.type = "text"` as requested, and I swept the other Anthropic response mocks in `tests/ai/test_ai_functions.py` so each one sets every attribute the production code reads:

- `TestAnthropicStructuredOutput::test_structured_output_extracts_tool_use_block` — `tool_block.type = "tool_use"` was already set; added realistic integer `usage` token counts.
- `TestAnthropicStructuredOutput::test_plain_text_response` — added `text_block.type = "text"` and integer `usage`.
- `TestAnthropicMultimodal::test_multimodal_prompt_assembly` — added `text_block.type = "text"` and integer `usage`.
- `TestAnthropicMultimodal::test_dict_with_role_as_complete_message` — added `text_block.type = "text"` and integer `usage`.

The `usage` values matter beyond realism: `prompt()` feeds `usage.input_tokens` / `usage.output_tokens` into the global token-metrics counters, so a bare `MagicMock` usage was leaking mock objects into that shared state.

Verified with the Anthropic extra (anthropic 0.117.0) installed:

```
python -m pytest tests/ai/test_ai_functions.py -k anthropic
```

Before the fix, `test_plain_text_response` failed with `AssertionError: assert None == 'Hello world'` (exactly the failure you described); after, all Anthropic tests pass. (`test_process_ndarray` is unrelated — it needs Pillow, which my local env lacks; it passes in CI.)

No production code or CHANGELOG changes in this commit.

caomaocao:
CI note: lint/format/type checks pass after syncing the branch with `main` (e4e876b9). The remaining red job fails on `tests/fast/test_ray_cpp_bindings.py::test_ray_worker_manager_snapshot_refresh_shutdown_has_no_deadlock` — the concurrency test added on `main` today in #205, unrelated to this PR (no Ray code touched; the same assertion failed 11 times across four of my open PRs today in random matrix jobs). All Anthropic tests pass. Happy to retrigger on request.

caomaocao:
Rebased onto current `main` (includes the #229 stabilization) per your request — the branch is now two clean commits (1bd595c8 production fix + 55c32544 test-mock fix; the zero-sum CHANGELOG add/revert pair from earlier history was dropped). All checks green.

vane/ai/providers/anthropic.py
```diff
@@ -317,5 +317,8 @@ def _flush_content() -> None:
             return None
 
         if response.content:
-            return response.content[0].text
+            return next(
+                (block.text for block in response.content if getattr(block, "type", None) == "text"),
+                None,
+            )
```
kaka11chen: The type-based filtering is the right fix for leading thinking blocks. Could we concatenate all text blocks instead of returning only the first one? The Anthropic SDK helper follows that behavior, and it avoids silently truncating compatible or future responses containing multiple text blocks. This is non-blocking for the response shape covered by #140.

```suggestion
            text_blocks = [
                block.text
                for block in response.content
                if getattr(block, "type", None) == "text"
            ]
            return "".join(text_blocks) if text_blocks else None
```

If we take this suggestion, please also cover `[thinking, text("a"), text("b")] -> "ab"`.

kaka11chen:
@codex review

vane/ai/providers/anthropic.py
caomaocao: Applied in f7910f02: `prompt()` now joins all text blocks (`"".join(text_blocks) if text_blocks else None`), and `tests/fast/test_anthropic_prompter_content.py` covers `[thinking, text("a"), text("b")] -> "ab"`. PR title/body updated to match the concatenation behavior.

kaka11chen (approved):
LGTM

Issue #152, opened 2026-07-21
[Enhancement][AI] Provider contract improvements benchmarked against Vercel AI SDK


caomaocao:
## Context

A design review of the AI provider abstraction (reviewed commit c43bd706) benchmarked against the Vercel AI SDK's provider architecture (`ai` v7.0.33, spec `LanguageModelV4`/`EmbeddingModelV4`, verified from source). The concrete bugs found are filed separately (#139 #140 #141 #142 #143 #144 #145 #146 #147 #148 #150 #151); this issue tracks the structural gaps that keep producing them.

## Proposed contract improvements

1. **Unified error taxonomy + core-owned retry.** Today retry policy differs per provider (Google: RetryAfterError + 3 retries; OpenAI: bare ValueError, SDK-internal retries only; Anthropic: nothing — see #148). AI SDK model: providers only *classify* (`APICallError` with `statusCode`/`isRetryable`, default retryable = 408/409/429/5xx, standardized `retry-after`/`retry-after-ms` header parsing); the orchestration layer owns the single backoff loop. Vane already has a sound retry loop — it lacks the error classification protocol.

2. **Capability declaration on descriptors, chunking in the core.** AI SDK embedding models declare `maxEmbeddingsPerCall` / `supportsParallelCalls`; `embedMany` chunks and parallelizes centrally. Vane's batching is ad-hoc per provider (OpenAI: internal token-based batching with its own bugs; Google: none at all — #144, #145). Add e.g. `max_inputs_per_call` / `max_tokens_per_call` to the Descriptor contract and centralize splitting in `_EmbedTextBatch`.

3. **Warnings channel instead of silent drops.** AI SDK makes `warnings` a required field on every result — "unsupported setting dropped" must surface. The largest bug family in this review is silent option dropping (#142, #146, #147). Either reject unknown options loudly or add a warnings channel through the UDF results.

4. **Usage returned in-band; driver-side aggregation.** Usage should be part of each call's result (AI SDK: structured usage on every `doGenerate`/`doEmbed`), aggregated by the orchestration layer — fixing the driver-side zero-metrics problem (#150) properly.

5. **Shared message normalization.** AI SDK maps user prompts to a standardized internal prompt type; providers map only from that. Vane's five prompters each reimplement multimodal/message handling with drifting semantics (role flattening in Google #145, repr-stringification and dropped image columns in vLLM #147).

[...]

Issue #145, opened 2026-07-21
[P2][AI] Google provider: stale embedding dims table with silent 768 fallback, no batch-size cap, role-tagged messages flattened


caomaocao:
## Problem

Three correctness hazards in `vane/ai/providers/google.py`:

1. **`_EMBEDDING_DIMS` contains only dead models; unknown models silently default to 768** (`google.py:79-82,173-177`). `gemini-embedding-001` outputs 3072 dims but is absent → `get_dimensions()` returns 768. In-repo consequence: `_embedding_zero_size` (`functions.py:347-366`) zero-fills failed rows with 768-wide vectors while successful rows are 3072-wide → **mixed-width FLOAT[] column** under `on_error="log"/"ignore"`. Also `if self.dimensions:` treats 0 as unset — use `is not None` like OpenAI.
2. **No batch-size cap on `embed_content`** (`google.py:236-252`): `get_udf_options` sets no `batch_size`, so whatever arrow batch arrives is forwarded as one API call; Google enforces a per-request input cap (order of 100–250 texts) → oversized batches fail outright. Unlike OpenAI there is no request-splitting at all.
3. **Role-tagged dict messages are flattened into a single user turn** (`google.py:396-403`): `{"role": "assistant", ...}` becomes a text part inside one `Content(role="user")` — few-shot conversations degrade to a concatenated user blob, and structured content stringifies as a Python repr. OpenAI/Anthropic providers append role dicts as genuine messages.

## Suggested fix

Add current models to the dims table and probe or error on unknown models instead of guessing; set a default batch_size (≤100) and/or chunk inside `embed_text`; build one `types.Content` per role dict.

Related: #141 (default models shut down).

Reviewed commit: c43bd706.

caomaocao:
I'll take this one (filed from my provider-layer review; assignment welcome — no triage rights to self-assign).

kaka11chen:
Consolidating the focused Google conversation-structure report from #102 here. The canonical role-handling acceptance scope should retain: map system instructions through the provider-supported field; preserve ordered user/model turns and mixed image/text parts; reject unsupported roles explicitly; and add structural recording-client tests for system/user, alternating turns, unsupported/tool roles, and image messages. #102 is being closed as absorbed by this issue.

kaka11chen:
[...]

PR #161, opened 2026-07-22
[fix](ai) Remove library-owned Google default models (#141)

### LINKED ISSUE #141: [P1][AI] Google provider default models are shut down
## Problem

Both Google defaults are retired models (`vane/ai/providers/google.py:93-94`):

- `DEFAULT_EMBED_MODEL = "text-embedding-004"` — shut down 2026-01-14
- `DEFAULT_PROMPT_MODEL = "gemini-2.0-flash"` — shut down 2026-06-01

(both listed with those shutdown dates on Google's Gemini API deprecations page).

## Failure scenario

Every `embed(provider="google")` / `prompt(provider="google")` call that omits `model` fails with a model-not-found error. The provider is unusable out of the box.

## Suggested fix

Bump defaults to current models (e.g. `gemini-embedding-001` and a current flash), and update `_EMBEDDING_DIMS` in the same change (`gemini-embedding-001` outputs 3072 dims; the current table only contains dead models — tracked separately with the other Google issues). Consider a policy note: default model ids are a liability (Vercel AI SDK avoids shipping model defaults; explicit model ids are the documented reproducibility recommendation in Vane's own docs).

Reviewed commit: c43bd706.

caomaocao:
## Summary

Reworked per maintainer direction on this PR: instead of bumping the retired Google default model IDs to newer ones, Vane stops choosing Google models altogether. No compatibility shim, deprecation window, legacy fallback, or automatic fallback model.

**Explicit-model contract.** `DEFAULT_EMBED_MODEL` / `DEFAULT_PROMPT_MODEL` are deleted. A model must be supplied per call (`model=...`) or through explicit user-owned provider configuration via new named constructor parameters: `GoogleProvider(prompt_model=..., embedding_model=..., embedding_dimensions=...)`. The constructor is now fully named (`name`, `api_key`, and the three config parameters), so a mistyped kwarg raises `TypeError` immediately instead of silently leaking into API request bodies. Generic `**options` passthrough for request options is unchanged on `get_prompter` / `get_text_embedder` and the expression-level option objects.

**Precedence chain.** One rule for models and dimensions: call-site argument > provider constructor configuration > `ValueError`. The error is raised at expression-build time (descriptor construction), not on a worker, and names both fix paths, e.g.:

> No prompt model configured for the Google provider. Pass model=... or configure GoogleProvider(prompt_model=...).

**Required descriptor model.** `model_name` is now a required field (no dataclass default) on `GoogleTextEmbedderDescriptor` and `GooglePrompterDescriptor`, so the provider constants and dataclass defaults cannot drift apart again.

**Model/option validation before dispatch.** A per-provider capability table

```python
_MODEL_UNSUPPORTED_OPTIONS: dict[str, frozenset[str]] = {
    "gemini-3.6-flash": frozenset({"temperature", "top_p", "top_k"}),
    "gemini-3.5-flash-lite": frozenset({"temperature", "top_p", "top_k"}),
}
```

plus `_validate_prompt_options(model_name, options)` runs in `GooglePrompterDescriptor.__post_init__`, so both the `get_prompter` path and direct descriptor construction reject unsupported combinations with a clear `ValueError` instead of forwarding parameters the Gemini API ignores today and will reject with HTTP 400 in future model generations ([latest-model migration guide](https://ai.google.dev/gemini-api/docs/latest-model) — both listed models are documented there). The table/helper shape is deliberately convention-aligned across hosted providers so a later lift into a core capability framework (#152) is mechanical.

**Embedding dimensions (absorbs #145 item 1 per maintainer comment).** The silent `_EMBEDDING_DIMS.get(model, 768)` fallback is deleted. Dimensions resolve as: explicit `dimensions` (call-site, else provider `embedding_dimensions`) > trusted per-model metadata > `ValueError`. The metadata table now carries only current models (`gemini-embedding-001` and `gemini-embedding-2`, both 3072-dim by default per the [embeddings guide](https://ai.google.dev/gemini-api/docs/embeddings)); the retired `text-embedding-004` / `embedding-001` entries are gone. Explicit dimensions outside the documented 128–3072 range for a known model, or non-positive values, are rejected at build time. The `if self.dimensions:` 0-is-falsy bug is fixed (`is not None`), so `dimensions=0` can never again silently degrade to a 768-wide guess.

**Review round (2026-07-28).** Addresses the three review findings:

1. **Row preservation (blocker).** `embed_text` now sends each input as its own `types.Content` — a multi-input `contents` list would be aggregated into a single embedding by `gemini-embedding-2` — and raises `ValueError` if the API returns a different number of embeddings than inputs.
2. **Resource-name canonicalization.** All local capability/metadata lookups strip the `models/` prefix, so `models/gemini-3.6-flash` can no longer bypass the sampling-parameter rejection and `models/gemini-embedding-001` hits the trusted dimensions table; the caller-provided name is still sent to the SDK verbatim.
3. **Dimensions type validation.** `dimensions` must be a non-boolean integer: `256.5`, `True`, or `"256"` raise the promised configuration `ValueError` before the bounds check.

Model resolution additionally uses explicit `is not None` precedence and rejects blank/whitespace model IDs at expression-build time (matching the same hardening requested on #168); the unreachable options-`model` branch is removed.

## Related issue

close: #141

Also implements the 768-fallback item of #145 (maintainer asked for it to ride in this PR); the remaining #145 items (batch cap, role-structured conversations) follow in a stacked PR.

## Documentation impact

      `AstroVela/vane-website`.

Provider and function docstrings in this repo are updated to describe the explicit-model contract.

## Validation

- `pytest tests/ai/test_ai_functions.py tests/ai/test_descriptor_secret_redaction.py` (local stub harness): 308 passed, including 38 new tests; the only failures are the same environment-dependent ones as unmodified `main` (missing Pillow/duckdb/transformers/vllm locally).
- New coverage: missing-model failure (prompt + embed) with both fix paths asserted; explicit `model=` flow-through; provider-config flow-through; call-beats-config precedence (models and dimensions); `gemini-3.6-flash` / `gemini-3.5-flash-lite` × `temperature`/`top_p`/`top_k` rejected pre-dispatch (provider path and direct descriptor construction); the same options pass for a model absent from the table; known-model dimension derivation (3072); explicit-dimensions flow-through; `dimensions=0` and out-of-documented-range dimensions rejected; unknown model without dimensions rejected; constructor typo raises `TypeError`; `model_name` required on both descriptors; two-input row preservation via separate `Content`s and count-mismatch rejection; both bare and `models/`-prefixed forms for capability and dimensions lookups; non-integer dimensions (`256.5`/`True`/`"256"`) rejected; blank/whitespace model rejection (call-site and provider-level, prompt + embed).
- `ruff check` / `ruff format` via pre-commit: passed.
- Model facts verified against live Google docs: [latest-model migration guide](https://ai.google.dev/gemini-api/docs/latest-model) (deprecated sampling parameters, affected models, future HTTP 400), [embeddings guide](https://ai.google.dev/gemini-api/docs/embeddings) (current models, 3072 default, 128–3072 supported output dimensionality).

## Checklist

Breaking changes (intentional, per maintainer direction — no backward compatibility):

- Google calls without an explicit model now raise `ValueError` at expression-build time.
- The `GoogleProvider` constructor no longer accepts arbitrary kwargs (`TypeError` on unknown names); request options belong on the call (`get_prompter(...)` / `get_text_embedder(...)` / option objects), where generic passthrough is unchanged.
- Unknown embedding models require explicit `dimensions`; known models validate the requested dimensionality against documented bounds.

https://claude.ai/code/session_01CXGCZZzpBqP7D1kivLJABc

### DIFF
--- a/vane/ai/providers/google.py
+++ b/vane/ai/providers/google.py
@@ -7,6 +7,12 @@ Supports text embedding via ``embed_content`` and prompting via
 ``generate_content`` with multimodal input (text + images) and
 structured output via ``response_schema``.
 
+Vane does not choose Google models: every prompt or embedding call must
+name a model, either per call (``model=...``) or through explicit
+provider configuration (``GoogleProvider(prompt_model=...,
+embedding_model=...)``). Missing configuration raises :class:`ValueError`
+at expression-build time.
+
 Requires::
 
     pip install 'vane-ai[google]'
@@ -25,6 +31,8 @@ from vane.ai.provider import Provider, ProviderImportError
 from vane.ai.typing import EmbeddingDimensions, UDFOptions
 
 if TYPE_CHECKING:
+    from collections.abc import Mapping
+
     from vane.ai.protocols import Prompter, TextEmbedder
     from vane.ai.typing import Embedding, Options
 
@@ -76,10 +84,71 @@ def _raise_retry_after_on_google_error(exc: Exception) -> None:
 # Model metadata
 # ---------------------------------------------------------------------------
 
+# Default output dimensionality per known embedding model, from the Gemini
+# embeddings guide (https://ai.google.dev/gemini-api/docs/embeddings). Only
+# models with trusted metadata belong here; any other model requires the
+# caller to supply ``dimensions`` explicitly.
 _EMBEDDING_DIMS: dict[str, int] = {
-    "text-embedding-004": 768,
-    "embedding-001": 768,
+    "gemini-embedding-001": 3072,
+    "gemini-embedding-2": 3072,
 }
+
+# Documented ``output_dimensionality`` bounds per known embedding model.
+_EMBEDDING_DIM_RANGE: dict[str, tuple[int, int]] = {
+    "gemini-embedding-001": (128, 3072),
+    "gemini-embedding-2": (128, 3072),
+}
+
+# Request options rejected per model before dispatch. Gemini 3.6 Flash and
+# 3.5 Flash-Lite deprecate the classic sampling parameters: the API ignores
+# them today and returns HTTP 400 in future model generations
+# (https://ai.google.dev/gemini-api/docs/latest-model).
+_MODEL_UNSUPPORTED_OPTIONS: dict[str, frozenset[str]] = {
+    "gemini-3.6-flash": frozenset({"temperature", "top_p", "top_k"}),
+    "gemini-3.5-flash-lite": frozenset({"temperature", "top_p", "top_k"}),
+}
+
+
+def _validate_prompt_options(model_name: str, options: Mapping[str, Any]) -> None:
+    """Reject request options the selected model is documented not to support.
+
+    Raises:
+        ValueError: If ``options`` contains keys listed for ``model_name``
+            in :data:`_MODEL_UNSUPPORTED_OPTIONS`.
+    """
+    unsupported = _MODEL_UNSUPPORTED_OPTIONS.get(model_name, frozenset())
+    offending = sorted(unsupported.intersection(options))
+    if offending:
+        raise ValueError(
+            f"Google model {model_name!r} does not support options {offending}: "
+            "the Gemini API ignores these sampling parameters and rejects them "
+            "in future model generations. Remove them from the request."
+        )
+
+
+def _validate_embedding_dimensions(model_name: str, dimensions: int | None) -> None:
+    """Validate the embedding-dimensions configuration for ``model_name``.
+
+    Raises:
+        ValueError: If ``dimensions`` is omitted for a model without trusted
+            metadata in :data:`_EMBEDDING_DIMS`, is not a positive integer,
+            or falls outside the documented range for a known model.
+    """
+    if dimensions is None:
+        if model_name not in _EMBEDDING_DIMS:
+            raise ValueError(
+                f"Cannot derive embedding dimensions for Google model {model_name!r}. "
+                "Pass dimensions=... or configure GoogleProvider(embedding_dimensions=...)."
+            )
+        return
+    if dimensions < 1:
+        raise ValueError(f"Embedding dimensions must be a positive integer, got {dimensions!r}.")
+    bounds = _EMBEDDING_DIM_RANGE.get(model_name)
+    if bounds is not None and not bounds[0] <= dimensions <= bounds[1]:
+        raise ValueError(
+            f"Google model {model_name!r} supports output dimensionality "
+            f"between {bounds[0]} and {bounds[1]}, got {dimensions}."
+        )
 
 
 # ---------------------------------------------------------------------------
@@ -88,15 +157,46 @@ _EMBEDDING_DIMS: dict[str, int] = {
 
 
 class GoogleProvider(Provider):
-    """Provider backed by Google Generative AI (Gemini)."""
+    """Provider backed by Google Generative AI (Gemini).
 
-    DEFAULT_EMBED_MODEL = "text-embedding-004"
-    DEFAULT_PROMPT_MODEL = "gemini-2.0-flash"
+    The provider ships no default model IDs: a model must be supplied per
+    call (``model=...``) or through the named constructor parameters below.
+    Call-site arguments always win over constructor configuration.
+
+    Args:
+        name: Optional display-name override (default ``"google"``).
+        api_key: Google API key. Prefer the ``GOOGLE_API_KEY`` environment
+            variable over passing keys in code.
+        prompt_model: Model used by ``get_prompter`` when the call does not
+            pass ``model=...``.
+        embedding_model: Model used by ``get_text_embedder`` when the call
+            does not pass ``model=...``.
+        embedding_dimensions: Embedding output dimensionality used when the
+            call does not pass ``dimensions=...``. Required (here or per
+            call) for embedding models without trusted dimension metadata.
+
+    All parameters are named; a mistyped keyword raises :class:`TypeError`
+    instead of silently leaking into API request options.
+    """
+
     _CLIENT_KEYS: ClassVar[frozenset[str]] = frozenset({"api_key"})
 
-    def __init__(self, name: str | None = None, **options: Any):
+    def __init__(
+        self,
+        name: str | None = None,
+        *,
+        api_key: str | None = None,
+        prompt_model: str | None = None,
+        embedding_model: str | None = None,
+        embedding_dimensions: int | None = None,
+    ):
         self._name = name or "google"
-        self._options: dict[str, Any] = options
+        self._prompt_model = prompt_model
+        self._embedding_model = embedding_model
+        self._embedding_dimensions = embedding_dimensions
+        self._options: dict[str, Any] = {}
+        if api_key is not None:
+            self._options["api_key"] = api_key
 
     @property
     def name(self) -> str:
@@ -114,12 +214,27 @@ class GoogleProvider(Provider):
         dimensions: int | None = None,
         **options: Any,
     ) -> TextEmbedderDescriptor:
+        """Build an embedder descriptor for an explicitly selected model.
+
+        Raises:
+            ValueError: If neither ``model=...`` nor the provider's
+                ``embedding_model`` is configured, or if dimensions cannot
+                be resolved for the selected model.
+        """
         provider_options, embed_options = self._split_options(options)
-        model_name = model or self.DEFAULT_EMBED_MODEL
+        options_model = embed_options.pop("model", None)
+        model_name = model or options_model or self._embedding_model
+        if model_name is None:
+            raise ValueError(
+                "No embedding model configured for the Google provider. "
+                "Pass model=... or configure GoogleProvider(embedding_model=...)."
+            )
+        if dimensions is None:
+            dimensions = self._embedding_dimensions
         return GoogleTextEmbedderDescriptor(
+            model_name=model_name,
             provider_name=self._name,
             provider_options=provider_options,
-            model_name=model_name,
             dimensions=dimensions,
             embed_options=embed_options,
         )
@@ -131,11 +246,25 @@ class GoogleProvider(Provider):
         return_format: Any | None = None,
         **options: Any,
     ) -> PrompterDescriptor:
+        """Build a prompter descriptor for an explicitly selected model.
+
+        Raises:
+            ValueError: If neither ``model=...`` nor the provider's
+                ``prompt_model`` is configured, or if the selected model
+                does not support one of the requested options.
+        """
         provider_options, prompt_options = self._split_options(options)
+        options_model = prompt_options.pop("model", None)
+        model_name = model or options_model or self._prompt_model
+        if model_name is None:
+            raise ValueError(
+                "No prompt model configured for the Google provider. "
+                "Pass model=... or configure GoogleProvider(prompt_model=...)."
+            )
         return GooglePrompterDescriptor(
+            model_name=model_name,
             provider_name=self._name,
             provider_options=provider_options,
-            model_name=model or prompt_options.pop("model", self.DEFAULT_PROMPT_MODEL),
             system_message=system_message,
             return_format=return_format,
             prompt_options=prompt_options,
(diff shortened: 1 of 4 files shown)


vane/ai/providers/google.py
```diff
@@ -90,8 +92,8 @@ def _raise_retry_after_on_google_error(exc: Exception) -> None:
 class GoogleProvider(Provider):
     """Provider backed by Google Generative AI (Gemini)."""
 
-    DEFAULT_EMBED_MODEL = "text-embedding-004"
-    DEFAULT_PROMPT_MODEL = "gemini-2.0-flash"
+    DEFAULT_EMBED_MODEL = "gemini-embedding-2"
```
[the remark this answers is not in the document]
caomaocao: Fixed in 7be8f3c8 — the default is now `gemini-embedding-001`. Confirmed against the embeddings guide: `gemini-embedding-2` aggregates a multi-input `contents` list into a single embedding, which breaks the one-embedding-per-row contract given that `GoogleTextEmbedder.embed_text` sends the whole batch as `contents`; `gemini-embedding-001` embeds each input separately, is listed Stable, and shares the 3072-dim default so dimensions are unchanged. The `gemini-embedding-2` dims-table entry stays for explicit-model callers.

vane/ai/providers/google.py
```diff
@@ -153,7 +155,7 @@ class GoogleTextEmbedderDescriptor(TextEmbedderDescriptor):
 
     provider_name: str = "google"
     provider_options: dict[str, Any] = field(default_factory=dict)
-    model_name: str = "text-embedding-004"
+    model_name: str = "gemini-embedding-2"
```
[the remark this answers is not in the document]
caomaocao: Fixed in 7be8f3c8 by the same default switch to `gemini-embedding-001`, which supports `task_type` (the guide says Embedding 2 does not — task instructions go in the prompt instead), so existing `GoogleEmbeddingOptions(task_type=...)` callers relying on the default keep working.

vane/ai/providers/google.py
```diff
@@ -90,8 +92,8 @@ def _raise_retry_after_on_google_error(exc: Exception) -> None:
 class GoogleProvider(Provider):
     """Provider backed by Google Generative AI (Gemini)."""
 
-    DEFAULT_EMBED_MODEL = "text-embedding-004"
-    DEFAULT_PROMPT_MODEL = "gemini-2.0-flash"
+    DEFAULT_EMBED_MODEL = "gemini-embedding-2"
+    DEFAULT_PROMPT_MODEL = "gemini-3.6-flash"
```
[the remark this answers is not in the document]
caomaocao: Deliberately keeping `gemini-3.6-flash` with no code change here. Per the linked latest-model guide, temperature/top_p/top_k are deprecated starting with Gemini 3.6 Flash / 3.5 Flash-Lite "and all future Gemini model releases", so no current-or-future default escapes this — staying on an older model only postpones it, and `gemini-2.5-flash` already has a scheduled shutdown (2026-10-16). Filtering the params client-side would be exactly the silent-drop antipattern this project is eliminating; proper surfacing (warn/reject on unsupported options) belongs to the option-handling/warnings work tracked in #145/#152. 7be8f3c8 adds a docstring note on `GoogleProvider` recording the deprecation and the tracking issues.

kaka11chen (approved):
LGTM

kaka11chen:
@codex review

kaka11chen (changes requested):
Maintainer direction: please revise this PR to remove the library-owned Google model defaults instead of replacing retired IDs with newer IDs.

Vane is still at an early stage, so this change does not need backward compatibility. Please do not add a compatibility shim, deprecation window, legacy fallback, or automatic fallback model.

Requested behavior:

Require an explicit Google prompt/embedding model, supplied either by the call (model=...) or by explicit user-owned provider configuration.
If no model is configured, fail fast with a clear configuration error.
Make the Google descriptor model_name required as well, so the provider constant and dataclass default cannot drift again.
Do not silently drop unsupported options. Validate the selected model plus options before dispatch; for example, reject temperature, top_p, and top_k for gemini-3.6-flash with a clear error.
For embeddings, require dimensions when they cannot be derived safely from the explicitly selected model.
Update tests to cover missing-model failure, explicit model flow-through, and unsupported model/option combinations.
This should replace the current default-bump approach, not be layered on top of it. We can apply the same no-library-default policy to other hosted providers separately.

caomaocao:
Implemented as directed in 498b9920 — the branch was reset to `main` and rewritten from scratch; the earlier default-bump commits are gone.

- **No library-owned defaults.** `DEFAULT_EMBED_MODEL` / `DEFAULT_PROMPT_MODEL` are deleted. A model comes from the call (`model=...`) or from explicit user-owned provider configuration (`GoogleProvider(prompt_model=..., embedding_model=...)`), with call-site winning. No shim, deprecation window, or fallback model.
- **Fail fast on missing configuration.** A missing model raises `ValueError` at expression-build time (descriptor construction, before anything ships to a worker), and the message names both fix paths, e.g. "No prompt model configured for the Google provider. Pass model=... or configure GoogleProvider(prompt_model=...)."
- **Descriptor `model_name` required.** Both Google descriptors now declare `model_name` as a required dataclass field, so the constant/default drift class of bug is gone.
- **No silent option drops.** A per-model capability table (`_MODEL_UNSUPPORTED_OPTIONS`) is validated in `__post_init__`: `temperature`/`top_p`/`top_k` are rejected with a clear `ValueError` for `gemini-3.6-flash` and `gemini-3.5-flash-lite` (both documented as deprecating these parameters in the latest-model migration guide, which also states future generations return HTTP 400).
- **Dimensions.** The silent 768 fallback is deleted (this also covers item 1 of #145 as you requested there): explicit `dimensions` (call or provider config) > trusted metadata for the selected model (`gemini-embedding-001` / `gemini-embedding-2`, 3072) > `ValueError`. Out-of-documented-range values for known models are rejected, and the `dimensions=0`-is-falsy bug is fixed.
- **Tests.** Added coverage for missing-model failure, explicit-model and provider-config flow-through, precedence, unsupported model/option combinations (both entry points), the dimensions matrix, and constructor-typo `TypeError` (the constructor is fully named now, so typos cannot leak into request options).

The capability table and validation helper deliberately follow a shape that can be lifted into a core, provider-agnostic framework later under #152 once this pattern has been validated here and on the Anthropic side.

vane/ai/providers/google.py
```diff
@@ -88,10 +90,16 @@ def _raise_retry_after_on_google_error(exc: Exception) -> None:
...
-    DEFAULT_EMBED_MODEL = "text-embedding-004"
-    DEFAULT_PROMPT_MODEL = "gemini-2.0-flash"
+    Note: Google deprecates the ``temperature``, ``top_p``, and ``top_k`` sampling
+    parameters starting with Gemini 3.6 Flash and all future Gemini releases
+    (currently ignored, later HTTP 400); surfacing this to callers is tracked in
+    vane#145/vane#152.
+    """
+
+    DEFAULT_EMBED_MODEL = "gemini-embedding-001"
+    DEFAULT_PROMPT_MODEL = "gemini-3.6-flash"
```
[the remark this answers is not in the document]
caomaocao: Resolved by the rework in 498b9920, in the direction the maintainer requested. The scenario this comment describes — a caller omitting `model` and getting a default whose docstring admits it ignores `temperature`/`top_p`/`top_k` — no longer exists, because there is no default model to fall back to: omitting `model` now raises a configuration `ValueError` at expression-build time. And for callers who explicitly select `gemini-3.6-flash` (or `gemini-3.5-flash-lite`) while passing any of those sampling options, the combination is rejected pre-dispatch with a `ValueError` naming the offending keys, instead of being forwarded for the API to ignore now and reject with HTTP 400 later. The options remain public in `vane/ai/options.py` and keep working for models that support them.

kaka11chen:
@codex review

vane/ai/providers/google.py
```diff
@@ -76,27 +84,119 @@ def _raise_retry_after_on_google_error(exc: Exception) -> None:
...
 
+# Default output dimensionality per known embedding model, from the Gemini
+# embeddings guide (https://ai.google.dev/gemini-api/docs/embeddings). Only
+# models with trusted metadata belong here; any other model requires the
+# caller to supply ``dimensions`` explicitly.
 _EMBEDDING_DIMS: dict[str, int] = {
-    "text-embedding-004": 768,
-    "embedding-001": 768,
+    "gemini-embedding-001": 3072,
+    "gemini-embedding-2": 3072,
```
kaka11chen, line 93: **Blocker — keep `gemini-embedding-2` row-preserving**

Adding this model to the trusted metadata makes descriptor construction succeed, but `GoogleTextEmbedder.embed_text()` still sends the whole `list[str]` as one `contents` value. Google documents that Embedding 2 aggregates multiple direct inputs into one embedding, while `_EmbedTextBatch` requires one result per input row: https://ai.google.dev/gemini-api/docs/embeddings#generating-embeddings

For any batch larger than one, this can therefore return too few vectors and break row preservation. Please encode each input as a separate `types.Content` (preferred), temporarily reject this model, or at minimum force single-input calls. The runtime should also verify `len(result.embeddings) == len(text)`, with a regression test covering two inputs.

vane/ai/providers/google.py
```diff
@@ -76,27 +84,119 @@ def _raise_retry_after_on_google_error(exc: Exception) -> None:
...
+
+
+def _validate_prompt_options(model_name: str, options: Mapping[str, Any]) -> None:
+    """Reject request options the selected model is documented not to support.
+
+    Raises:
+        ValueError: If ``options`` contains keys listed for ``model_name``
+            in :data:`_MODEL_UNSUPPORTED_OPTIONS`.
+    """
+    unsupported = _MODEL_UNSUPPORTED_OPTIONS.get(model_name, frozenset())
```
kaka11chen: **Normalize model resource names before capability lookup**

The Google Gen AI SDK accepts both a bare model ID such as `gemini-3.6-flash` and the full Gemini API form `models/gemini-3.6-flash`. This exact-key lookup lets the valid prefixed form bypass the unsupported-option check and forward `temperature` / `top_p` / `top_k`. The same mismatch also prevents `models/gemini-embedding-001` from using the trusted dimensions metadata.

Please derive a canonical ID for all local metadata/capability lookups, for example with `model_name.removeprefix("models/")`, while preserving the caller-provided value for the SDK request. Tests should cover both accepted forms.

vane/ai/providers/google.py
```diff
@@ -76,27 +84,119 @@ def _raise_retry_after_on_google_error(exc: Exception) -> None:
...
+            or falls outside the documented range for a known model.
+    """
+    if dimensions is None:
+        if model_name not in _EMBEDDING_DIMS:
+            raise ValueError(
+                f"Cannot derive embedding dimensions for Google model {model_name!r}. "
+                "Pass dimensions=... or configure GoogleProvider(embedding_dimensions=...)."
+            )
+        return
+    if dimensions < 1:
```
kaka11chen, line 157: **Validate the dimensions type, not only its sign**

The documented contract and error both require a positive integer, but this condition accepts values such as `256.5`; for an untabled model, `True` also passes as dimension 1. A string instead raises an incidental `TypeError` from the comparison rather than the promised configuration `ValueError`. These values then reach `EmbeddingDimensions` / PyArrow or the Google SDK.

Please explicitly require a non-boolean integral value before the bounds check (and normalize it to `int` if supporting integral scalar types), with regression coverage for a float, a boolean, and a non-numeric value.

vane/ai/providers/google.py
caomaocao: Fixed in 28342978 with the preferred encoding: each input is now its own `types.Content`, and `embed_text` verifies `len(result.embeddings) == len(text)`, raising `ValueError` naming the model on a mismatch. New tests cover two inputs → two vectors via two separate `Content`s (asserting the exact `contents` structure sent) and a count-mismatch → `ValueError` regression. The stacked #228 (batch cap) is rebased on top: with per-`Content` encoding its one-input-per-request carve-out for aggregating models became unnecessary, so it now chunks every model at the 100-input cap.

vane/ai/providers/google.py
caomaocao: Fixed in 28342978: `_canonical_model_id()` (`removeprefix("models/")`) is applied in `_validate_prompt_options`, `_validate_embedding_dimensions`, and `get_dimensions`, while the caller-provided name still goes to the SDK verbatim. Tests cover both accepted forms: `models/gemini-3.6-flash` × `temperature` rejected, and `models/gemini-embedding-001` deriving the trusted 3072 dimensions.

vane/ai/providers/google.py
caomaocao: Fixed in 28342978: `_validate_embedding_dimensions` now requires a non-boolean `int` before the bounds check, so `256.5`, `True`, and `"256"` all raise the promised configuration `ValueError` instead of passing or raising an incidental `TypeError`. A parametrized regression covers exactly those three.

caomaocao:
All three findings are addressed in 28342978: row-preserving embedding via one `types.Content` per input plus a returned-count check, `models/` prefix canonicalization for capability/dimension lookups, and strict non-boolean-integer `dimensions` validation. The precedence/blank-model hardening requested on #168 is applied to both Google resolution paths as well, keeping the providers parallel. The stacked #228 is rebased on top (54d8857f). Local stub-harness suite is at the known-environment-failure baseline with 15 new tests; CI running. Ready for another look.

kaka11chen:
@codex review

kaka11chen (approved):
LGTM

Issue #147, opened 2026-07-21
[P2][AI] Execution options are silently dropped or misrouted across API surfaces


caomaocao:
## Problem

A family of silent option drops/misroutes where the SQL, expression, and relation surfaces diverge without any diagnostic:

1. **Relation `prompt()` drops user `batch_size`** (`vane/ai/functions.py:878-885`): API prompter descriptors never read `batch_size` from their options (`providers/openai.py:376-383`, `anthropic.py:110-117`, `google.py:288-295`), so `batch_size=udf_opts.batch_size or 1` always yields 1 → one API call per `asyncio.run`, serialized per worker. Consequence: the `asyncio.gather`/semaphore machinery never runs >1 request on the relation path, making the per-provider `max_api_concurrency` defaults (OpenAI 32 / Anthropic 16 / Google 16) dead code there — while the SQL surface honors the same option (`_sql.py:181`, test-pinned). Same for `GoogleTextEmbedderDescriptor`.
2. **Bare `concurrency=` kwarg is a no-op on the Python path**: only `actor_number` is read; the SQL layer and options objects do map `concurrency → actor_number`, making the Python-kwarg path uniquely silent.
3. **Incoherent GPU guard across surfaces** (`functions.py:192-194` vs `:646-647`): `provider_options=OpenAIProviderOptions(concurrency=4)` on the relation path raises `UDFOptions.num_gpus is required when actor_number is set` — for a pure HTTP provider — while the expression API accepts the identical options object.
4. **`return_format` smuggled via `prompt_options` dict** (`functions.py:964-972`): the expression-API guard only rejects the top-level kwarg; a dict-borne `return_format` reaches the prompter, which returns parsed objects while `_PromptBatch` skips serialization → opaque `ArrowInvalid` in the worker instead of a clean TypeError.
5. **`OpenAIProvider` lacks the `_split_options` pattern** (`providers/openai.py:112-135,187-189`): all provider-constructor options are copied into `provider_options`, so provider-level defaults (`batch_size=`, `max_api_concurrency=`, `on_error=`) are silently ignored (inconsistent with `TransformersProvider`, test-pinned), and `get_dimensions()` passes them verbatim to `OpenAI(**provider_options)` → `TypeError` for direct consumers.
[...]

PR #170, opened 2026-07-22
[fix](ai) Preserve RetryAfterError pickling; log ignored substitutions; unify Retry-After handling

### LINKED ISSUE #148: [P2][AI] Retry and on_error handling drifts per provider; chunked embedding bypasses both
## Problem

Retry/error semantics are inconsistent across providers and have concrete gaps:

1. **Chunked embedding bypasses retry and `on_error` entirely** (`vane/ai/functions.py:420-421,461-463`): with `max_chunk_chars` set, `embed_text` is called directly — no `_retry_call`, no `on_error`, no zero-fill. A single 429 raises straight out of the actor despite `max_retries=3, on_error="ignore"`. The non-chunked branch honors both.
2. **`on_error="log"` never logs** (`functions.py:148-153,179-183`): "log" and "ignore" are byte-for-byte identical; no logger exists in the module.
3. **Batch-granularity failure substitution** (`functions.py:429-430,509-510,585-586`): one bad row zero-fills/nulls the entire batch; zero vectors look like valid embeddings to downstream cosine/similarity code.
4. **Per-provider retry drift (undocumented)**: Google gets `RetryAfterError` + 3 wrapper retries; OpenAI wraps all `OpenAIError`s as bare `ValueError` (`providers/openai.py:338-339`), discarding status/Retry-After, with wrapper `max_retries=0` (SDK-internal retries only); Anthropic gets nothing. The same 429 behaves three different ways.
5. **`RetryAfterError` does not survive pickling** (`functions.py:99-110`): default exception reduce reconstructs from `args=(str(original),)` → `retry_after` becomes a string and `__cause__` is lost across Ray/subprocess boundaries.

## Suggested fix

Route the chunked path through `_retry_call`; emit `logger.warning` on `on_error="log"`; split-batch or per-row retry before substituting (prefer NULL over zeros, or document zero-fill); introduce a unified provider error type carrying `status_code`/`is_retryable`/`retry_after` so the core retry loop owns policy uniformly (Vercel AI SDK model: providers classify via `APICallError.isRetryable` + standardized `retry-after` header parsing; the core owns the backoff loop — providers never retry); implement `RetryAfterError.__reduce__`.

Reviewed commit: c43bd706.

caomaocao:
## Summary

Rescoped per the maintainer's request (2026-08-05) to the still-open #148 scope: `RetryAfterError` serialization and the remaining classification/logging gaps. The branch was rebuilt from current `main`; the previously bundled #147 work (execution-option routing, silent-option handling) is dropped — #147 is closed, superseded by #397's API replacement. Of #148's original report, the chunked-embedding retry bypass and the batch-granularity substitution are already fixed on current `main` and are not re-fixed here.

- **Preserve `RetryAfterError` across pickling.** The default `BaseException` reduction rebuilds via `__init__(*args)`, feeding the sanitized message in as `retry_after` and regenerating a generic message, so a Ray worker → driver round trip degraded the sanitized summary. A custom `__reduce__` now rebuilds from the already-sanitized instance state (numeric `retry_after`, message, whitelisted status attributes) without re-running constructor sanitization or resurrecting the original provider error.
- **Make `on_error="ignore"` substitutions observable.** Every provider failure converted to NULL now emits exactly one bounded `WARNING`: exception class, sanitized numeric-status summary, and attempt count where a retry loop knows it — never prompt text, row payloads, option mappings, credentials, or tracebacks (the vane#105 posture). A sentinel default distinguishes an already-logged substituted NULL from a genuine provider NULL, so a structured-output contract violation on a provider NULL also warns instead of disappearing silently. `on_error="raise"` is byte-for-byte unchanged; no new `on_error` mode is introduced.
- **Unify Retry-After handling.** A shared provider-agnostic helper owns the 429/503 qualifying set, header lookup, the #469 malformed-value guards, and the 5-second default. The Google provider's extraction becomes a thin adapter over it (its extraction tests pass unmodified); OpenAI (embeddings, Chat Completions, Responses) and Anthropic (Messages) adopt the same helper, raising the converted error outside the SDK-error handler exactly as Google does so no raw provider error is retained as `__context__`. A server-requested `Retry-After` is now honored identically across providers instead of falling back to exponential backoff. Anthropic 529 overload responses deliberately stay on the existing transient-backoff path.

User-visible changes to call out for release notes: `on_error="ignore"` now logs one warning per substituted provider call, and OpenAI/Anthropic 429/503 retries wait the server-requested `Retry-After` (capped at 120s) instead of exponential backoff.

## Related issue

Addresses the remaining #148 scope as named in the rescope request; whether this fully closes #148 is left to maintainer judgment. The one remaining design-level item — a unified provider error carrier owning `is_retryable`/`status_code`/`retry_after` — is the #152 benchmark discussion and is intentionally not attempted here.

No overlap with #572: terminal-state validation leaves the retry helpers and the non-retryable classification tuple untouched, and this PR does not cross that seam from the other side; the two compose regardless of landing order.

## Documentation impact

      `AstroVela/vane-website`.

This is a provider retry-semantics correctness fix; the compatibility notes above stay in the PR body, per the #480 precedent.

## Validation

- `pre-commit` (ruff check, ruff format, mypy) passes on every commit.
- Targeted tests run locally under a package-stub harness (the native engine cannot be built on this dev machine): `tests/ai/test_ai_functions.py`, `tests/ai/test_prompt_provider_contracts.py`, `tests/fast/test_anthropic_prompter_content.py` — 247 passed, including the new regression coverage for pickling round-trips, substitution logging across both wrappers, shared-helper extraction, and per-provider Retry-After adoption. The 8 remaining local failures are pre-existing environment limits identical on `main` (missing `cloudpickle`; engine-dependent tests).
- Full-suite validation is CI-only for this branch; no local full-suite green is claimed.

## Checklist

      (Scope is exactly the three-item list from the rescope request; each behavior change carries regression tests.)
      licenses and are recorded where required. (None added.)
      and untrusted input have been considered. (The custom `__reduce__` rebuilds only from
      already-sanitized state and cannot resurrect the original provider error; substitution
      warnings emit class names and numeric statuses only.)
      changes passed relevant checks. (Python-only.)

### DIFF
--- a/vane/ai/functions.py
+++ b/vane/ai/functions.py
@@ -142,22 +142,29 @@ def _validate_on_error(on_error: object) -> _OnError:
     return cast(_OnError, on_error)
 
 
-def _log_substituted_failure(exc: Exception, *, on_error: _OnError) -> None:
+def _log_substituted_failure(exc: Exception, *, on_error: _OnError, attempts: int | None = None) -> None:
     """Record one bounded WARNING when a provider failure is replaced with NULL.
 
     Called at each site that swallows a provider error and yields NULL under
     ``on_error='ignore'``; a no-op under ``'raise'`` so callers cannot log a
-    failure they are about to re-raise. Only the exception class and its
-    sanitized numeric-status summary are emitted — never prompt text, row
-    payloads, option mappings, credentials, or a traceback (vane#105) — so a
-    silent data degradation becomes observable without leaking sensitive input.
+    failure they are about to re-raise. ``attempts`` is the number of tries made
+    when the caller is a retry loop (omitted where no attempt count is
+    meaningful). Only the exception class, its sanitized numeric-status summary,
+    and that count are emitted — never prompt text, row payloads, option
+    mappings, credentials, or a traceback (vane#105) — so a silent data
+    degradation becomes observable without leaking sensitive input.
     """
     if on_error == "raise":
         return
-    logger.warning(
-        "vane.ai substituted NULL for a failed provider call: %s",
-        _safe_original_error_summary(exc),
-    )
+    summary = _safe_original_error_summary(exc)
+    if attempts is not None:
+        logger.warning(
+            "vane.ai substituted NULL after %d attempt(s) for a failed provider call: %s",
+            attempts,
+            summary,
+        )
+    else:
+        logger.warning("vane.ai substituted NULL for a failed provider call: %s", summary)
 
 
 def _rebuild_retry_after_error(args: tuple[Any, ...], state: dict[str, Any]) -> RetryAfterError:
@@ -389,7 +396,9 @@ def _retry_call(
     """
     _validate_on_error(on_error)
     last_exc: Exception | None = None
+    attempts = 0
     for attempt in range(1 + max(0, max_retries)):
+        attempts = attempt + 1
         wait: float | None = None
         try:
             result = fn(*args, **kwargs)
@@ -426,7 +435,7 @@ def _retry_call(
     assert last_exc is not None
     if on_error == "raise":
         raise last_exc
-    _log_substituted_failure(last_exc, on_error=on_error)
+    _log_substituted_failure(last_exc, on_error=on_error, attempts=attempts)
     return default
 
 
@@ -448,7 +457,9 @@ async def _retry_call_async(
     """
     _validate_on_error(on_error)
     last_exc: Exception | None = None
+    attempts = 0
     for attempt in range(1 + max(0, max_retries)):
+        attempts = attempt + 1
         wait: float | None = None
         try:
             result = fn(*args, **kwargs)
@@ -474,7 +485,7 @@ async def _retry_call_async(
     assert last_exc is not None
     if on_error == "raise":
         raise last_exc
-    _log_substituted_failure(last_exc, on_error=on_error)
+    _log_substituted_failure(last_exc, on_error=on_error, attempts=attempts)
     return default
 
 
--- a/vane/ai/functions.py
+++ b/vane/ai/functions.py
@@ -148,8 +148,10 @@ def _log_substituted_failure(exc: Exception, *, on_error: _OnError, attempts: in
     """Record one bounded WARNING when a provider failure is replaced with NULL.
 
     Called at each site that swallows a provider error and yields NULL under
-    ``on_error='ignore'``; a no-op under ``'raise'`` so callers cannot log a
-    failure they are about to re-raise. ``attempts`` is the number of tries made
+    ``on_error='ignore'`` — including the native vLLM executor, whose ``"null"``
+    policy is the lowered form of ``'ignore'`` (see the mapping in
+    ``vane/ai/providers/vllm.py``); a no-op under ``'raise'`` so callers cannot
+    log a failure they are about to re-raise. ``attempts`` is the number of tries made
     when the caller is a retry loop (omitted where no attempt count is
     meaningful). Only the exception class, its sanitized numeric-status summary,
     and that count are emitted — never prompt text, row payloads, option
@@ -1360,9 +1362,10 @@ class _ValidateStructuredOutputBatch:
                 continue
             try:
                 results.append(self._return_format.validate_json(raw_text))
-            except OutputValidationError:
+            except OutputValidationError as exc:
                 if self._on_error == "raise":
                     raise
+                _log_substituted_failure(exc, on_error=self._on_error)
                 results.append(None)
         return pa.table({self._output_column: pa.array(results, type=pa.string())})
 
--- a/vane/ai/functions.py
+++ b/vane/ai/functions.py
@@ -167,6 +167,13 @@ def _log_substituted_failure(exc: Exception, *, on_error: _OnError, attempts: in
         logger.warning("vane.ai substituted NULL for a failed provider call: %s", summary)
 
 
+# Sentinel a retry-helper caller can pass as ``default`` to tell an
+# already-logged substituted failure apart from a genuine provider NULL —
+# the two must not be conflated, or a NULL that fails downstream result
+# validation would be swallowed without its own warning.
+_SUBSTITUTED_FAILURE = object()
+
+
 def _rebuild_retry_after_error(args: tuple[Any, ...], state: dict[str, Any]) -> RetryAfterError:
     """Reconstruct a pickled :class:`RetryAfterError` from its sanitized state.
 
@@ -355,17 +362,6 @@ def _retry_after_error(exc: Exception) -> RetryAfterError | None:
     if retry_after is None:
         retry_after = _DEFAULT_RETRY_AFTER_SECONDS
     return RetryAfterError(retry_after=retry_after, original=exc)
-
-
-def _raise_if_retry_after(exc: Exception) -> None:
-    """Re-raise a rate-limited provider error as a :class:`RetryAfterError` when
-    it qualifies (429/503); a no-op otherwise, so the caller re-raises the
-    original error. Lets every HTTP provider honour ``Retry-After`` uniformly
-    (issue #148) by delegating to :func:`_retry_after_error`.
-    """
-    retry_error = _retry_after_error(exc)
-    if retry_error is not None:
-        raise retry_error from None
 
 
 def _retry_call(
@@ -1245,6 +1241,7 @@ class _PromptBatch:
                     row_messages[index],
                     max_retries=max_retries,
                     on_error=on_error,
+                    default=_SUBSTITUTED_FAILURE,
                     on_awaitable=self._mark_loop_bound,
                 )
             except ProviderCapabilityError as exc:
@@ -1258,17 +1255,16 @@ class _PromptBatch:
             if provider_error is not None:
                 provider, model = _descriptor_identity(self._descriptor)
                 raise _safe_provider_execution_error(provider, model, "Prompt execution", provider_error) from None
+            if result is _SUBSTITUTED_FAILURE:
+                # The retry helper already logged this substitution before
+                # returning the sentinel; surface it as a NULL row.
+                return None
             try:
                 return self._validate_result(result)
             except (_ProviderResultError, OutputValidationError, RawResponseSerializationError) as exc:
                 if on_error == "raise":
                     raise
-                # A NULL reaching validation is either a legitimate provider NULL
-                # or an already-logged swallowed failure (the retry helper logged
-                # it before substituting None); only a non-NULL payload that fails
-                # the result contract is a fresh substitution to record here.
-                if result is not None:
-                    _log_substituted_failure(exc, on_error=on_error)
+                _log_substituted_failure(exc, on_error=on_error)
                 return None
 
         async def run_all(indices: list[int]) -> list[str | None]:
--- a/vane/ai/functions.py
+++ b/vane/ai/functions.py
@@ -284,6 +284,72 @@ def _is_transient_provider_error(exc: Exception) -> bool:
     return False
 
 
+_RATE_LIMIT_STATUS_CODES = frozenset({429, 503})
+_DEFAULT_RETRY_AFTER_SECONDS = 5.0
+
+
+def _provider_status_code(exc: Exception) -> int | None:
+    """Discover an HTTP status from a provider SDK error.
+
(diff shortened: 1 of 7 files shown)


vane/ai/typing.py
```diff
@@ -24,6 +25,43 @@
...
+def api_worker_options(options: Mapping[str, Any], *, default_batch_size: int | None = None) -> dict[str, Any]:
+    """Shared execution-option reads for pure-HTTP provider descriptors.
+
+    Pure HTTP providers need no GPU unless one is explicitly declared,
+    honour an explicit ``batch_size``, and accept ``concurrency`` as the
+    public alias for ``actor_number``. Returns keyword arguments for
+    :class:`UDFOptions`.
+    """
+    return {
+        "batch_size": options.get("batch_size", default_batch_size),
```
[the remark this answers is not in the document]
caomaocao: Fixed in ea9e0bcf: `api_worker_options` now int-coerces a non-None `batch_size` and raises `ValueError("batch_size must be a positive integer")` for values <= 0, mirroring `actor_number_from_options` and the SQL layer's `_int_or_none`; `None` still passes through. Covered by new red-run-verified tests in `tests/fast/test_ai_silent_options.py` (zero/negative rejected, positive and `None` still flow, float coerced).

vane/ai/providers/openai.py
```diff
@@ -129,18 +146,15 @@ def get_prompter(
...
             provider_name=self._name,
-            provider_options=provider_opts,
-            model_name=model or self.DEFAULT_PROMPTER_MODEL,
+            provider_options=provider_options,
+            model_name=model or prompt_options.pop("model", self.DEFAULT_PROMPTER_MODEL),
             system_message=system_message,
             return_format=return_format,
             use_chat_completions=use_chat_completions,
-            prompt_options=options,
+            prompt_options=prompt_options,
```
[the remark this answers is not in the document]
caomaocao: Fixed in ea9e0bcf: `get_prompter`/`get_text_embedder` now pop the descriptor-owned keys (`model`, `system_message`, `return_format`, `use_chat_completions`; `model`/`dimensions` for the embedder) out of the request options and use them as fallbacks when the call-level argument is omitted — explicit call arguments still win, and `instantiate()` no longer hits `TypeError: got multiple values for keyword argument`. `use_chat_completions` became `bool | None = None` at the provider method, resolving call-level -> provider-level -> `True`, so the pinned default is unchanged. New red-run-verified tests cover provider-level defaults being honored, call-level precedence, and instantiate not crashing; the descriptor repr/pickle redaction suites stay green.

vane/ai/functions.py
```diff
@@ -557,6 +681,27 @@ def _serialize_result(self, result: Any) -> str | None:
...
+        Each row gets a single attempt through ``prompt_batch`` — the batch
+        call already exhausted the retry budget — and only rows that fail
+        individually are substituted with ``None``. A malformed response
+        (wrong row count) is treated as that row's failure.
+        """
+        results: list[Any] = []
+        for text in texts:
+            row = _retry_call(
+                self._prompter.prompt_batch,
+                [text],
```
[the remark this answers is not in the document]
caomaocao: Good catch on the interaction — the root cause is the per-call `finished_submitting()` in `VLLMPrompter.prompt_batch`, which is fixed on the sibling branch of PR #163 (it defers the call to teardown via `close()`). Once #163 merges, the per-row fallback resubmissions work fine: no submit path checks the finished flag (verified in #163's review). If #170 merges first, the fallback's failing rows behave no worse than main's current whole-batch substitution, so there is no regression window. Rather than duplicating #163's fix here, ea9e0bcf adds a one-line comment above the fallback's `prompt_batch` call documenting the #163 dependency.

kaka11chen:
Repository-state update after #397 merged: #147 is now closed because the replacement Prompt/Embed API removed the legacy paths described there. #148 remains open: `RetryAfterError` serialization and legacy classification/logging gaps are not fully resolved on current `main`. Please rebase and rescope this PR around the still-relevant #148 work before merge.

caomaocao:
Rescope is underway per the request above: rebasing onto current `main` and narrowing the diff to the still-open #148 scope (`RetryAfterError` serialization plus the classification/logging gaps). Two sequencing notes now that #572 is in flight for #567:

- #572 keeps its terminal-state raises inside the providers and leaves the `_retry_call` / `_retry_call_async` classification tuple in `functions.py` untouched, so its overlap with this PR is minimal — the rescope doesn't need to wait for it to land.
- #572's new terminal-state tests parameterize `on_error` over the full mode set that exists on `main` today. If the rescope introduces an additional mode, I'll extend those tests here rather than asking #572 to pre-account for it.

caomaocao:
Rescoped and force-pushed per the request above. What changed:

- **Rebuilt from current `main`** — the old diff was discarded wholesale rather than rebased hunk-by-hunk. Everything tied to the closed #147 (execution-option routing, silent-option handling, the old three-mode `on_error` design) is gone, superseded by #397's API replacement.
- **Kept, reworked against current `main`, exactly the scope named:** `RetryAfterError` serialization (the old diff's `__reduce__` would be wrong on current `main` — the constructor now deliberately retains only sanitized state, so the reducer rebuilds from that state instead of re-passing the original error); substitution logging under `on_error="ignore"` as always-warn — no new mode, so #572's `on_error`-parameterized terminal-state tests need no extension; and uniform Retry-After extraction shared by Google/OpenAI/Anthropic, with Google's existing extraction tests passing unmodified.
- **Not re-fixed:** the chunked-embedding bypass and batch-granularity substitution from the original #148 report — both already fixed on current `main`.
- **#572:** no contact — the retry helpers and the non-retryable classification tuple are the seam, and neither PR crosses it.

vane/ai/functions.py
```diff
@@ -132,13 +133,62 @@ def _provider_is_loop_bound(provider: Any, method_name: str) -> bool:
...
 def _validate_on_error(on_error: object) -> _OnError:
     if not isinstance(on_error, str) or on_error not in ("raise", "ignore"):
         raise ValueError("on_error must be 'raise' or 'ignore'")
     return cast(_OnError, on_error)
 
 
+def _log_substituted_failure(exc: Exception, *, on_error: _OnError, attempts: int | None = None) -> None:
+    """Record one bounded WARNING when a provider failure is replaced with NULL.
+
+    Called at each site that swallows a provider error and yields NULL under
```
kaka11chen, line 150: **[P2] Cover NULL substitutions on the native vLLM path**

The public vLLM path never reaches this helper: `NativeVLLMPromptPlan` is lowered directly and maps `on_error="ignore"` to the native `"null"` policy. `LocalVLLMExecutor._generate` and `_append_error_rows` enqueue `None` without warning, and `_ValidateStructuredOutputBatch` catches `OutputValidationError` and appends `None` without calling this helper. An invalid structured vLLM result therefore still produces NULL with zero `vane.ai.functions` warning records, contrary to this docstring and the PR's every-provider guarantee. Please add the same bounded warning at the native substitution sites, with native-path regression coverage (or explicitly narrow the claimed scope).

vane/ai/functions.py
```diff
@@ -239,6 +298,72 @@ def _is_transient_provider_error(exc: Exception) -> bool:
...
+        return None
+    headers = getattr(response, "headers", None) or {}
+    try:
+        raw = headers.get("Retry-After") or headers.get("retry-after")
+    except Exception:
+        return None
+    if raw is None:
+        return None
+    try:
+        parsed = float(raw)
```
kaka11chen, line 378: **[P2] Parse the HTTP-date form of Retry-After**

[`Retry-After` permits either delay-seconds or an HTTP-date](https://www.rfc-editor.org/rfc/rfc9110#section-10.2.3). This `float()`-only path treats a valid value such as `Thu, 13 Aug 2026 15:37:13 GMT` as malformed, so `_retry_after_error` silently substitutes 5 seconds and may retry before the server allows it. Please parse HTTP-date relative to the current UTC time (clamping an already-past date to zero and retaining the existing 120-second sleep cap), and add a fixed-clock regression test.

vane/ai/functions.py
```diff
@@ -239,6 +298,72 @@ def _is_transient_provider_error(exc: Exception) -> bool:
...
+    uniform: the server-requested ``Retry-After`` is honoured when present and
+    usable, otherwise a default wait applies. The returned error carries the
+    original for its sanitized summary and status attributes only.
+    """
+    if _provider_status_code(exc) not in _RATE_LIMIT_STATUS_CODES:
+        return None
+    retry_after = _parse_retry_after_header(exc)
+    if retry_after is None:
+        retry_after = _DEFAULT_RETRY_AFTER_SECONDS
+    return RetryAfterError(retry_after=retry_after, original=exc)
```
kaka11chen: **[P3] Preserve a status discovered only on the response**

`_provider_status_code` deliberately qualifies errors whose 429/503 exists only at `exc.response.status_code` or `.status`, but `RetryAfterError.__init__` copies only direct attributes from `exc`. The returned wrapper therefore has only `retry_after`; after retry exhaustion its bounded warning (and its pickled state) says just `RetryAfterError` and loses the numeric status this helper already discovered. Please attach the normalized status to the wrapper (for example as `status_code`) and extend the response-attribute test through warning/pickle serialization.

kaka11chen (commented):
Reviewed `d3c3cc1580` against current `main` after #572 merged. The trees merge cleanly, but please refresh the branch before merge so Required CI exercises the combined terminal-validation and retry/error-policy paths. I left three inline findings: two behavior/contract gaps and one diagnostic-metadata gap.

vane/ai/functions.py
caomaocao: Fixed in 33f0201e. `_parse_retry_after_header` now falls back to an HTTP-date parse (`email.utils.parsedate_to_datetime`) when the value isn't delay-seconds, converting it to the remaining delay against current UTC. A `_utcnow()` seam lets the regression tests pin the clock; a naive `-0000` offset is read as UTC per the HTTP GMT convention. An already-past date clamps to zero (retry immediately) rather than being rejected as malformed, and the far-future case is still bounded by the existing 120-second cap in `_retry_wait_seconds` — I assert that end-to-end rather than duplicating the cap in the parser. Fixed-clock tests cover the honored, offset-form, past-clamped, and far-future-capped cases, and a bad-month date joins the malformed-header parametrization.

vane/ai/functions.py
caomaocao: Fixed in 05110a3c. `_retry_after_error` already has the normalized status in hand from `_provider_status_code` (which is what qualified the error in the first place), so it now attaches it to the wrapper as `status_code` right after construction — covering exactly the response-only case the constructor's direct-attribute whitelist can't reach. `RetryAfterError` gained a `status_code: int` annotation, and its existing `__reduce__` carries `__dict__`, so the attribute survives pickling for free. The response-attribute test now asserts the status shows up on both post-exhaustion surfaces: `status_code=429` in the sanitized `_safe_original_error_summary`, and a pickle round-trip that preserves it.

vane/ai/functions.py
caomaocao: Fixed in 26689154. The native `"null"` policy is the lowered form of the public `on_error="ignore"`, so I routed all three native substitution sites through the same bounded `_log_substituted_failure` the HTTP providers use: per failed row in `LocalVLLMExecutor._generate`, once per substituted batch in `_append_error_rows` (the shared sanitized engine-init failure — one warning, not one per row), and per rejected row in `_ValidateStructuredOutputBatch` (the structured-output validator #572's native path reaches). The warnings stay sanitized — class plus numeric-status summary, never prompt text or row payloads. Native-path regression tests cover each site, and assert that `raise` mode and a genuine provider NULL both stay silent so a real NULL isn't misreported as a substitution.

caomaocao: Follow-up in 7117aa38: the two executor sites no longer hardcode `on_error="ignore"` — a single `_log_null_substitution` helper now maps `self.on_error` back to the public mode in one place, so a raise-mode caller is structurally unable to log a substitution that is not happening (pinned by a red-run-verified raise-mode `_append_error_rows` test). A sibling cleanup, ac7b0327, folds the status attach from the thread above into `RetryAfterError`'s constructor (`status: int | None = None`) instead of post-construction assignment — behavior unchanged, same tests pin it.

kaka11chen:
@codex review

kaka11chen (approved):
LGTM

PR #171, opened 2026-07-22
[fix](ray) Prevent cross-fragment FTE lock inversion

### LINKED ISSUE #63: [P0][Bug] Cross-fragment state lock ordering can deadlock the Ray control plane
## Problem

A task-finished path can hold fragment F1's `_state_lock` and call pressure accounting, which invokes a global pending drain without a query filter and acquires other fragment locks. Admission paths can hold F2 while consulting or scheduling F1.

This creates an F1→F2 / F2→F1 ABBA cycle across fragments and can block the shared driver control plane.

kaka11chen:
## Summary

- make each per-query FTE scheduler the single writer for admission and status mutations
- route every capacity release through one global single-flight dirty-bit admission pump
- preserve admission priority in explicit phases: eager speculative, standard, then speculative only when no active query has standard work pending
- admit at most one attempt per query/class event quantum and coalesce duplicate internal admission pulses
- release terminal pressure outside fragment state locks and keep global fragment scans outside every fragment lock
- fence result-handle publication through status-watcher registration against query teardown
- isolate reservation and admission-scan failures to the affected query while continuing healthy queries
- preserve retryable registry ownership until teardown succeeds

## Root cause

Completion could hold fragment F1 state and synchronously drain all pending fragments, while admission held fragment F2 state and inspected F1 for global priority. The opposite F1 -> F2 and F2 -> F1 order formed an ABBA deadlock. Incremental wakeup fixes still left multiple callers able to enter cross-query admission, which created separate fairness, lost-wakeup, and teardown races.

The long-term fix gives cross-query admission one owner. Capacity changes only dirty the global pump; the pump delegates bounded work to each query scheduler, and each scheduler serializes that query state. Fragment callbacks never perform a global traversal while holding a fragment state lock.

## Regression coverage

Deterministic tests cover:

- the original two-query/two-fragment lock inversion
- scheduler idle-handoff and pump dirty-bit wakeups
- duplicate internal pulse coalescing
- failed reservation and failed global-scan isolation
- standard/speculative priority and cross-query fairness
- reservation completion versus concurrent drains and query drop
- create-before-follow-up command ordering
- result-handle publication versus watcher registration and retryable teardown ownership
- capacity release during terminal status and query teardown

## Testing

- `pre-commit run --files <changed files>`
- `python -m pytest tests/fast/test_ray_fte_event_scheduler.py tests/fast/test_ray_fragment_submission.py tests/fast/test_ray_fte.py` (319 passed)
- `scripts/run_release_tests.sh` (266 non-real-Ray passed; 5 real-Ray passed)

Closes #63

### DIFF
--- a/duckdb/runners/fte/fte_events.py
+++ b/duckdb/runners/fte/fte_events.py
@@ -78,10 +78,7 @@ class MemoryPressureDetected(FteEvent):
 
 @dataclass(frozen=True)
 class ResourceAdmissionChanged(FteEvent):
-    """Wake pending FTE reservations after resource or worker capacity changes."""
-
-    # Worker capacity is shared across queries; query-resource changes are not.
-    global_scope: bool = False
+    """Wake pending FTE reservations after query resource ownership changes."""
 
 
 @dataclass(frozen=True)
--- a/duckdb/runners/fte/fte_events.py
+++ b/duckdb/runners/fte/fte_events.py
@@ -78,7 +78,14 @@ class MemoryPressureDetected(FteEvent):
 
 @dataclass(frozen=True)
 class ResourceAdmissionChanged(FteEvent):
-    """Wake pending FTE reservations after query resource ownership changes."""
+    """Wake pending FTE reservations after resource ownership changes.
+
+    Internal pump pulses are deliberately excluded from user-facing event
+    progress counters.
+    """
+
+    execution_class: str | None = None
+    internal: bool = False
 
 
 @dataclass(frozen=True)
--- a/duckdb/runners/fte/fte_events.py
+++ b/duckdb/runners/fte/fte_events.py
@@ -78,7 +78,10 @@ class MemoryPressureDetected(FteEvent):
 
 @dataclass(frozen=True)
 class ResourceAdmissionChanged(FteEvent):
-    """Wake pending FTE reservations after query resource ownership changes."""
+    """Wake pending FTE reservations after resource or worker capacity changes."""
+
+    # Worker capacity is shared across queries; query-resource changes are not.
+    global_scope: bool = False
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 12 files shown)


kaka11chen:
Implemented the long-term admission design in `9d364663`, with the final control-failure handoff fix in `a4048582`.

The main change is that capacity release no longer calls an arbitrary global drain. A single-flight dirty-bit pump owns cross-query admission and delegates bounded pulses to each query scheduler, so query mutations remain serialized while global scans run outside fragment locks. Retry-delay, terminal pressure release, reservation completion, and query teardown now all use that same path.

I also added deterministic coverage for idle handoff, re-entrant wakes, duplicate-pulse coalescing, reservation/scan failure isolation, command ordering, and handle publication versus teardown. Final local gates: 319 affected tests, 266 release tests, 5 real-Ray release tests, plus pre-commit/mypy.

Final GitHub CI is green, including native build and fast tests on Python 3.12, 3.13, and 3.14, all source-package jobs, and Required CI.

kaka11chen:
Fixed the two remaining handoff races in 326645c4.

- Reservation pressure now moves to running pressure under one registry-lock critical section, so admission never observes a false capacity gap after create succeeds.
- A control RPC failure now quarantines the worker immediately, puts WorkerFailed ahead of already-queued reservation completions, and fans reconciliation out to every active query scheduler before draining. This prevents a second create on the failed worker and also retries work owned by other queries.
- Added deterministic regressions for both interleavings.

Validation: 321 focused FTE/Ray tests passed; full pre-commit (including ruff and mypy) passed; all 5 real-Ray release tests passed. The non-real-Ray release shard passed 265/266 locally; the sole failure was the SourceID check because the worktree intentionally reused the pre-change native install for this Python-only patch.

kaka11chen:
@codex  review

hubgeter (approved):
LGTM

kaka11chen:
@codex  review

PR #177, opened 2026-07-23
[fix](video) Reject invalid decode concurrency

### LINKED ISSUE #68: [P1][Bug] Video decoder concurrency accepts zero and blocks every decode

freemandealer:
## Summary

- validate `VANE_MAX_CONCURRENT_DECODES` before constructing the process-level semaphore
- reject zero, negative, empty, and non-integer values with the variable name and accepted range in the error
- cover default and positive values plus invalid import-time configurations in isolated subprocesses

Fixes #68.

## Validation

- `pre-commit run --files duckdb/datasource/video_reader.py tests/fast/test_video_reader.py` (all hooks passed, including mypy and Ruff format/check)
- `python -m pytest tests/fast/test_video_reader.py` (68 passed)
- `env -u http_proxy -u https_proxy scripts/run_release_tests.sh` (414 passed, 1 skipped in the non-Ray shard; 7 passed in the real-Ray shard)
- `git diff --check origin/main...HEAD`

## Notes

Rebased onto `origin/main` at `e583c7e5`. The updated main branch pins the mypy hook to `numpy<2.4` and uses dynamic imports for optional video dependencies, so the previous PIL/psutil/decord and NumPy-stub lint failures are resolved.

### DIFF
--- a/duckdb/datasource/video_reader.py
+++ b/duckdb/datasource/video_reader.py
@@ -145,6 +145,19 @@ def _read_optional_positive_int_env(name: str) -> int | None:
     return result
 
 
+def _read_positive_int_env(name: str, default: int) -> int:
+    value = os.environ.get(name)
+    if value is None:
+        return default
+    try:
+        result = int(value)
+    except ValueError:
+        raise ValueError(f"{name} must be an integer >= 1, got {value!r}") from None
+    if result < 1:
+        raise ValueError(f"{name} must be an integer >= 1, got {value!r}")
+    return result
+
+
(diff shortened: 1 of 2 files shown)


kaka11chen (approved):
LGTM

kaka11chen:
@codex review

PR #184, opened 2026-07-24
[fix](ray) Preserve persistent worker CUDA visibility

### LINKED ISSUE #74: [Bug] Preserve node CUDA visibility mapping in persistent Ray workers
## Context

Vane intentionally starts one persistent `RayWorkerActor` per Ray node without declaring the node's full `num_cpus` or `num_gpus` in the actor options. The actor hosts the node-local execution engine; CPU/GPU capacity is passed to its constructor as metadata for Vane's internal scheduling and query admission.

This follows the Daft/Flotilla worker model: one engine actor per node with full access to the node, while the framework scheduler accounts for work running inside it. Keeping the persistent actor from reserving the node's full CPU/GPU capacity is also necessary so that real `ray_task` and `ray_actor` UDF executions can request those resources through Ray.

Therefore, the missing `num_cpus`/`num_gpus` actor options are deliberate and are not the bug tracked here.

## Problem

The worker currently reconstructs CUDA visibility from the GPU count:

```python
if num_gpus > 0:
    os.environ["CUDA_VISIBLE_DEVICES"] = ",".join(str(i) for i in range(num_gpus))
```

This was inherited from the Daft workaround for Ray versions that clear `CUDA_VISIBLE_DEVICES` for an actor that requests no GPUs. It restores GPU visibility on the common `0..n-1` layout, but it loses the node's original device mapping.

For example, a node started with:

```text
CUDA_VISIBLE_DEVICES=2,5
```

must not become:

```text
CUDA_VISIBLE_DEVICES=0,1
```

The same concern applies to GPU UUIDs and MIG device identifiers. Because the persistent worker intentionally requests zero Ray GPU resources, there is no non-zero Ray-assigned actor mapping to preserve; the mapping that matters is the node's original visibility mapping.

Ray's behavior also differs across supported versions. Ray 2.53-2.55 clears accelerator visibility for zero-GPU actors by default, while Ray 2.56+ preserves the inherited environment by default. See ray-project/ray#54868.

Relevant files:

- `duckdb/runners/ray/worker_pool.py`
- `duckdb/runners/ray/worker.py`

## References

- Eventual-Inc/Daft#4731
- ray-project/ray#54868

kaka11chen:
## Summary

- keep persistent node workers free of Ray CPU/GPU reservations; their node resource counts remain Vane capacity and admission metadata
- preserve each node process CUDA visibility in zero-GPU actors across Ray 2.53+, instead of rebuilding it as contiguous local ordinals
- cover non-contiguous ordinals plus UUID and MIG mappings, and assert worker actor options do not reserve CPU or GPU resources

Shared-cluster isolation and resource leasing or placement groups remain a separate architecture concern.

## Tests

- `python -m pytest -m "not real_ray" tests/fast/test_ray_worker_cuda_visibility.py tests/fast/test_ray_fragment_submission.py` (145 passed)
- real-Ray non-contiguous mapping test on Ray 2.56.1 and Ray 2.55.1
- equivalent actor compatibility check on Ray 2.53.0
- `scripts/run_release_tests.sh` (272 passed, 1 skipped; 5 Ray tests passed)

Closes #74

### DIFF
--- a/duckdb/runners/ray/worker.py
+++ b/duckdb/runners/ray/worker.py
@@ -1,5 +1,6 @@
 # SPDX-FileCopyrightText: 2026 Vane contributors
 # SPDX-License-Identifier: Apache-2.0
+# mypy: disable-error-code="untyped-decorator"
 
 from __future__ import annotations
 
@@ -8,7 +9,8 @@ import os
 import sys
 import threading
 import time
-from typing import Any, NamedTuple
+from collections.abc import Mapping
+from typing import Any, NamedTuple, cast
 
 import ray
 
@@ -137,7 +139,7 @@ def _maybe_chaos_kill_worker(task_id: FteTaskAttemptId) -> None:
     os._exit(88)
 
 
-def _normalize_native_task_result(result: Any):
+def _normalize_native_task_result(result: Any) -> tuple[Any, ...]:
     native_type = require_ray_cxx_attr(
         "NativeDistributedTaskResult",
         hint="Ensure the C++ ray extension is built and importable in this process.",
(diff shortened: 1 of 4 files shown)


kaka11chen:
@codex review

hubgeter (approved):
LGTM

PR #186, opened 2026-07-24
[fix](udf) Preserve dispatcher context lifetime

### LINKED ISSUE #79: [P1][Bug] UDF unregister timeout can permanently poison the process and retain ClientContext
## Problem

A dispatcher unregister timeout records a process-global dispatcher error with no clear recovery path. Later UDF registrations can fail for the lifetime of the process.

The timeout cleanup path also differs from normal dispatcher cleanup and can retain a raw `ClientContext*` while the caller destroys the connection, creating a potential use-after-free when delayed conversion resumes.

kaka11chen:
## Summary

- replace the unregister boolean with explicit `ACTIVE -> RETIRING -> DETACHED -> CLEANED` slot states
- generation-fence queued commands, delayed results, statistics, and wakeup callbacks
- store `ClientContext` weakly and acquire a generation-validated `shared_ptr` lease for input and output Arrow conversion
- strip non-owning `ClientProperties::client_context` pointers before work enters dispatcher queues
- keep unregister and async-shutdown deadline failures slot-local instead of permanently poisoning the dispatcher
- add deterministic regressions for blocked result handling and blocked input conversion, connection destruction, eventual stale-slot cleanup, and a subsequent healthy UDF query

## Root cause

Queued dispatcher work retained a non-owning `ClientContext *`. An unregister timeout returned control to the caller without stopping a blocked Arrow conversion, so closing the connection could free the context before the dispatcher resumed. A generation check after conversion was too late because `ArrowConverter::ToArrowSchema` had already dereferenced the stale pointer.

## Validation

- `python -m pytest tests/fast/test_udf_process.py` — 6 passed
- both unregister-timeout regressions repeated 10 times each — 20 passed
- `scripts/run_release_tests.sh` — 414 passed, 1 skipped, 11 deselected; real-Ray shard 7 passed
- native ASan comparison: the pre-fix binary reproduces `heap-use-after-free`; the fixed binary passes both lifecycle regressions

Closes #79.

### DIFF
--- a/src/duckdb_py/duckdb_python.cpp
+++ b/src/duckdb_py/duckdb_python.cpp
@@ -1167,6 +1167,7 @@ PYBIND11_MODULE(DUCKDB_PYTHON_LIB_NAME, m) { // NOLINT
 	m.def("_udf_executor_debug_counters", &GetUDFExecutorDebugCounters);
 	m.def("_reset_udf_executor_debug_counters", &ResetUDFExecutorDebugCounters);
 	m.def("_shutdown_udf_executor_dispatcher", &ShutdownUDFExecutorDispatcher);
+	m.def("_wake_udf_executor_slots_for_testing", &WakeUDFExecutorSlotsForTesting);
 	try {
 		py::module_::import("atexit").attr("register")(m.attr("_shutdown_udf_executor_dispatcher"));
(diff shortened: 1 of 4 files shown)


kaka11chen:
@codex review

hubgeter (approved):

PR #194, opened 2026-07-25
[feat](packaging) Add video and image optional dependency extras

### LINKED ISSUE #72: [P2][Packaging] Add a complete optional dependency group for video support
## Problem

The public video data source uses Pillow and depends on psutil/decord at runtime, but the project does not provide one supported optional dependency group that installs the complete feature.

Users can enable the API without having a documented installation command that supplies all required packages. ndarray image paths in AI providers also depend on Pillow.

zy-kkk:
## Summary

Implements the complete optional dependency story for the video data source, plus a shared `image` extra for the ndarray image paths in AI providers.

- Add `video = ["pillow>=10", "psutil>=5.9", "decord>=0.6; platform_system == 'Linux' and platform_machine == 'x86_64'"]`. decord is gated with the same environment marker as the existing `vllm` extra: the marker matches Vane's currently supported native platform (Linux x86-64). The dependency-side constraint is that decord 0.6.0 publishes no wheels for modern Python on macOS and none for any ARM platform, and has no sdist, so an unmarked extra would make `pip install 'vane-ai[video]'` unresolvable there. decord's existing `win_amd64` wheel can be enabled explicitly if Vane adds Windows support.
- Add `image = ["pillow>=10"]` as a shared extra for the ndarray→image conversion sites in the OpenAI/Anthropic/Google providers. Provider extras stay SDK-only; the conversion sites raise `ProviderImportError("image")` with the exact install command. The main pipeline that feeds ndarrays into prompts (video frames) already has Pillow via the `video` extra.
- Make all three optional imports in `video_reader.py` (PIL / psutil / decord) lazy through one helper that names the exact extra on failure; the decord message states the platform bound explicitly. The base import stays light and the base install gains no new dependencies (asserted by a new metadata test). With the static imports gone, the `# type: ignore` comments they carried are no longer needed.
- `all` gains `pillow` only, consistent with its existing note that heavy/platform-gated runtimes stay opt-in.
- Document the extras in the README installation section.

## Related issue

Closes #72.

## Documentation impact

README installation section now documents the `image`/`video` extras and the

## Validation

- Rebased onto `59b9aaec`; the typing annotations from #189 are preserved (the lazy-import change subsumes their `# type: ignore` comments).
- New metadata tests extend the existing `_requirements_for_extra` pattern i_metadata.py`; the decord assertion is platform-aware so it passes on anyrunner. A base-install test asserts pillow/psutil/decord stay out of the default dependency set.
- New `tests/fast/test_video_optional_deps.py` covers the error messages forlus the provider `image` path, simulating missing dependenciesdeterministically via `sys.modules` blocking — the tests pass regardless of which optional packages the environment has installed (relevant now that CI's test dependencies include pillow).
- Local verification on this host: `ruff check` / `ruff format --check` clean, `py_compile` on all touched files, `tomllib` parse of `pyproject.toml`. Test execution is deferred to CI: this host is macOS arm64, which the native build does not currently target, so the fast-test suite cannot run locally. Python-only change, no native rebuild required.          
## Checklist

- [x] The change is focused and includes regression coverage.                                                                                                                       - [x] Public behavior and native compatibility impact are documented. (New ohavior change for existing installs.)
- [x] New dependencies, copied code, model assets, and datasets have compatible licenses and are recorded where required. (Declared optional extras only — pillow (MIT-CMU), psutil (BSD-3-Clause), decord (Apache-2.0); nothing is vendored, matching how existclared.)
- [x] Security implications of UDFs, serialization, remote code, network accave been considered. (Import-error paths only; no new code executesthird-party input.)
- [x] Native changes were compiled; affected native, Python, release, formatting, and lint checks passed. (No native changes; formatting and lint verified locally, tests via CI.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,16 @@ Install the `vane-ai` package from PyPI:
 pip install vane-ai
 ```
 
+Optional features are provided as extras:
+
+```bash
+pip install 'vane-ai[openai]'   # OpenAI provider (anthropic / google / transformers / vllm likewise)
+pip install 'vane-ai[image]'    # ndarray image inputs for AI providers (Pillow)
+pip install 'vane-ai[video]'    # video data source (Pillow, psutil, decord)
+```
+
+The `video` extra installs `decord` on Linux x86-64 only, matching the platforms covered by its prebuilt wheels.
+
 For more details, see the [Installation Guide](https://vane.astrovela.ai/docs/data/quickstart/installation).
 
 ### Quick Start
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ pip install 'vane-ai[image]'    # ndarray image inputs for AI providers (Pillow)
 pip install 'vane-ai[video]'    # video data source (Pillow, psutil, decord)
 ```
 
-The `video` extra installs `decord` on Linux x86-64 only, matching the platforms covered by its prebuilt wheels.
+The `video` extra installs `decord` on Linux x86-64, Vane's currently supported native platform. decord itself publishes no wheels for modern Python on macOS or for any ARM platform; if Vane adds Windows support later, decord's existing `win_amd64` wheel can be enabled explicitly.
 
 For more details, see the [Installation Guide](https://vane.astrovela.ai/docs/data/quickstart/installation).
 
--- a/duckdb/datasource/video_reader.py
+++ b/duckdb/datasource/video_reader.py
@@ -17,20 +17,33 @@ Usage::
 
 from __future__ import annotations
 
+import importlib
 import os
 import threading
 import time
 from collections.abc import Iterator
 from concurrent.futures import ThreadPoolExecutor
 from itertools import repeat
+from types import ModuleType
 from typing import Any, cast
 
 import numpy as np
 import numpy.typing as npt
 import pyarrow as pa
-from PIL import Image  # type: ignore[import-not-found]
 
 from duckdb.datasource import DataSource, DataSourceTask
+
+
+def _import_video_dependency(module_name: str, package_name: str) -> ModuleType:
+    """Import an optional video dependency, reporting the extra to install on failure."""
+    try:
+        return importlib.import_module(module_name)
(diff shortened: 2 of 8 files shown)


kaka11chen:
@codex review

kaka11chen:
One clarification on the platform rationale: `decord==0.6.0` does not ship wheels *only* for Linux x86-64. PyPI also provides a `py3-none-win_amd64` wheel. Its macOS wheels are limited to CPython 3.6–3.8 on Intel, and it provides neither ARM64 wheels nor an sdist: https://pypi.org/project/decord/0.6.0/#files

Keeping the dependency marker limited to Linux x86-64 can still be reasonable because that is Vane’s currently tested and published native platform. However, the `pyproject.toml` comment, README wording, commit/PR description, and runtime hint currently attribute the restriction to decord’s wheel coverage alone, which is not quite accurate. Could we instead describe this as Vane’s current supported video platform, with decord’s incomplete modern macOS/ARM coverage as the dependency constraint? If Vane later supports Windows, the existing Windows AMD64 decord wheel could then be included explicitly.

kaka11chen:
The Copilot suggestion is directionally correct: `Requirement.marker.evaluate()` fills its environment from Packaging’s defaults, where `platform_system` comes from `platform.system()`, so using `platform.system() == "Linux"` keeps the test aligned with the exact PEP 508 key used in `pyproject.toml`. In normal CPython Linux environments, this is effectively equivalent to `sys.platform == "linux"`, so I would treat it as a consistency improvement rather than a medium-severity defect.

It would be better to strengthen the assertion at the same time so the test also verifies that decord is absent off the supported platform:

```python
supports_decord = (
    platform.system() == "Linux"
    and platform.machine() == "x86_64"
)
assert ("decord" in selected) is supports_decord
```

The current conditional checks only the positive Linux x86-64 case and silently makes no assertion about decord on every other platform.

kaka11chen (approved):
LGTM

PR #195, opened 2026-07-25
[fix](ray) Keep result collectors query-local

### LINKED ISSUE #82: [P1][Bug] Distributed plan binding can permanently replace a connection result collector
## Problem

Distributed execution modifies `client_config.get_result_collector` on the connection. Restoration is not guaranteed across every success and exception path, so a long-lived connection can remain configured for the distributed/materialized collector after the operation ends.

kaka11chen:
## Summary

- add a query-local materialized result collector override to `PendingQueryParameters`, ahead of the legacy connection-level callback
- use the query-local override for native distributed execution so the connection callback is never mutated
- cover success, planning and collector initialization failures, execution failure, cancellation, teardown failure, and same-connection local/distributed/local behavior
- verify that an existing custom connection callback remains installed and observes only local materialized queries

## Tests

- `build/native-cxx20/test/unittest [result-collector]` — 54 assertions passed
- `.venv/bin/python -m pytest -q tests/fast/test_ray_cpp_bindings.py::test_execute_native_keeps_result_collector_query_local` — passed
- `scripts/run_release_tests.sh` — 304 non-Ray tests and 7 real-Ray tests passed
- CPU-only `scripts/run_fast_tests.sh` profile — 5,238 non-Ray tests and 77 shared-Ray tests passed; one test-owned Ray process reported PASS but exited before its pytest summary on the aggregate run, and the complete owner-Ray phase then passed on an isolated rerun (6 passed, 2 multi-node skips)
- pre-commit hooks for all changed files — passed

Fixes #82

### DIFF
--- a/external/duckdb/src/include/duckdb/main/client_context.hpp
+++ b/external/duckdb/src/include/duckdb/main/client_context.hpp
@@ -65,6 +65,9 @@ struct PendingQueryParameters {
 	optional_ptr<case_insensitive_map_t<BoundParameterData>> parameters;
 	//! Whether a stream/buffer-managed result should be allowed
 	QueryParameters query_parameters;
+	//! Query-local callback used to create a materialized result collector.
+	//! Takes precedence over the connection-level callback without modifying connection state.
+	get_result_collector_t get_result_collector = nullptr;
 };
 
 //! The ClientContext holds information relevant to the current client session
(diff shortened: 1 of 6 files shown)


kaka11chen:
@codex review

kaka11chen:
@codex review

QuakeWang (approved):
LGTM

PR #198, opened 2026-07-25
[fix](udf) Restore socket timeout after subprocess handshake

### LINKED ISSUE #65: [P1][Bug] UDF startup handshake does not restore blocking socket timeout

jingdaws:
PR description

  ## Summary

  Fix the UDF subprocess control socket leaking its temporary handshake timeout
  into normal task communication.

  `_recv_expected` previously used `None` both as the default sentinel and as a
  valid timeout value for a blocking socket. As a result, sockets that started in
  blocking mode were not restored after the handshake.

  This change:

  - tracks timeout overrides independently from the previous timeout value;
  - restores blocking, finite-timeout, and non-blocking socket modes;
  - restores the timeout before broken-executor cleanup closes the socket;
  - adds regression coverage using real socket pairs for successful, worker-error,
    and transport-error paths.

  There are no public API, protocol, dependency, or configuration changes.

  ## Related issue

  close: #65

  ## Documentation impact

        `AstroVela/vane-website`.

  ## Validation

  - `ruff check duckdb/execution/udf_subprocess.py tests/fast/test_udf_executor_lifecycle.py`
  - `ruff format --check duckdb/execution/udf_subprocess.py tests/fast/test_udf_executor_lifecycle.py`
  - `python -m py_compile duckdb/execution/udf_subprocess.py tests/fast/test_udf_executor_lifecycle.py`
  - Targeted exact-method harness: 8 timeout and failure-path scenarios passed.
  - Full pytest suite was not run locally because a built `_duckdb` native
    extension was not available in the local environment. CI is expected to run
    the repository test suite.

  ## Checklist

  - [ ] Native changes were compiled; Python-only, documentation, and workflow

  The final item is left unchecked because the full repository pytest suite could
  not be run in the local environment.

### DIFF
--- a/duckdb/execution/udf_subprocess.py
+++ b/duckdb/execution/udf_subprocess.py
@@ -389,21 +389,24 @@ class _SingleSubprocessExecutor(BaseUDFExecutor):
 
     def _recv_expected(self, expected: tuple[int, ...], *, timeout_s: float | None = None) -> tuple[int, bytes]:
         sock = self._require_socket()
+        timeout_overridden = False
         restore_timeout = None
         if timeout_s is not None and hasattr(sock, "settimeout") and hasattr(sock, "gettimeout"):
             restore_timeout = sock.gettimeout()
             sock.settimeout(max(0.0, float(timeout_s)))
+            timeout_overridden = True
         try:
-            msg_type, payload = _recv_message(sock)
+            try:
+                msg_type, payload = _recv_message(sock)
+            finally:
+                if timeout_overridden:
+                    try:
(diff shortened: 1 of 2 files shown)


kaka11chen:
@codex review

kaka11chen (approved):
LGTM

PR #227, opened 2026-07-27
[fix](ai) Enforce a strict Anthropic option contract with explicit max_tokens (#146)

### LINKED ISSUE #146: [P2][AI] Anthropic provider: silent option whitelist, silent 1024-token truncation
## Problem

Two issues in `vane/ai/providers/anthropic.py`:

1. **Silent option dropping** (`anthropic.py:156-173,271-288`): only `max_tokens`, `temperature`, `top_p`, `top_k`, `stop_sequences` are forwarded; everything else (`thinking`, `metadata`, `tool_choice`, `service_tier`, `betas`, …) is silently discarded — e.g. enabling extended thinking via options is impossible and produces no error. Contrast: `OpenAIPrompter` forwards all unfiltered options.
2. **Silent truncation**: `max_tokens` defaults to 1024 and `stop_reason` is never checked, so long responses truncate silently; with `return_format`, a truncated turn can yield `None` with no diagnostic.

> The third sub-item originally filed here (stale default model) was split out and promoted to **#167 [P1]** after its announced retirement date passed. This issue keeps the two P2 items above.

## Suggested fix

Forward the remaining options dict (minus reserved execution keys) like the OpenAI prompter does; surface `stop_reason == "max_tokens"` (warning or error). Benchmark note: Vercel AI SDK requires a `warnings` array on every result — unsupported/dropped settings must be reported, never swallowed.

Related: #140 (thinking-block extraction bug, fixed in PR #160), #167 (stale default model, P1).

Reviewed commit: c43bd706.

caomaocao:
> [!NOTE]
> Was stacked on #168; now rebased onto `main` (single commit `d744dfd7`, on top of the merged #168 and #160) and marked ready for review.

## Summary

Per maintainer direction on #146, the Anthropic prompter's silent 5-key forwarding whitelist and the invisible `max_tokens=1024` default are replaced by a strict, validated contract enforced at descriptor construction — before anything ships to workers:

- **Forward the full known surface.** Every Messages API request option accepted by the installed `anthropic` SDK (0.117.0 `Messages.create`) forwards verbatim: `max_tokens`, `temperature`, `top_p`, `top_k`, `stop_sequences`, `thinking`, `metadata`, `tool_choice`, `service_tier`, `output_config`, `cache_control`, `container`, `inference_geo`, `user_profile_id`, plus the SDK per-request extras `extra_headers`/`extra_query`/`extra_body`. Keys Vane owns (`model`, `messages`, `system`, `tools`, `stream`) and client configuration (`api_key`, `base_url`, `timeout`, `max_retries`) stay out of the set by design. Beta features remain reachable via `extra_headers={"anthropic-beta": ...}` (the non-beta `messages.create` has no `betas` parameter).
- **Reject unknown options.** A key that is neither a known request option nor a Vane execution option (`on_error`, `max_api_concurrency`, ...) raises `ValueError` at expression-build time, listing the offending keys, the model, and the supported set. Together with #168's fully-named constructor (ctor typos are a `TypeError`), a misspelled option can no longer ride silently into (or vanish before) the request on any path.
- **Reject documented-unsupported combinations.** A per-model capability table `_MODEL_UNSUPPORTED_OPTIONS: dict[str, frozenset[str]]` plus `_validate_prompt_options(model_name, options)` (invoked from `AnthropicPrompterDescriptor.__post_init__`) rejects non-default sampling parameters (`temperature`/`top_p`/`top_k`) on the model families Anthropic documents as returning a 400 for them: `claude-sonnet-5` ([What's new in Claude Sonnet 5](https://platform.claude.com/docs/en/about-claude/models/whats-new-sonnet-5)) and `claude-opus-4-7`/`claude-opus-4-8`/`claude-opus-5` ([migration guide](https://platform.claude.com/docs/en/about-claude/models/migration-guide): "Setting `temperature`, `top_p`, or `top_k` to a non-default value returns a 400 error on Claude Opus 4.7 or later models"). Matching is by family prefix (exact id or `<family>-<suffix>`, so dated snapshots are covered). The table shape is intentionally identical to the Google one in #161 so both can be lifted into a core capability framework under #152.
- **Explicit `max_tokens`.** The implicit `1024` is deleted. `max_tokens` resolves as **call-site option > `AnthropicProvider(max_tokens=...)` (new keyword-only constructor parameter) > `ValueError`** naming both fix paths: `"No max_tokens configured for the Anthropic provider. Pass max_tokens=... or configure AnthropicProvider(max_tokens=...)."` Enforced in the descriptor `__post_init__`, so direct descriptor construction is covered too.
- **Truncation raises.** A response with `stop_reason == "max_tokens"` raises `ValueError` (naming the model and the configured `max_tokens`) instead of returning partial text — or, on the `return_format` path, a silent `None`. The raise flows through the existing UDF `on_error` policy (`raise`/`log`/`ignore`) like any other request failure.

No backward compatibility or deprecation window, per maintainer direction (early-stage project).

## Related issue

close: #146

## Documentation impact

      `AstroVela/vane-website`.

The option contract, `max_tokens` chain, and truncation behavior are documented in the provider module/class docstrings updated in this PR.

## Validation

- Stub-harness runs (local venv, `anthropic` 0.117.0 installed; duckdb-engine-dependent tests and Pillow-dependent tests are environment-blocked and fail identically on the base branch):
  - `pytest -p vane_pkg_stub_plugin --noconftest tests/ai/test_ai_functions.py -k Anthropic`: 49 passed, 1 environment-blocked (Pillow).
  - Full `tests/ai/test_ai_functions.py` + `tests/ai/test_descriptor_secret_redaction.py`: 294 passed, 16 environment-blocked failures — the identical 16 fail on the base commit; no new failures.
  - `tests/ai/test_expression_ai_functions.py` + `tests/fast/test_ai_options_repr.py`: 47 passed / 15 failed, failure set identical to the base commit (duckdb-engine stubs).
- New tests pin: unknown option (`thinkng=...`) rejected pre-dispatch on both the provider and direct-descriptor paths (a ctor typo like `promt_model=` is a `TypeError` from #168's named constructor); `thinking`/`metadata`/`tool_choice`/`service_tier` reaching the recorded `messages.create` kwargs while Vane execution options stay out; `claude-sonnet-5` (and a dated snapshot id) + `temperature`/`top_p` rejected while `claude-sonnet-4-6` passes; missing `max_tokens` error with both fix paths on both construction paths; per-call and provider-config `max_tokens` flow-through with call beating config; `stop_reason="max_tokens"` raising on both the plain-text and structured paths (mocks carry `.type`/`.text`/`.stop_reason`).
- Option-surface verification: `_KNOWN_PROMPT_OPTIONS` was generated from the installed SDK's `Messages.create` signature (`anthropic` 0.117.0), not from memory; the capability-table entries carry their documentation URLs in a source comment.
- `ruff check` / `ruff format --check` clean; pre-commit hooks passed on commit.

## Checklist

      licenses and are recorded where required. (None added.)
      and untrusted input have been considered. (No new surface; the change adds
      pre-dispatch validation and removes an implicit request parameter. Secret
      redaction behavior is unchanged and still covered by
      `test_descriptor_secret_redaction.py`.)
      changes passed relevant checks. (Python-only; ruff + pre-commit clean,
      stub-harness tests pass locally, full suite delegated to CI.)

https://claude.ai/code/session_01CXGCZZzpBqP7D1kivLJABc

### DIFF
--- a/tests/ai/test_ai_functions.py
+++ b/tests/ai/test_ai_functions.py
@@ -1201,6 +1201,58 @@ class TestAnthropicStrictOptions:
         with pytest.raises(ValueError, match="top_p"):
             self._provider(max_tokens=64).get_prompter(model="claude-sonnet-5", top_p=1.0)
 
+    @pytest.mark.parametrize("model", ["claude-fable-5", "claude-mythos-5", "claude-mythos-preview"])
+    def test_fable_mythos_sampling_option_rejected(self, model):
+        """The Fable/Mythos families document the same non-default rejection."""
+        with pytest.raises(ValueError, match="temperature") as excinfo:
+            self._provider(max_tokens=64).get_prompter(model=model, temperature=0.2)
+        assert model in str(excinfo.value)
+
+    def test_fable_default_temperature_passes(self):
+        """The documented default temperature is accepted on Fable too."""
+        desc = self._provider(max_tokens=64).get_prompter(model="claude-fable-5", temperature=1.0)
+        assert desc.prompt_options["temperature"] == 1.0
+
+    def test_fable_top_p_rejected(self):
+        with pytest.raises(ValueError, match="top_p"):
+            self._provider(max_tokens=64).get_prompter(model="claude-fable-5", top_p=1.0)
+
+    # --- per-model thinking.type restrictions -----------------------------
+
+    @pytest.mark.parametrize(
+        "model",
+        ["claude-sonnet-5", "claude-opus-4-7", "claude-fable-5", "claude-sonnet-5-20260601"],
+    )
+    def test_manual_thinking_rejected_on_adaptive_only_models(self, model):
+        """The removed manual form fails pre-dispatch on adaptive-only families."""
+        with pytest.raises(ValueError, match="thinking.type") as excinfo:
+            self._provider(max_tokens=64).get_prompter(model=model, thinking={"type": "enabled", "budget_tokens": 1024})
+        assert model in str(excinfo.value)
+
+    @pytest.mark.parametrize("model", ["claude-fable-5", "claude-mythos-5", "claude-mythos-preview"])
+    def test_disabled_thinking_rejected_on_always_on_models(self, model):
+        """Thinking cannot be turned off where it is always on."""
+        with pytest.raises(ValueError, match="disabled"):
+            self._provider(max_tokens=64).get_prompter(model=model, thinking={"type": "disabled"})
+
+    @pytest.mark.parametrize("model", ["claude-mythos-preview", "claude-opus-4-6"])
+    def test_manual_thinking_passes_where_still_supported(self, model):
+        """Mythos Preview and the 4.6 models still accept the manual form."""
+        desc = self._provider(max_tokens=64).get_prompter(
+            model=model, thinking={"type": "enabled", "budget_tokens": 1024}
+        )
+        assert desc.prompt_options["thinking"]["type"] == "enabled"
+
+    @pytest.mark.parametrize("model", ["claude-haiku-4-5", "claude-opus-4-5", "claude-sonnet-4-5"])
+    def test_adaptive_thinking_rejected_on_extended_only_models(self, model):
+        """Extended-thinking-only models reject the adaptive form."""
+        with pytest.raises(ValueError, match="adaptive"):
+            self._provider(max_tokens=64).get_prompter(model=model, thinking={"type": "adaptive"})
+
+    def test_adaptive_thinking_passes_on_adaptive_models(self):
+        desc = self._provider(max_tokens=64).get_prompter(model="claude-sonnet-5", thinking={"type": "adaptive"})
+        assert desc.prompt_options["thinking"] == {"type": "adaptive"}
+
     # --- extra_body cannot bypass the option contract ---------------------
 
     @pytest.mark.parametrize(
@@ -1396,6 +1448,42 @@ class TestAnthropicStrictOptions:
             prompt_options={"max_tokens": 64, "thinking": {"type": "adaptive"}},
         )
         assert desc.return_format is dict
+
+    # --- forced tool choice needs tool definitions ------------------------
+
+    @pytest.mark.parametrize(
+        "tool_choice",
+        [{"type": "any"}, {"type": "tool", "name": "extract_data"}],
+    )
+    def test_forced_tool_choice_without_return_format_rejected(self, tool_choice):
+        """Vane sends no tools without return_format, so forced tool use cannot be satisfied."""
+        with pytest.raises(ValueError, match="tool_choice"):
+            self._provider(max_tokens=64).get_prompter(tool_choice=tool_choice)
+
+    @pytest.mark.parametrize("tool_choice", [{"type": "auto"}, {"type": "none"}])
+    def test_no_tools_safe_tool_choice_passes(self, tool_choice):
+        """auto/none remain valid without tool definitions."""
+        desc = self._provider(max_tokens=64).get_prompter(tool_choice=tool_choice)
+        assert desc.prompt_options["tool_choice"] == tool_choice
+
+    # --- execution options flow into UDFOptions ---------------------------
+
+    def test_concurrency_and_batch_size_flow_into_udf_options(self):
+        """The raw-dict path carries concurrency/batch_size instead of dropping them."""
+        desc = self._provider(max_tokens=64).get_prompter(concurrency=3, batch_size=7)
+        udf_options = desc.get_udf_options()
+        assert udf_options.actor_number == 3
+        assert udf_options.batch_size == 7
+
+    def test_actor_number_and_matching_concurrency_pass(self):
+        """Equal values for the alias pair are not a conflict."""
+        desc = self._provider(max_tokens=64).get_prompter(actor_number=3, concurrency=3)
+        assert desc.get_udf_options().actor_number == 3
+
+    def test_conflicting_actor_number_and_concurrency_rejected(self):
+        """concurrency aliases actor_number; disagreeing values raise."""
+        with pytest.raises(ValueError, match="alias"):
+            self._provider(max_tokens=64).get_prompter(actor_number=2, concurrency=3)
 
     @pytest.mark.skipif(not _has_module("anthropic"), reason="anthropic not installed")
     def test_structured_output_dispatch_kwargs(self):
--- a/tests/ai/test_ai_functions.py
+++ b/tests/ai/test_ai_functions.py
@@ -1185,6 +1185,64 @@ class TestAnthropicStrictOptions:
         desc = self._provider(max_tokens=64).get_prompter(model="claude-sonnet-4-6", temperature=0.2)
         assert desc.prompt_options["temperature"] == 0.2
 
+    @pytest.mark.parametrize("default", [1, 1.0])
+    def test_sonnet5_default_temperature_passes(self, default):
+        """The documented default temperature (1.0) is accepted by the API."""
+        desc = self._provider(max_tokens=64).get_prompter(model="claude-sonnet-5", temperature=default)
+        assert desc.prompt_options["temperature"] == default
+
+    def test_sonnet5_boolean_temperature_rejected(self):
+        """True == 1 in Python, but a bool is not the documented default."""
+        with pytest.raises(ValueError, match="temperature"):
+            self._provider(max_tokens=64).get_prompter(model="claude-sonnet-5", temperature=True)
+
+    def test_sonnet5_default_like_top_p_still_rejected(self):
+        """top_p has no documented default, so any explicit value is non-default."""
+        with pytest.raises(ValueError, match="top_p"):
+            self._provider(max_tokens=64).get_prompter(model="claude-sonnet-5", top_p=1.0)
+
+    # --- extra_body cannot bypass the option contract ---------------------
+
+    @pytest.mark.parametrize(
+        "smuggled",
+        [
+            {"thinking": {"type": "enabled", "budget_tokens": 1024}},
+            {"temperature": 0.2},
+            {"max_tokens": 5},
+            {"model": "claude-other-model"},
+            {"stream": True},
+        ],
+    )
+    def test_extra_body_with_validated_field_rejected(self, smuggled):
+        """Vane-owned and known request fields cannot ride in extra_body."""
+        from vane.ai.providers.anthropic import AnthropicPrompterDescriptor
+
+        with pytest.raises(ValueError, match="extra_body") as excinfo:
+            AnthropicPrompterDescriptor(
+                model_name="claude-test-model",
+                prompt_options={"max_tokens": 64, "extra_body": smuggled},
+            )
+        assert next(iter(smuggled)) in str(excinfo.value)
+
+    def test_extra_body_non_mapping_rejected(self):
+        from vane.ai.providers.anthropic import AnthropicPrompterDescriptor
+
+        with pytest.raises(ValueError, match="extra_body"):
+            AnthropicPrompterDescriptor(
+                model_name="claude-test-model",
+                prompt_options={"max_tokens": 64, "extra_body": ["thinking"]},
+            )
+
+    def test_extra_body_with_unmodeled_field_passes(self):
+        """extra_body keeps its purpose: fields the SDK does not model."""
+        from vane.ai.providers.anthropic import AnthropicPrompterDescriptor
+
+        desc = AnthropicPrompterDescriptor(
+            model_name="claude-test-model",
+            prompt_options={"max_tokens": 64, "extra_body": {"future_api_field": {"enabled": True}}},
+        )
+        assert desc.prompt_options["extra_body"] == {"future_api_field": {"enabled": True}}
+
     # --- explicit max_tokens chain ----------------------------------------
 
     def test_get_prompter_without_max_tokens_raises(self):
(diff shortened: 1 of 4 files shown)


caomaocao:
Rebased onto `main` (single commit `d744dfd7`) after #168 and #160 merged, and marked ready for review — this is the last open PR of the no-default-model series. The tree combines cleanly with #160's text-block concatenation (no overlap in the response path after the rebase); local stub-harness suite is at the known-environment-failure baseline (359 passed across the touched files). CI running on the fresh head.

vane/ai/providers/anthropic.py
```diff
@@ -365,6 +471,16 @@ def _flush_content() -> None:
                 output_tokens=getattr(usage, "output_tokens", None),
             )
 
+        # Truncation is an error, never a silent partial answer. The raise
+        # routes through the UDF on_error policy like any other failure.
+        if getattr(response, "stop_reason", None) == "max_tokens":
```
kaka11chen, line 666: **P2: Preserve valid zero-token cache prewarming**

Anthropic documents `max_tokens=0` as a supported cache-prewarming request whose successful response intentionally has empty content and `stop_reason="max_tokens"`. This unconditional branch converts that successful response into a truncation error, so the newly exposed `cache_control`/zero-token path cannot be used through Vane. Please treat an effective `max_tokens == 0` response as successful prewarming, while rejecting the documented incompatible zero-token combinations before dispatch, and add a regression test.

Reference: https://platform.claude.com/docs/en/build-with-claude/prompt-caching#pre-warming-the-cache

vane/ai/providers/anthropic.py
```diff
@@ -331,20 +439,18 @@ def _flush_content() -> None:
...
             kwargs["system"] = self._system_message
 
-        # Forward extra options
-        for k in ("temperature", "top_p", "top_k", "stop_sequences"):
-            if k in self._options:
-                kwargs[k] = self._options[k]
-
-        # Structured output: use tool_use with forced tool choice
+        # Structured output: use tool_use with forced tool choice. This
+        # overrides any caller-supplied tool_choice for the extraction turn.
```
kaka11chen: **P2: Reject conflicting structured-output options instead of overriding them**

The strict contract says known options are forwarded or rejected, but `return_format=...` plus caller-supplied `tool_choice={"type":"none"}` passes descriptor validation and is then silently replaced with a forced `extract_data` choice here. That recreates the option-loss behavior this PR is meant to remove. It also makes `return_format` plus manual `thinking={"type":"enabled", ...}` dispatch a forced-tool/manual-thinking combination that Anthropic rejects, including on older models where manual thinking itself is valid. Please validate these combinations at descriptor construction and reject conflicting `tool_choice`/manual-thinking settings rather than overwriting them; add coverage for the dispatched kwargs.

Reference: https://platform.claude.com/docs/en/about-claude/models/extended-thinking-models#thinking-with-tool-use

vane/ai/providers/anthropic.py
caomaocao: Fixed in 1cda38e4: an effective `max_tokens == 0` response with `stop_reason == "max_tokens"` now returns `None` (successful prewarming) instead of raising, and the documented incompatible zero-token combinations — manual thinking (`thinking.type == "enabled"`), forced `tool_choice` (`tool`/`any`), `output_config.format`, and `return_format` — are rejected at descriptor construction with a ValueError naming the offending options (`stream` is not part of Vane's option surface). Regression tests cover the prewarm response (`content=[]` → `None`), each rejected combination, a plain `max_tokens=0` + `cache_control` descriptor constructing fine, and that positive-max_tokens truncation still raises.

vane/ai/providers/anthropic.py
caomaocao: Fixed in 1cda38e4: `return_format` plus a caller-supplied `tool_choice` now raises at descriptor construction instead of being silently replaced, and `return_format` plus manual extended thinking (`thinking={"type": "enabled", ...}`) is rejected the same way since the API refuses forced tool use with manual thinking; adaptive thinking remains allowed alongside `return_format` and is regression-tested. Dispatched kwargs are pinned by a recording test: with `return_format` the `messages.create` call carries `tool_choice={"type": "tool", "name": "extract_data"}` and the `extract_data` tool, while a caller `tool_choice` without `return_format` continues to flow through untouched (existing forwarding test).

kaka11chen:
@codex review

vane/ai/providers/anthropic.py
```diff
@@ -157,6 +333,13 @@ class AnthropicPrompterDescriptor(PrompterDescriptor):
     def __post_init__(self) -> None:
         self.provider_options = wrap_sensitive_options(self.provider_options)
         self.prompt_options = wrap_sensitive_options(self.prompt_options)
+        _validate_prompt_options(self.model_name, self.prompt_options)
+        if "max_tokens" not in self.prompt_options:
+            raise ValueError(
+                "No max_tokens configured for the Anthropic provider. "
+                "Pass max_tokens=... or configure AnthropicProvider(max_tokens=...)."
+            )
```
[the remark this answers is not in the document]
caomaocao: Fixed in e2b698d0: `max_tokens=None` is now treated as unconfigured on both paths — `get_prompter` falls back to the provider-level default when the call passes `None`, and descriptor construction raises the existing pre-dispatch error naming both fix paths when no default exists. The resolved value is additionally validated as a non-negative `int` with `bool` excluded, so `False` can no longer satisfy the `== 0` prewarming check either. Regression tests cover None-falls-back-to-config, None-without-config raising on both construction paths, and rejection of `"64"`, `64.0`, `False`, and `-1`.

tests/ai/test_ai_functions.py
```diff
@@ -1083,6 +1088,346 @@ def test_ctor_typo_is_a_type_error(self):
...
+        from vane.ai.providers.anthropic import AnthropicPrompterDescriptor
+
+        desc = AnthropicPrompterDescriptor(
+            model_name="claude-test-model",
+            return_format=dict,
+            prompt_options={"max_tokens": 64, "thinking": {"type": "adaptive"}},
+        )
+        assert desc.return_format is dict
+
+    def test_structured_output_dispatch_kwargs(self):
```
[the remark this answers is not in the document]
caomaocao: Fixed in e2b698d0: the three client-instantiating tests added in 1cda38e4 (`test_structured_output_dispatch_kwargs`, `test_zero_max_tokens_prewarm_returns_none`, `test_positive_max_tokens_truncation_still_raises`) now carry the same `_has_module("anthropic")` skip marker as the other client tests in the file, so the AI suite stays runnable without the optional `anthropic` extra.

vane/ai/providers/anthropic.py
```diff
@@ -32,6 +57,142 @@
     from vane.ai.typing import Options
 
 
+# Messages API request options Vane forwards verbatim. Verified against the
+# ``anthropic`` SDK (v0.117.0) ``Messages.create`` signature. Keys Vane owns
+# are deliberately absent: ``model``, ``messages``, ``system``, ``tools``, and
+# ``stream`` are managed by the prompter, and client configuration
+# (``api_key``, ``base_url``, ``timeout``, ``max_retries``) travels through
+# provider options.
```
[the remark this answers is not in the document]
caomaocao: Fixed in e2b698d0: the `anthropic` optional dependency is pinned to `>=0.117.0` in both the `anthropic` and `all` extras — the exact SDK surface `_KNOWN_PROMPT_OPTIONS` was generated from — so an older installed SDK can no longer accept options its `Messages.create` does not take and fail only on the worker.

vane/ai/providers/anthropic.py
```diff
@@ -32,6 +57,142 @@
...
+_KNOWN_PROMPT_OPTIONS: frozenset[str] = frozenset(
+    {
+        "cache_control",
+        "container",
+        "extra_body",
```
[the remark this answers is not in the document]
caomaocao: Fixed in e2b698d0: `extra_body` keeps its legitimate purpose — request fields the SDK does not model — but is now validated at descriptor construction: it must be a mapping, and any key that is Vane-owned (`model`, `messages`, `system`, `tools`, `stream`) or a known request option (the `_KNOWN_PROMPT_OPTIONS` surface, including `thinking`, the sampling parameters, and `max_tokens`) raises a ValueError naming the offending keys. That closes the bypass around the structured-output/manual-thinking rejection and keeps the truncation check reading the real `max_tokens`. Tests cover a smuggled `thinking`/`temperature`/`max_tokens`/`model`/`stream`, a non-mapping `extra_body`, and an unmodeled field still constructing fine.

vane/ai/providers/anthropic.py
```diff
@@ -32,6 +57,142 @@
...
+    unknown = sorted(key for key in options if key not in _KNOWN_PROMPT_OPTIONS and key not in _VANE_EXECUTION_OPTIONS)
+    if unknown:
+        raise ValueError(
+            f"Unknown Anthropic prompt option(s) {unknown} for model {model_name!r}. "
+            f"Supported Messages API request options: {sorted(_KNOWN_PROMPT_OPTIONS)}."
+        )
+    for family, unsupported in _MODEL_UNSUPPORTED_OPTIONS.items():
+        if model_name == family or model_name.startswith(f"{family}-"):
+            rejected = sorted(key for key in options if key in unsupported)
+            if rejected:
```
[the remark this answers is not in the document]
caomaocao: Fixed in e2b698d0, scoped to what the API documents. `temperature` at its documented default of `1` (int or float; `bool` excluded) now passes validation on the restricted families and forwards verbatim — per the Messages API reference (default 1.0) and the migration guide ("Setting temperature, top_p, or top_k to a non-default value returns a 400 error"). `top_p`/`top_k` intentionally stay rejected on presence: the API reference documents no default for either ("recommended for advanced use cases only"), so any explicit value is non-default by definition — there is no documented value to compare against. The validator error message now spells out that distinction. Tests: `temperature=1`/`1.0` pass on `claude-sonnet-5`, `temperature=True` and non-default values stay rejected, and `top_p=1.0` is still rejected.

vane/ai/providers/anthropic.py
```diff
@@ -32,6 +63,182 @@
...
+# https://platform.claude.com/docs/en/api/messages), so passing exactly that
+# value is accepted; ``top_p`` and ``top_k`` have no documented default, so
+# any explicit value is non-default and rejected. Documented sources:
+# https://platform.claude.com/docs/en/about-claude/models/whats-new-sonnet-5
+# (Claude Sonnet 5 returns a 400 error for non-default ``temperature``,
+# ``top_p``, or ``top_k``) and
+# https://platform.claude.com/docs/en/about-claude/models/migration-guide
+# ("Setting temperature, top_p, or top_k to a non-default value returns a
+# 400 error" on Claude Opus 4.7 and later Opus models).
+_MODEL_UNSUPPORTED_OPTIONS: dict[str, frozenset[str]] = {
```
kaka11chen, line 133: **P2: Cover the current Fable/Mythos sampling restrictions**

This table omits `claude-fable-5`, `claude-mythos-5`, and `claude-mythos-preview`, although all three are in the SDK 0.117.0 `ModelParam` surface targeted by this PR, and Anthropic documents that they reject non-default `temperature`, `top_p`, and `top_k`. Consequently, `get_prompter(model="claude-fable-5", max_tokens=64, temperature=0.2)` passes descriptor construction and fails with a 400 only after worker dispatch, contrary to the strict pre-dispatch contract. Please cover these families and add regression cases.

References: https://github.com/anthropics/anthropic-sdk-python/blob/v0.117.0/src/anthropic/types/model_param.py and https://platform.claude.com/docs/en/about-claude/models/extended-thinking-models#limits-and-feature-compatibility

vane/ai/providers/anthropic.py
```diff
@@ -32,6 +63,182 @@
...
+# not be smuggled past pre-dispatch validation through ``extra_body``.
+_VANE_OWNED_REQUEST_FIELDS: frozenset[str] = frozenset({"messages", "model", "stream", "system", "tools"})
+
+# Vane execution options that ride in ``prompt_options`` but are consumed by
+# the UDF machinery and never forwarded to the Messages API.
+_VANE_EXECUTION_OPTIONS: frozenset[str] = frozenset(
+    {
+        "actor_number",
+        "batch_size",
+        "concurrency",
```
kaka11chen, line 113: **P2: Do not accept execution options that are silently dropped**

The validator admits both `concurrency` and `batch_size`, but `AnthropicPrompterDescriptor.get_udf_options()` reads neither: `get_prompter(concurrency=3, batch_size=7)` stores both values while producing `UDFOptions(actor_number=None, batch_size=None)`, and the prompter then filters them out before the API call. The typed `AnthropicProviderOptions(concurrency=...)` path maps correctly, but the documented raw-dict/direct path silently ignores these settings—the same class of behavior this PR is intended to remove. Please map `concurrency` to `actor_number` and carry `batch_size`, or reject unsupported keys, with coverage for the raw-dict path.

kaka11chen (commented):
Two additional inline findings from review of the current head. The existing manual-thinking and tool-choice threads already cover the other two outstanding contract gaps, so they are not duplicated here.

vane/ai/providers/anthropic.py
caomaocao: Fixed in cd029b09: `claude-fable-5`, `claude-mythos-5`, and
`claude-mythos-preview` are now in `_MODEL_UNSUPPORTED_OPTIONS`, matching the
extended-thinking reference ("On Claude Fable 5, Claude Mythos 5, Claude
Mythos Preview, Claude Opus 5, Claude Opus 4.8, Claude Opus 4.7, and Claude
Sonnet 5, non-default temperature, top_p, or top_k values return a 400 error
on every request") — the table's comment now cites that page, and the shared
value set is factored into `_RESTRICTED_SAMPLING_OPTIONS`. Mythos Preview
gets its own entry since the family-prefix match on `claude-mythos-5` cannot
reach it. Regression tests: `temperature=0.2` rejected on all three ids,
`temperature=1.0` still accepted on `claude-fable-5`, and `top_p=1.0`
rejected there.

vane/ai/providers/anthropic.py
caomaocao: Fixed in cd029b09, taking the mapping option: `get_udf_options()` now
resolves `actor_number` from `actor_number` or, when absent, `concurrency` —
the same aliasing `AnthropicProviderOptions.to_descriptor_options()` performs
— and carries `batch_size` into `UDFOptions`. Passing both `actor_number`
and `concurrency` with disagreeing values raises at validation instead of
picking one silently; equal values pass. Tests cover the raw-dict path
(`concurrency=3, batch_size=7` → `UDFOptions(actor_number=3, batch_size=7)`),
the equal-pair case, and the conflict rejection.

vane/ai/providers/anthropic.py
```diff
@@ -28,10 +44,95 @@
...
+        "extra_headers",
+        "extra_query",
+        "inference_geo",
+        "max_tokens",
+        "metadata",
+        "output_config",
+        "service_tier",
+        "stop_sequences",
+        "temperature",
+        "thinking",
```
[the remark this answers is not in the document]
caomaocao: Fixed in cd029b09 (kaka11chen's follow-up review confirmed this thread as an
outstanding gap): a new `_MODEL_REJECTED_THINKING_TYPES` table transcribes
the per-model rejection table from the thinking-troubleshooting reference,
and `_validate_prompt_options` rejects a `thinking.type` the selected family
documents as a 400 — the removed manual `"enabled"` form on the
adaptive-only families (Sonnet 5, Opus 4.7/4.8, Opus 5, Fable 5, Mythos 5),
`"disabled"` on the always-on families (Fable 5, Mythos 5, Mythos Preview),
and `"adaptive"` on the extended-thinking-only families (Haiku 4.5,
Sonnet 4.5, Opus 4.5, Opus 4.1). Mythos Preview still accepts the manual
form and the deprecated-but-working 4.6 usage stays allowed; Opus 5's
effort-conditional `"disabled"` rejection is deliberately not modeled (it
depends on `output_config.effort`, and a false rejection would be worse than
deferring that case to the API). Tests cover all three rejection classes, a
dated-snapshot id, and the still-supported forms.

vane/ai/providers/anthropic.py
```diff
@@ -28,10 +44,95 @@
...
+        "tool_choice",
```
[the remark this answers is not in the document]
caomaocao: Fixed in cd029b09 (kaka11chen's follow-up review confirmed this thread as an
outstanding gap): without `return_format` Vane sends no tool definitions, so
`_validate_option_combinations` now rejects a forced `tool_choice`
(`{"type": "any"}` / `{"type": "tool", ...}`) on that path with an error
naming the two fixes — use `return_format` for structured output or a
no-tools-safe `"auto"`/`"none"` choice, which both remain forwarded. With
`return_format`, a caller-supplied `tool_choice` was already rejected in
1cda38e4, so every combination is now covered pre-dispatch. The
`max_tokens=0` prewarming combination check runs first so a zero-token
request with a forced choice keeps reporting the full offending set. Tests:
`any`/`tool` rejected without `return_format`, `auto`/`none` forwarded.

kaka11chen:
@codex review

kaka11chen (approved):
LGTM

PR #237, opened 2026-07-28
[fix](ray) Publish completion after stream retirement

### LINKED ISSUE #217: [P2][Flaky] Terminal stream weakref can outlive collector retirement in the full fast suite
## Problem

`tests/fast/test_udf_stream_result_collector.py::test_terminal_stream_is_retired_without_waiting_for_collector_shutdown` intermittently fails near the end of the full non-Ray fast suite.

At the failure point, `collector._records == {}` already holds, but the weak reference to `TaskLeaseObjectRefGenerator` remains alive after the test’s 1-second `gc.collect()` polling deadline:

```text
assert collector._records == {}
assert source_ref() is None
E assert <...TaskLeaseObjectRefGenerator object ...> is None
```

This was observed twice while validating an unrelated video-reader-only change based on upstream `main` at `01183216`:

- Full run 1: 1 failed, 5732 passed (plus the normal skips/xfails)
- Full run 2: 1 failed, 5737 passed (plus the normal skips/xfails)

No collector or UDF stream implementation files were changed in those runs.

## Isolation result

The exact test passed when rerun alone, and then passed in 10/10 separate pytest processes:

```bash
python -m pytest tests/fast/test_udf_stream_result_collector.py::test_terminal_stream_is_retired_without_waiting_for_collector_shutdown -q
```

This points to suite-state/load-sensitive object retention or a GC/timing assumption, rather than a deterministic failure in the isolated test.

## Expected behavior

The result of the terminal-stream lifecycle test is deterministic and does not depend on suite load, garbage-collection timing, or a fixed one-second wall-clock window.

kaka11chen:
## Summary

- keep a terminal stream record pending while `RayStreamAdapter.retire()` releases task/source ownership and deletes the ObjectRef stream
- publish the collector `complete` event only after that local retirement finishes
- replace the GC-timing assertion with a deterministic retirement gate that proves completion cannot overtake cleanup

## Root cause

`AsyncResultCollector._maybe_complete_record()` removed the record and published `complete` before calling `record.adapter.retire()`. Consumers could therefore observe an empty collector while the adapter still held strong references to `TaskLeaseObjectRefGenerator`, making the weakref assertion scheduler-dependent in the full fast suite.

## Testing

- `python -m pytest tests/fast/test_udf_stream_result_collector.py` (21 passed)
- related stream adapter and collector tests (33 passed)
- `python -m pytest tests/fast/test_udf_process.py` (6 passed)
- `scripts/run_release_tests.sh` (418 passed, 1 skipped; 7 real-Ray tests passed)
- `scripts/run_fast_tests.sh non-ray` (5895 passed, 28 skipped, 8 xfailed, 1 xpassed)

Closes #217

### DIFF
--- a/duckdb/execution/udf_stream_result_collector.py
+++ b/duckdb/execution/udf_stream_result_collector.py
@@ -791,13 +791,16 @@ class AsyncResultCollector:
         record.adapter.mark_drained()
         key = (record.slot_id, record.submit_id)
         with self._cv:
-            if self._records.pop(key, None) is not record:
+            if self._records.get(key) is not record:
                 return
             record.terminal = True
+        record.adapter.retire()
+        with self._cv:
+            if self._records.pop(key, None) is not record:
+                return
             self._ready_by_slot[record.slot_id].append(_ReadyEvent(record.slot_id, record.submit_id, "complete", None))
             self._cv.notify_all()
         _collector_debug_log("retired", record)
-        record.adapter.retire()
         self._notify_wakeup()
 
     def _cancel_record_control(self, record: _StreamRecord) -> None:
--- a/tests/fast/test_udf_stream_result_collector.py
+++ b/tests/fast/test_udf_stream_result_collector.py
@@ -4,7 +4,6 @@
 from __future__ import annotations
 
 import asyncio
-import gc
 import threading
 import time
 import weakref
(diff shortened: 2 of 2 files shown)


kaka11chen:
@codex review

hubgeter (approved):

QuakeWang (approved):
+1

PR #263, opened 2026-07-29
[fix](distributed) Bound stalled Flight server shutdown

### LINKED ISSUE #258: [P1][Bug] Stalled Flight RPC can block graceful worker shutdown
## Affected area

Distributed execution or Ray

## What happened?

A stalled Arrow Flight RPC can block graceful worker shutdown because the process-local Flight server uses the no-deadline `Shutdown()` call and then unconditionally joins its server thread.

Arrow documents this overload as blocking until current requests finish. A remote `DoGet` client can keep an RPC in flight by stopping consumption, leaving a half-open connection, or remaining blocked behind flow control. The client timeout is optional and defaults to disabled (`flight_timeout_seconds = 0.0`).

`ShutdownLocalFlightServerInternal` holds the process-local service mutex while it calls `FlightServer::Stop()`. While that call is waiting:

- the worker's graceful `finish_shutdown` RPC cannot complete;
- any concurrent local service start, port lookup, or epoch lookup waits on the same mutex;
- the `FlightServer` destructor has the same no-deadline shutdown and join behavior.

The standard Ray worker lifecycle bounds its wait for `finish_shutdown` to 30 seconds and then force-kills the actor. The user-visible result is a failed graceful shutdown after the full timeout and forced actor termination. A direct/static destruction path without that outer Ray timeout can remain blocked indefinitely.

Relevant code at `64736abf8bb1296f75b8f1dc52a5b65524248ba0`:

- `external/duckdb/src/execution/distributed/exchange/flight_server.cpp:312-319,357-367`
- `external/duckdb/src/execution/distributed/exchange/flight_exchange_manager.cpp:235-295`
- `external/duckdb/src/include/duckdb/execution/distributed/exchange/flight_exchange_manager.hpp:44-50`
- `duckdb/runners/ray/fragment_worker_lifecycle.py:465-495`

The pinned Arrow Flight server API describes `Shutdown()` as blocking until current requests finish.

## Reproduction

1. Start a worker's process-local Flight service and create a local-disk shuffle.
2. Open a remote `DoGet` stream for a partition.
3. Stop consuming the stream while leaving the RPC open.
4. Begin the worker's normal prepare/finish shutdown sequence.
5. The worker enters `ShutdownLocalFlightServerInternal` and blocks in `FlightServer::Stop()`.
6. The driver-side `finish_shutdown` wait expires after 30 seconds and force-kills the Ray actor.

When `FlightServer` is destroyed outside that bounded Ray wrapper with an in-flight request, its destructor waits in the same shutdown/join sequence without an outer deadline.

## Environment

- Vane commit: `64736abf8bb1296f75b8f1dc52a5b65524248ba0`
- OS: Linux, x86-64
- Execution mode: Ray/FTE with local-disk Arrow Flight shuffle
- Arrow Flight transport: gRPC

## Additional context

This is related to but not covered by #73/#185. That work introduced explicit process ownership, shutdown, and thread join. The issue here is the liveness behavior of that explicit shutdown when an RPC does not finish.

The normal worker lifecycle fences and drains work before `finish_shutdown`, which limits the usual blast radius. It does not bound the server-side wait inside `Stop()`; the outer actor shutdown timeout supplies the only demonstrated bound in that path.

kaka11chen:
## Summary

- bound Flight server shutdown with a five-second grace deadline shared by explicit stop and destruction
- remove the process-local server from published state under the service mutex, then stop and destroy it outside the lock; concurrent starts fail fast while shutdown is active
- add a real Arrow Flight regression that leaves a 16 MiB `DoGet` stream unconsumed and verifies bounded shutdown, non-blocking endpoint lookup, and non-blocking start rejection

## Root cause

The server used Arrow Flight `Shutdown()` without a deadline and immediately joined its serving thread. A remote client that stopped consuming a `DoGet` stream could keep the gRPC handler active indefinitely. The process-local service mutex remained held across that wait, extending the stall to endpoint lookups and service starts.

The fix passes Arrow a bounded shutdown deadline so gRPC cancels stalled transport calls after the grace period. The process-local lifecycle publishes an empty endpoint before the blocking stop and tracks a `shutting_down` state until the old server has been fully destroyed.

## Verification

- `scripts/format duckdb --changed`
- regression test before fix: failed because shutdown did not finish while the reader remained open
- `scripts/run_native_tests.sh "[distributed][exchange]"` — 47 cases, 616 assertions passed
- `scripts/run_native_tests.sh "[distributed]"` — 116 cases, 3970 assertions passed
- regression test repeated 10 times successfully
- native package rebuilt with the required incremental PEP 517 build directory
- `scripts/run_release_tests.sh` — 448 non-Ray tests passed, 1 skipped; 7 real-Ray tests passed

Closes #258

### DIFF
--- a/external/duckdb/src/execution/distributed/exchange/flight_exchange_manager.cpp
+++ b/external/duckdb/src/execution/distributed/exchange/flight_exchange_manager.cpp
@@ -113,11 +113,13 @@ DuckDBResult<ParsedFlightHost> ParseFlightHost(std::string host) {
 struct LocalFlightServiceState {
 	std::mutex mutex;
 	std::unique_ptr<FlightServer> server;
+	bool shutting_down = false;
 	std::string bind_host;
 	std::string advertise_host;
 	int requested_port = 0;
 	int actual_port = 0;
 	std::string server_epoch;
+	std::chrono::milliseconds shutdown_grace_period {FlightServerConfig::DEFAULT_SHUTDOWN_GRACE_PERIOD};
 };
 
 LocalFlightServiceState &LocalFlightService() {
@@ -315,6 +317,10 @@ DuckDBResult<void> EnsureLocalFlightServerStartedInternal(const FlightServiceCon
 
 	auto &service = LocalFlightService();
 	std::lock_guard<std::mutex> guard(service.mutex);
+	if (service.shutting_down) {
+		return DuckDBResult<void>::err(
+		    DuckDBError::invalid_state_error("process-local Flight service is shutting down"));
+	}
 	if (service.server) {
 		if (service.bind_host != bind_host || service.advertise_host != advertise_host ||
 		    service.requested_port != config.port) {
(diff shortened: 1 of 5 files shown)


kaka11chen:
@codex review

QuakeWang (approved):
LGTM

hubgeter (approved):

Issue #80, opened 2026-07-21
[P1][Bug] UDF terminal events can violate result queue capacity accounting


kaka11chen:
## Problem

Terminal/control events do not consume the same reserved capacity as data events but are still enqueued. A terminal event can fill the queue and make a later valid data event fail the real-time capacity assertion, turning a healthy execution into a capacity-contract error.

PR #221, opened 2026-07-27
[fix](udf) Require output-consumer execution

kaka11chen:
## Summary

- Remove pull-based and fallback UDF result delivery. Every supported relation, scalar, batch, flat-map, and registered table UDF is planned through PhysicalStreamingUDF and must register an output consumer before its first submission.
- Define output capacity for DATA events only; COMPLETE, ERROR, and FINISHED remain control events. Reserve an already scheduled Ray block read so capacity one cannot admit a second DATA read.
- Force Ray backends to block streams and subprocess backends to local shared-memory ref bundles, remove the public streaming_breaker opt-out, and reject stale pull-based physical plans with an explicit capability error.
- Preserve scalar passthrough layouts and registered table UDF column projection under strict lazy-ref delivery.

## Related issue

close: #80

This supersedes #157: the fallback queue whose mixed event accounting it repaired is removed instead of retained.

## Documentation impact

- [x] No user-facing documentation update is required.
- [ ] Documentation is updated in this PR or in a linked website PR.
- [ ] A follow-up issue is required in AstroVela/vane-website.

The removed streaming_breaker option is not referenced by repository documentation. Public stubs, benchmarks, and the video integration are updated together, and the intentional compatibility break is described above.

## Validation

- Incremental Release native build and installation with SKBUILD_BUILD_DIR=build/python-release and uv pip install . --no-build-isolation
- Affected UDF suites: 402 passed
- Subprocess lifecycle, scalar layout, and collector follow-up suites: 260 passed
- Native strict-contract tests: 2 passed
- Fast tests: non-Ray phase passed; shared-Ray 86 passed and 6 skipped; test-owned Ray clusters 6 passed and 3 environment-dependent cases skipped
- Release tests: 414 passed and 1 skipped; real-Ray shard 7 passed
- Full changed-file pre-commit gate passed, including Ruff, mypy, and clang-format
- Real-Ray registered table UDF serialization and multi-column projection probe passed
- git diff --check upstream/main...HEAD
- Two consecutive post-fix review rounds found no remaining issues

## Checklist

- [x] The change is focused and includes tests or a reason tests are unnecessary.
- [x] Public behavior and compatibility impact are documented.
- [x] New dependencies, copied code, model assets, and datasets have compatible licenses and are recorded where required.
- [x] No credentials, private endpoints, personal paths, generated data, model weights, or build artifacts are included.
- [x] Security implications of UDFs, serialization, remote code, network access, and untrusted input have been considered.
[...]

Issue #38, opened 2026-07-20
[P1][Bug] Ray UDF queries exceed default object-store demand budget


kaka11chen:
## What happened?

When `tests/fast/test_ray_e2e.py` is run in isolation, 24 Python-UDF and streaming-batch cases fail with:

```text
ValueError: minimum query demand exceeds desired demand for object_store_bytes
```

Non-UDF scan, group, join, limit, and window paths pass.

The test Ray cluster has a 1 GiB object store. The driver applies the default query fraction of 0.5, giving a desired demand of 512 MiB. Graph inspection for a representative UDF query shows a minimum of about 608 MiB:

- upstream FTE output window: 256 MiB
- downstream FTE output window: 256 MiB
- two UDF actors: 2 x 48 MiB
- total minimum: 608 MiB

The resource coordinator therefore rejects the query before execution.

## Reproduction

```bash
PYTHONPATH=build/python-release .venv/bin/python -m pytest   tests/fast/test_ray_e2e.py -q
```

Observed in isolation: `24 failed, 38 passed, 6 skipped, 1 deselected`.

A representative `test_ray_udf_distributed` case passes when `VANE_QUERY_OBJECT_STORE_FRACTION=1.0` is set.

## Environment

- Vane commit: `2c113bc335e2e0246c89fc72fa467d0fab0b0b95`
- OS: Linux 6.17 x86_64
- Python: 3.12.3
- pytest: 9.1.1
- NumPy: 2.5.1
- PyArrow: 25.0.0
- Ray: 2.56.1

## Additional context

- Test object-store size: `tests/conftest.py`, Ray cluster fixture.
- Default fraction: `duckdb/runners/ray/driver.py`.
- Rejection: `duckdb/runners/ray/cluster_resource_coordinator.py`.

This is separate from the suite-wide Ray GCS fork/ENOMEM cascade: it remains reproducible after Ray startup is healthy.

kaka11chen:
Latest-main audit update for `02389fd568f635e612348e8da135139f4a2468be`:

- the standard fixture still provisions a 1 GiB object store;
- the default query fraction still exposes 512 MiB;
- graph demand still includes fixed actor/task/progress components that can exceed that capacity.

The full Ray E2E group was not rerun in this audit, so the earlier dynamic counts in this issue remain historical evidence rather than a new current-main count. The default sizing contradiction is still present statically.

PR #264, opened 2026-07-29
[fix](udf) Restore concurrent Ray block prefetch

### LINKED ISSUE #260: [P1][Performance] Streaming UDF output capacity serializes Ray block reads
## Problem

The image-classification benchmark regressed by about 2.4x after strict UDF output-capacity accounting was merged. Both the baseline and regressed versions already execute `map_batches` with streaming breaker behavior, so this is a regression inside the streaming output path rather than a physical-plan path change.

## Reproduction

Environment: 10,000 local ImageNet images, `batch_size=100`, one RTX 2080 Ti, identical model/data caches and Python dependencies.

| Revision | Runtime |
| --- | ---: |
| Parent of PR #221 (`5ef1edd2e0`) | 29.27s |
| PR #221 merge commit (`7d7050dfb5`) | 71.27s |
| PR #245 head (`4db06af388`) | 70.89s |

All runs produced 10,000 rows with identical `(image_url, label)` sets.

A controlled run retaining the PR #221 native build and all other package files, while restoring only the previous scheduled-read capacity accounting condition, completed in 29.57s.

## Observed behavior

The default UDF output target is 128 MiB and the per-item reservation is also 128 MiB, producing one DATA-event credit. Since PR #221, a generator block read that has been scheduled but has not completed immediately consumes that credit. Other active streams in the same slot cannot pull their next block/metadata pair until the current output completes its handoff. Image tensor blocks in this benchmark are approximately 57 MiB, and the resulting serialized output handoff reduces downstream GPU utilization.

## Related issues and pull requests

- #80 records the terminal/control-event capacity-accounting correctness problem.
- #221 fixed #80 and is the first bad adjacent revision in the benchmark.
- #38 and #245 cover managed object-byte soft budgets; adjacent testing excludes #245 as the source of this regression.

kaka11chen:
## Summary

- decouple one-block ObjectRef prefetch for each admitted active stream from downstream DATA-event admission, so a slow producer no longer occupies the only output credit before its block is ready
- reserve DATA capacity atomically before pulling paired metadata and retain that reservation through output-lease admission; ready DATA events and in-progress reservations remain within the advertised event capacity
- add capacity-one regression coverage for overlapping stream prefetch, strict metadata admission, and progress when one stream is slow

The streaming breaker remains enabled. This changes the collector scheduling point inside that path rather than selecting a different physical plan.

## Performance

On the 10,000-image classification benchmark with identical model/data caches and dependencies:

| Revision | Runtime |
| --- | ---: |
| Parent of #221 (`5ef1edd2e0`) | 29.27s |
| #245 head before this fix (`4db06af388`) | 70.89s |
| This branch (`f32f570b43`) | 29.57s |

All compared runs produced 10,000 rows with identical `(image_url, label)` sets.

## Validation

- affected collector/adapter/lifecycle/process tests: `339 passed`
- collector concurrency race tests: 50 repeated runs passed
- release suite: `448 passed, 1 skipped, 11 deselected`; real-Ray release shard: `7 passed, 453 deselected`
- complete fast suite: non-Ray `6056 passed, 28 skipped, 104 deselected, 8 xfailed, 1 xpassed`; shared-cluster Ray `87 passed, 6 skipped, 6104 deselected`; all 6 runnable isolated Ray-owner tests passed and the vLLM-only test skipped
- two consecutive clean review passes after the final implementation changes

Closes #260

Related: #80, #221, #38, #245

### DIFF
--- a/duckdb/execution/udf_stream_result_collector.py
+++ b/duckdb/execution/udf_stream_result_collector.py
@@ -107,6 +107,7 @@ class _StreamRecord:
     wait_future: Any | None = None
     completion_future: Any | None = None
     terminal_signal_observed: bool = False
+    data_credit_reserved: bool = False
 
 
 class AsyncResultCollector:
@@ -382,14 +383,11 @@ class AsyncResultCollector:
     def _pending_data_count_locked(self, slot_id: int) -> int:
         ready = sum(1 for event in self._ready_by_slot.get(slot_id, ()) if event.kind == "data")
         in_progress = sum(
-            1
-            for record in self._records.values()
-            if record.slot_id == slot_id
-            and (record.phase != "block" or (record.wait_kind == "data" and record.wait_future is not None))
+            1 for record in self._records.values() if record.slot_id == slot_id and record.data_credit_reserved
         )
         return ready + in_progress
 
-    def _may_read_block_locked(self, record: _StreamRecord) -> bool:
+    def _may_prefetch_block_locked(self, record: _StreamRecord) -> bool:
         capacity = self._capacity_by_slot.get(record.slot_id)
         if capacity is None or capacity.rows <= 0:
             return False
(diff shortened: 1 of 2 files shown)


kaka11chen:
Full image benchmark validation on `f32f570b43b6defb81be0ddcc386e46ebf50b4c8`:

- Dataset: full local ImageNet metadata, 803,580 rows
- Configuration: 1x RTX 2080 Ti, `BATCH_SIZE=100`, `NUM_GPU_NODES=1`
- Runtime: 848.52s (14:14.71 wall), about 947 images/s end-to-end; steady-state pipeline about 954 images/s
- Process: exit status 0, max RSS 747,800 KiB, 0 swaps, no fatal error signatures
- Materialized output: exactly 803,580 rows, 803,580 distinct URLs; 0 missing, extra, duplicate, null, or empty rows
- Input/output URL multiset SHA-256: `42a2f0bd4874ed06eb9c83bd083d27f68ddca72e5dce26211ba0da57e79d78de` on both sides
- Labels: all rows are valid members of the 1,000 ResNet18 categories; all 1,000 categories are represented
- Determinism: 0 URL/label mismatches against the prior 10,000-row run
- Independent reference check: 0 mismatches across 512 evenly spaced rows re-inferred directly with torchvision ResNet18
- Parquet footer and Vane direct-write commit manifest both validate; manifest reports `rows_copied=803580`.

kaka11chen:
Replacement implementation: #271. It restores the same image-benchmark throughput using Ray Data-style non-consuming `ray.wait` readiness and downstream soft admission, so block refs remain in Ray Core until capacity admits them. #271 supersedes this client-side prefetch approach.

PR #287, opened 2026-07-29
[fix](datasource) Preserve Arrow scan state across batches

### LINKED ISSUE #272: [P0][Bug] datasource_scan silently drops rows after 2048 in every Arrow batch
## Affected area

DuckDB or SQL execution; distributed execution or Ray uses the same scan path after deserialization.

## What happened?

`datasource_scan` returns at most the first 2,048 rows from every Arrow `RecordBatch` yielded by `DataSourceTask.execute()`. Rows after the first DuckDB standard vector are silently omitted without a query error.

The public `DataSourceTask.execute()` contract recommends yielding batches of approximately 10 MB. A narrow 10 MB batch can contain substantially more than 2,048 rows, so following the documented batching guidance can produce incomplete query results.

## Reproduction

```python
from collections.abc import Iterator

import duckdb
import pyarrow as pa
from duckdb.datasource import DataSource, DataSourceTask, read_datasource


class OneLargeBatchTask(DataSourceTask):
    def execute(self) -> Iterator[pa.RecordBatch]:
        yield pa.record_batch(
            {"value": pa.array(range(5000), type=pa.int64())}
        )


class OneLargeBatchSource(DataSource):
    @property
    def schema(self):
        return {"value": "BIGINT"}

    def get_tasks(self):
        yield OneLargeBatchTask()


con = duckdb.connect()
result = read_datasource(OneLargeBatchSource(), con=con).aggregate(
    "count(*) AS n, min(value) AS min_v, max(value) AS max_v"
).fetchone()
print(result)
```

Actual result:

```text
(2048, 0, 2047)
```

Expected result:

```text
(5000, 0, 4999)
```

The same row result was observed with the Ray runner and with the local runner.

## Environment

- OS: Linux
- Architecture: x86_64
- Python: 3.12
- Latest target verified: `upstream/main` at `a389186ad187a0f7de8627eef7339d0f7208229d` (2026-07-29)
- Reproduction build: `64736abf8bb1296f75b8f1dc52a5b65524248ba0`; the relevant datasource and Arrow scan files have no diff against target commit `a389186ad187a0f7de8627eef7339d0f7208229d`
- Ray: reproduced through the Ray execution path; also reproduced with `VANE_RUNNER=local`

## Additional context

At target commit `a389186ad187a0f7de8627eef7339d0f7208229d`:

- `external/duckdb/src/function/table/datasource_scan.cpp:203-224` calls `GetNextChunk()`, sets `output_size` to `min(STANDARD_VECTOR_SIZE, arrow_array.length)`, constructs a fresh `ArrowScanLocalState`, converts once, and returns.
- The fresh `ArrowScanLocalState` starts with `chunk_offset = 0`.
- `DataSourceScanLocalState` contains only the current stream and an exhausted flag; it contains no current Arrow array or per-array offset.
- On the next `DataSourceScanGetData()` call, `GetNextChunk()` advances the Arrow stream to its next array.
- In contrast, the native Arrow table scan retains `ArrowScanLocalState` across calls and increments `state.chunk_offset` after each output vector.
- `STANDARD_VECTOR_SIZE` is 2,048.
- `git blame` attributes the affected `DataSourceScanGetData()` block to init commit `57d4e3c166e`. Datasource factory identity change #196 does not modify this block.

Relevant source:

- https://github.com/AstroVela/vane/blob/a389186ad187a0f7de8627eef7339d0f7208229d/external/duckdb/src/function/table/datasource_scan.cpp#L195-L247
- https://github.com/AstroVela/vane/blob/a389186ad187a0f7de8627eef7339d0f7208229d/external/duckdb/src/include/duckdb/function/table/datasource_scan.hpp#L87-L92
- https://github.com/AstroVela/vane/blob/a389186ad187a0f7de8627eef7339d0f7208229d/external/duckdb/src/function/table/arrow.cpp#L192-L212
- https://github.com/AstroVela/vane/blob/a389186ad187a0f7de8627eef7339d0f7208229d/external/duckdb/src/include/duckdb/common/vector_size.hpp#L16-L20
- https://github.com/AstroVela/vane/blob/a389186ad187a0f7de8627eef7339d0f7208229d/duckdb/datasource/__init__.py#L46-L55

## Impact

Every yielded batch longer than 2,048 rows loses its suffix. The query completes with apparently valid but incomplete data, so downstream counts, aggregates, joins, writes, and model inputs can all be silently incorrect. Both connection-local execution and the worker-deserialized execution path reach the affected table function.

### LINKED ISSUE #275: [P0][Bug] datasource_scan treats an empty Arrow RecordBatch as end-of-stream
## Affected area

DuckDB or SQL execution. Distributed execution and Ray use the same `datasource_scan` table-function path after deserialization.

## What happened?

When `DataSourceTask.execute()` yields an empty Arrow `RecordBatch` followed by a non-empty batch, `datasource_scan` treats the empty batch as the end of the current stream. It never requests the following batch from that task.

The query completes without an error but silently omits every row after the empty batch.

## Reproduction

```python
from collections.abc import Iterator

import duckdb
import pyarrow as pa
from duckdb.datasource import DataSource, DataSourceTask, read_datasource


class EmptyThenRowsTask(DataSourceTask):
    def execute(self) -> Iterator[pa.RecordBatch]:
        yield pa.record_batch(
            {"value": pa.array([], type=pa.int64())}
        )
        yield pa.record_batch(
            {"value": pa.array([10, 20, 30], type=pa.int64())}
        )


class EmptyThenRowsSource(DataSource):
    @property
    def schema(self):
        return {"value": "BIGINT"}

    def get_tasks(self):
        yield EmptyThenRowsTask()


con = duckdb.connect()
result = read_datasource(EmptyThenRowsSource(), con=con).fetchall()
print(result)
```

Actual result:

```text
[]
```

Expected result:

```text
[(10,), (20,), (30,)]
```

## Environment

- OS: Linux
- Architecture: x86_64
- Python: 3.12
- Reproduction build DuckDB SourceID: `1da652c1ca`
- Latest target inspected: `upstream/main` at `a389186ad187a0f7de8627eef7339d0f7208229d` (2026-07-29)
- Runner used for the reproduction: local

## Additional context

At target commit `a389186ad187a0f7de8627eef7339d0f7208229d`, `DataSourceScanGetData()` only enters the conversion branch when both `arrow_array.release` is set and `arrow_array.length > 0`. A released zero-length batch takes the same path as an exhausted stream: the current stream is reset and the scanner advances to another task or finishes.

This is related to #272 because both concern Arrow batch consumption in `datasource_scan`, but it is independently observable with batches smaller than DuckDB standard vector size.

## Impact

A valid empty batch anywhere before the end of a task causes all later batches from that task to be silently dropped. Downstream counts, aggregates, joins, writes, and model inputs can therefore be incomplete without a query error.

kaka11chen:
## Summary

- replace the implicit datasource scan flow with explicit `NEED_TASK`, `NEED_BATCH`, `SCANNING`, and `EXHAUSTED` states
- retain `ArrowScanLocalState` across table-function calls so every DuckDB vector in a large Arrow RecordBatch is emitted
- treat only an Arrow array with `release == nullptr` as end-of-stream and skip valid zero-row batches without abandoning the task
- cover connection-local and Ray worker execution with regressions for batches larger than 2,048 rows and for leading, middle, consecutive, and trailing empty batches

## Related issue

close: #272
close: #275

## Documentation impact

- [x] No user-facing documentation update is required. The change restores the existing `DataSourceTask.execute()` contract.
      `AstroVela/vane-website`.

## Validation

- incremental non-editable Release package rebuild with `uv pip install . --no-build-isolation`
- `python -m pytest -v tests/fast/test_datasource_distributed_scan.py` — 12 passed
- `scripts/run_release_tests.sh` — 455 passed and 1 skipped in the non-Ray shard; 10 passed in the real-Ray shard
- `VANE_FAST_TEST_EXCLUDE_GPU=1 scripts/run_fast_tests.sh` — 6,100 passed, 28 skipped, 8 xfailed, and 1 xpassed in the non-Ray shard; 88 passed and 2 skipped in the shared-Ray shard; all isolated real-Ray shards passed
- `scripts/run_native_tests.sh "[distributed]"` — 130 test cases and 4,156 assertions passed
- manual four-thread boundary matrix — 10,241 exact rows across 2,047/2,048/2,049/4,097-row batches, leading/middle/trailing empty batches, an empty-only task, multiple tasks, and `BIGINT` plus `VARCHAR` columns

## Checklist

- [x] New dependencies, copied code, model assets, and datasets have compatible licenses and are recorded where required. No such additions are included.
- [x] Security implications of UDFs, serialization, remote code, network access, and untrusted input have been considered. The change adds no new trust boundary.

### DIFF
--- a/external/duckdb/src/function/table/datasource_scan.cpp
+++ b/external/duckdb/src/function/table/datasource_scan.cpp
@@ -164,27 +164,35 @@ static unique_ptr<GlobalTableFunctionState> DataSourceScanInitGlobal(ClientConte
 // ── Init Local ─────────────────────────────────────────────────────
 // Each pipeline thread gets its own local state. On init, grab first task.
 
+static void DataSourceScanStartNextTask(const DataSourceScanBindData &bind_data, DataSourceScanGlobalState &gstate,
+                                        DataSourceScanLocalState &lstate) {
+	D_ASSERT(lstate.state == DataSourceScanLocalState::ScanState::NEED_TASK);
+	D_ASSERT(!lstate.stream);
+
+	auto idx = gstate.next_task_idx.fetch_add(1);
+	if (idx >= gstate.total_tasks) {
+		lstate.state = DataSourceScanLocalState::ScanState::EXHAUSTED;
+		return;
+	}
+
+	auto &pickled = bind_data.pickled_tasks[idx];
+	auto stream_wrapper = make_uniq<ArrowArrayStreamWrapper>();
+	RequireProduceStream(bind_data.produce_stream)(pickled.c_str(), pickled.size(),
+	                                               &stream_wrapper->arrow_array_stream);
+	lstate.stream = std::move(stream_wrapper);
+	lstate.state = DataSourceScanLocalState::ScanState::NEED_BATCH;
+}
+
 static unique_ptr<LocalTableFunctionState> DataSourceScanInitLocal(ExecutionContext &context,
                                                                    TableFunctionInitInput &input,
                                                                    GlobalTableFunctionState *global_state) {
 	auto &bind_data = input.bind_data->Cast<DataSourceScanBindData>();
 	auto &gstate = global_state->Cast<DataSourceScanGlobalState>();
-	auto result = make_uniq<DataSourceScanLocalState>();
-
-	// Grab first task for this thread
-	auto idx = gstate.next_task_idx.fetch_add(1);
-	if (idx >= gstate.total_tasks) {
-		result->exhausted = true;
-		return std::move(result);
+	auto result = make_uniq<DataSourceScanLocalState>(context.client);
+	for (idx_t i = 0; i < bind_data.arrow_table.GetColumns().size(); i++) {
+		result->scan_state.column_ids.push_back(i);
 	}
-
-	// Produce stream for this task
-	auto &pickled = bind_data.pickled_tasks[idx];
-	auto stream_wrapper = make_uniq<ArrowArrayStreamWrapper>();
-	RequireProduceStream(bind_data.produce_stream)(pickled.c_str(), pickled.size(),
-	                                               &stream_wrapper->arrow_array_stream);
-	result->stream = std::move(stream_wrapper);
-
+	DataSourceScanStartNextTask(bind_data, gstate, *result);
 	return std::move(result);
 }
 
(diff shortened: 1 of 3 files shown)


kaka11chen:
@codex review

hubgeter (approved):
LGTM

PR #296, opened 2026-07-30
[fix](distributed) Partition window inputs by semantics

### LINKED ISSUE #281: [P0][Wrong result] Distributed window operators compute independently per input task
## Affected area

Distributed translation and execution of `PhysicalWindow` and `PhysicalStreamingWindow` when the child has multiple execution tasks.

## What happens

`TranslateWindow` and `TranslateStreamingWindow` directly create a `WindowNode` or `StreamingWindowNode`. During task production, the physical window operator is appended independently to every input task.

There is no exchange or gather before the operator. As a result, each task sees only its local input partition and evaluates the window independently.

## Trigger conditions

This occurs when both conditions hold:

- the optimized physical plan still contains `PhysicalWindow` or `PhysicalStreamingWindow`; and
- the child produces more than one actual execution task.

Multiple files alone are not sufficient to prove the trigger because scan splits can be coalesced, and some simple window expressions can be rewritten by the optimizer into other operators.

## Reproduction

Create four Parquet files containing consecutive values:

- file 1: `x = 0, 1, 2`
- file 2: `x = 3, 4, 5`
- file 3: `x = 6, 7, 8`
- file 4: `x = 9, 10, 11`

Force four scan tasks with:

```text
VANE_RAY_SCAN_TASK_SIZE_GROUPING=0
VANE_RAY_SCAN_TASK_MIN_PARTITION_NUM=4
VANE_FTE_DYNAMIC_SCAN_MAX_SPLITS_PER_PARTITION=1
```

Then run:

```sql
SELECT x, row_number() OVER (ORDER BY x) AS rn
FROM read_parquet('data/*.parquet');
```

The distributed plan contains a window node with four partitions above a scan with four tasks.

Expected `rn` values are `1` through `12`. The observed output resets the row number at every task boundary:

```text
(0, 1), (1, 2), (2, 3),
(3, 1), (4, 2), (5, 3),
(6, 1), (7, 2), (8, 3),
(9, 1), (10, 2), (11, 3)
```

The same per-task reset is observable with `PhysicalStreamingWindow`, for example `row_number() OVER ()`.

Ordered cumulative frames and offset functions are also local to each task. For example, with one common partition key, `SUM(x) OVER (PARTITION BY k ORDER BY x ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW)` and `LAG(x)` restart at the boundaries before `x = 3`, `6`, and `9`.

## Source evidence

- [`TranslateWindow` and `TranslateStreamingWindow`](https://github.com/AstroVela/vane/blob/main/src/data/physical_plan/plan_builder/translator_misc_unary.cpp#L118-L134)
- [Window task production appends the operator independently to each input task](https://github.com/AstroVela/vane/blob/main/src/data/physical_plan/nodes/window.cpp#L38-L106)

## Impact

Window queries can return silently incorrect results on distributed inputs, including repeated row numbers or ranks, incomplete partition aggregates, and broken cross-task `LAG`/`LEAD` behavior.

## Environment

Verified against upstream `main` at `eb6c2bae4c4e7b521e655cf8fbefa0a56c94d0a5`. The focused runtime reproduction used a native build whose relevant source files are byte-identical to that revision.

kaka11chen:
## Summary

- gather global and streaming windows before evaluation so they observe the complete input
- hash-partition `PARTITION BY` windows on the complete partition-key set while retaining distributed parallelism
- reuse only directly verified compatible partitioning and preserve accurate window schema/clustering metadata
- add constructor guardrails plus native topology and forced multi-scan-task Ray regressions

## Validation

- `VANE_NATIVE_BUILD_JOBS=8 scripts/run_native_tests.sh '[distributed]'` (138 cases, 4250 assertions)
- `python -m pytest tests/fast/test_ray_e2e.py -k 'windows_span_multiple_scan_tasks or test_ray_window or test_ray_streaming_window' -vv` (3 passed)
- `scripts/run_release_tests.sh` (455 passed + 10 real-Ray passed)
- `scripts/run_fast_tests.sh non-ray` (6083 passed)
- `scripts/run_fast_tests.sh shared-ray` (93 passed)
- `scripts/run_fast_tests.sh owner-ray` (7 passed, 1 optional-dependency skip)

Closes #281

### DIFF
--- a/external/duckdb/src/execution/distributed/pipeline_node/translator_misc_unary.cpp
+++ b/external/duckdb/src/execution/distributed/pipeline_node/translator_misc_unary.cpp
@@ -3,19 +3,25 @@
 
 #include "duckdb/execution/distributed/pipeline_node/translator.hpp"
 
+#include <algorithm>
+
+#include "duckdb/common/exception.hpp"
 #include "duckdb/execution/distributed/pipeline_node/expression_scan.hpp"
 #include "duckdb/execution/distributed/pipeline_node/pivot.hpp"
+#include "duckdb/execution/distributed/pipeline_node/shuffles/repartition.hpp"
 #include "duckdb/execution/distributed/pipeline_node/streaming_udf_passthrough.hpp"
 #include "duckdb/execution/distributed/pipeline_node/table_inout.hpp"
 #include "duckdb/execution/distributed/pipeline_node/unnest.hpp"
 #include "duckdb/execution/distributed/pipeline_node/window.hpp"
 #include "duckdb/execution/operator/aggregate/physical_streaming_window.hpp"
 #include "duckdb/execution/operator/aggregate/physical_window.hpp"
+#include "duckdb/execution/operator/exchange/repartition.hpp"
 #include "duckdb/execution/operator/projection/physical_pivot.hpp"
 #include "duckdb/execution/operator/projection/physical_tableinout_function.hpp"
 #include "duckdb/execution/operator/projection/physical_udf_inout.hpp"
 #include "duckdb/execution/operator/projection/physical_unnest.hpp"
 #include "duckdb/execution/operator/scan/physical_expression_scan.hpp"
+#include "duckdb/planner/expression/bound_window_expression.hpp"
 
 namespace duckdb {
 namespace distributed {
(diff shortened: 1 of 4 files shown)


kaka11chen:
@codex review

QuakeWang (approved):
+1

PR #298, opened 2026-07-30
[fix](relation) Preserve explode column order

### LINKED ISSUE #276: [P1][Bug] explode changes column order across Relation binding paths and corrupts positional inserts
## Summary

On upstream `main` at `c143f8e0cdac2a7bd6d7987ebbf0644312619609`, `UnnestRelation` has two binding representations with different output-column orders when the exploded column is not the final input column.

Direct binding replaces the list/array column in place. SQL serialization emits `* EXCLUDE(column), unnest(column) AS column`, which appends the exploded column to the end. Downstream Relation operators and statement consumers can select the serialized representation, so composing an otherwise unchanged `explode()` relation changes its schema order.

For a positional `insert_into()` whose target columns have compatible types, the reordered values are accepted and stored under the wrong target columns without an error.

## Affected code

- [`UnnestRelation::GetQueryNode` emits `* EXCLUDE(col)` followed by `unnest(col) AS col`](https://github.com/AstroVela/vane/blob/c143f8e0cdac2a7bd6d7987ebbf0644312619609/external/duckdb/src/main/relation/unnest_relation.cpp#L26-L42)
- [`UnnestRelation::Bind` projects the scalar element at the original child-column index](https://github.com/AstroVela/vane/blob/c143f8e0cdac2a7bd6d7987ebbf0644312619609/external/duckdb/src/main/relation/unnest_relation.cpp#L45-L119)
- [`UnnestRelation::InheritsColumnBindings` returns `false`](https://github.com/AstroVela/vane/blob/c143f8e0cdac2a7bd6d7987ebbf0644312619609/external/duckdb/src/include/duckdb/main/relation/unnest_relation.hpp#L34-L36)
- [`RequiresDirectRelationBinding` selects SQL serialization for a serializable, non-inheriting, single-source child](https://github.com/AstroVela/vane/blob/c143f8e0cdac2a7bd6d7987ebbf0644312619609/external/duckdb/src/main/relation.cpp#L329-L334)
- [`LimitRelation::Columns` delegates to the child while its `Bind` can take the serialized path](https://github.com/AstroVela/vane/blob/c143f8e0cdac2a7bd6d7987ebbf0644312619609/external/duckdb/src/main/relation/limit_relation.cpp#L48-L67)
- [`InsertRelation::Bind` consumes the child's serializable query node](https://github.com/AstroVela/vane/blob/c143f8e0cdac2a7bd6d7987ebbf0644312619609/external/duckdb/src/main/relation/insert_relation.cpp#L28-L42)
- [`CreateTableRelation::Bind` consumes the same serialized representation](https://github.com/AstroVela/vane/blob/c143f8e0cdac2a7bd6d7987ebbf0644312619609/external/duckdb/src/main/relation/create_table_relation.cpp#L31-L50)

## Reproduction

```python
import os

import duckdb

con = duckdb.connect()
base = con.sql(
    "SELECT 10::INTEGER AS x, "
    "[20, 21]::INTEGER[] AS a, "
    "30::INTEGER AS y"
)

exploded = base.explode("a")
limited = exploded.limit(10)

print(exploded.columns)
print(exploded.fetchall())
print(limited.columns)
print(limited.fetchall())

con.execute("CREATE TABLE sink(x INTEGER, a INTEGER, y INTEGER)")
base.explode("a").insert_into("sink")
print(con.sql("SELECT * FROM sink ORDER BY x, a, y").fetchall())
```

Observed output:

```text
['x', 'a', 'y']
[(10, 20, 30), (10, 21, 30)]
['x', 'a', 'y']
[(10, 30, 20), (10, 30, 21)]
[(10, 30, 20), (10, 30, 21)]
```

The direct relation exposes and returns `x, a, y`. Adding `limit()` leaves `LimitRelation::Columns()` reporting `x, a, y`, while execution returns the serialized `x, y, a` value order. The positional insert then stores `y` values in target column `a` and exploded `a` values in target column `y`.

## Other confirmed surfaces

Using the same input:

- `filter()` and `distinct()` retain child metadata `x, a, y` while returning values in `x, y, a` order.
- `order()`, projection with `*`, and `SELECT *` through a relation replacement scan expose `x, y, a`.
- `create()` creates the table schema as `x, y, a`.
- positional union consumes the serialized `x, y, a` representation of the exploded branch.

## Impact and trigger conditions

- Silent wrong-column writes through `insert_into()` when the target column count matches and the reordered values are assignment-compatible.
- Relation metadata can disagree with the physical result tuple order.
- Schema order changes after semantically order-preserving Relation operations.
- Positional set operations can combine different logical columns.

The mismatch requires the exploded column to have at least one following input column. Incompatible target types can turn the insert case into a visible binding or conversion error instead of a silent write.

## Validation status

Confirmed by static analysis against upstream `main` commit `c143f8e0cdac2a7bd6d7987ebbf0644312619609` on 2026-07-29 and reproduced with the local native build. All source files involved in the reproduced paths are byte-identical between the runtime checkout and that upstream commit.

## Related work

- #138 changed Relation input-binding and SQL-serialization behavior broadly; current `main` after that merge retains this two-path `UnnestRelation` column-order mismatch.
- #122 is the latest-main audit backlog and does not contain a dedicated explode/unnest column-order item.

kaka11chen:
## Summary

- serialize `UnnestRelation` with `SELECT * REPLACE (unnest(column) AS column)` so SQL rebinding preserves the exploded column's original position
- resolve the explode target by its direct-binding position, then address the SQL subquery's deduplicated input alias; this preserves the pre-existing exact first-match behavior for duplicate and case-only names without binding the wrong suffix-collision column
- restore duplicate public aliases explicitly at relation serialization boundaries, including composed filter/set-operation trees, while leaving the common unique-name path unwrapped
- mirror DuckDB's SQL UNNEST binder by casting ARRAY inputs to LIST before direct `PhysicalUnnest` execution
- add regressions for LIST/ARRAY parity, duplicate non-target and target names, case variants, suffix collisions, filter/union composition, limit, and positional insert

The public API is unchanged: users still call `relation.explode("a")`. The implementation builds a structured AST; it does not construct SQL from user-provided identifiers.

## Related issue

Closes #276

## Documentation impact

      `AstroVela/vane-website`.

This fixes internal Relation binding and serialization without changing the documented API.

## Validation

- incremental Release package install with `SKBUILD_BUILD_DIR="$PWD/build/python-release" SKBUILD_CMAKE_BUILD_TYPE=Release uv pip install . --no-build-isolation`
- focused explode and SQL-query regressions: 35 passed
- complete relational API suite: 406 passed, 1 skipped
- generated LIST/ARRAY serialization matrix: 1,524 direct-versus-serialized cases passed across positions, aliases, suffix collisions, filters, limits, and distinct
- manual composition matrix: nested explode, filter/order/project/distinct/union, replacement scan, insert/create, non-SQL child binding, special identifiers, NULL/empty inputs, and unchanged error paths
- `scripts/run_native_tests.sh "[relation_api]"`: 35 test cases and 651 assertions passed
- `scripts/run_release_tests.sh`: 455 passed, 1 skipped in the non-Ray shard; 10 passed in the real-Ray shard
- `scripts/run_fast_tests.sh`: 6,118 passed, 28 skipped, 8 xfailed, 1 xpassed in the non-Ray shard; 92 passed and 6 skipped in the shared-Ray shard; all 7 runnable isolated owner-Ray tests passed and the optional vLLM test skipped
- pre-commit hooks for all four changed files
- `git diff --check`
- two consecutive clean code-review passes after the final implementation and test run

## Checklist

- [x] New dependencies, copied code, model assets, and datasets have compatible licenses and are recorded where required. No such additions are included.
- [x] Security implications of UDFs, serialization, remote code, network access, and untrusted input have been considered. The change builds an internal AST and adds no trust boundary.
- [x] Native changes were compiled; Python, native, release, formatting, and static checks passed.

### DIFF
--- a/external/duckdb/src/include/duckdb/main/relation.hpp
+++ b/external/duckdb/src/include/duckdb/main/relation.hpp
@@ -280,6 +280,8 @@ protected:
 	static BoundStatement BindSelectNodeOnChild(Binder &binder, Relation &child, unique_ptr<SelectNode> select_node);
 	static unique_ptr<SelectNode> WrapQueryNode(unique_ptr<QueryNode> query_node, const string &alias,
 	                                            const vector<ColumnDefinition> &columns);
+	static unique_ptr<QueryNode> RestoreDuplicateColumnAliases(unique_ptr<QueryNode> query_node, const string &alias,
+	                                                           const vector<ColumnDefinition> &columns);
 	static unique_ptr<LogicalOperator> PlanRelationFilter(Binder &binder, unique_ptr<Expression> condition,
 	                                                      unique_ptr<LogicalOperator> child);
 	static void ExpandRelationFilter(Binder &binder, unique_ptr<ParsedExpression> &condition);
--- a/external/duckdb/src/main/relation.cpp
+++ b/external/duckdb/src/main/relation.cpp
@@ -294,7 +294,20 @@ unique_ptr<QueryNode> Relation::TryGetSerializableQueryNode(Binder &binder) {
 	if (!CanSerializeToQueryNodeInternal(binder)) {
 		return nullptr;
 	}
-	return GetQueryNode();
+	return RestoreDuplicateColumnAliases(GetQueryNode(), GetAlias(), Columns());
+}
+
+unique_ptr<QueryNode> Relation::RestoreDuplicateColumnAliases(unique_ptr<QueryNode> query_node, const string &alias,
+                                                              const vector<ColumnDefinition> &columns) {
+	case_insensitive_set_t output_names;
+	for (auto &column : columns) {
+		if (!output_names.insert(column.Name()).second) {
+			// SQL table scopes deduplicate names so expressions can address every
+			// column. Restore duplicate public aliases at this query boundary.
+			return WrapQueryNode(std::move(query_node), alias, columns);
+		}
+	}
+	return query_node;
 }
 
 unique_ptr<QueryResult> Relation::Execute() {
--- a/external/duckdb/src/main/relation/unnest_relation.cpp
+++ b/external/duckdb/src/main/relation/unnest_relation.cpp
@@ -12,6 +12,7 @@
 #include "duckdb/planner/operator/logical_projection.hpp"
 #include "duckdb/planner/expression/bound_unnest_expression.hpp"
 #include "duckdb/planner/expression/bound_columnref_expression.hpp"
+#include "duckdb/common/string_util.hpp"
 #include "duckdb/common/types.hpp"
 
 namespace duckdb {
@@ -25,19 +26,16 @@ UnnestRelation::UnnestRelation(shared_ptr<Relation> child_p, string column_name_
 
 unique_ptr<QueryNode> UnnestRelation::GetQueryNode() {
 	// Fallback AST path (used by default Relation::Bind if custom Bind not called).
-	// Builds: SELECT * EXCLUDE(col), unnest(col) AS col FROM (child)
+	// Builds: SELECT * REPLACE (unnest(col) AS col) FROM (child)
 	auto result = make_uniq<SelectNode>();
 	result->from_table = GetTableRefForSerialization(*child);
 
 	auto star = make_uniq<StarExpression>();
-	star->exclude_list.insert(QualifiedColumnName(column_name));
-	result->select_list.push_back(std::move(star));
-
 	vector<unique_ptr<ParsedExpression>> unnest_args;
 	unnest_args.push_back(make_uniq<ColumnRefExpression>(column_name));
 	auto unnest_func = make_uniq<FunctionExpression>("unnest", std::move(unnest_args));
-	unnest_func->SetAlias(column_name);
-	result->select_list.push_back(std::move(unnest_func));
+	star->replace_list.emplace(column_name, std::move(unnest_func));
+	result->select_list.push_back(std::move(star));
 
 	return std::move(result);
 }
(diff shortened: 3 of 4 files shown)


kaka11chen:
@codex review

hubgeter:
[P2] Preserve duplicate non-target column names across serialization

`UnnestRelation::Bind` only rejects duplicate names that case-insensitively match the exploded target. A different duplicated child name still passes, but `GetQueryNode()` wraps the child in a subquery and uses `* REPLACE`. Subquery binding then deduplicates those names through `BindContext::AliasColumnNames`, so the serialized relation no longer has the same schema as the directly bound relation.

Reproduced on `28fe7cfd40`:

```python
con = duckdb.connect()
rel = con.sql(
    "SELECT [20, 21]::INTEGER[] AS a, 10 AS x, 30 AS x"
).explode("a")

assert rel.columns == ["a", "x", "x"]
assert con.sql(rel.sql_query()).columns == ["a", "x", "x_1"]
```

The rows are now in the correct order, so the original positional corruption is fixed, but relation metadata, `sql_query()`, and name-based downstream consumers can still disagree. This also conflicts with the existing duplicate-name serialization invariant in `test_sql_query_preserves_result_modifier_boundaries`.

Please preserve the original aliases with an explicit outer projection, using the same deduplicated-input/original-output pattern as `Relation::WrapQueryNode`, or mark this child schema as non-serializable and keep direct binding. Please add a regression with a unique explode target and a duplicated non-target column that compares both columns and rows after rebinding `sql_query()`.

kaka11chen:
Addressed the P2 feedback in c0a261e4e4. Duplicate public aliases are now restored with the existing deduplicated-input/original-output projection pattern both at the explode node (for composed raw query trees) and at the final relation serialization boundary. The target itself is selected by position and referenced through its deduplicated SQL binding, which also covers case-only duplicates and suffix collisions without changing direct-binding resolution. Added direct-versus-serialized regressions for duplicated non-target names, filter and union composition, target suffix collisions, duplicate targets, and LIST/ARRAY parity.

kaka11chen:
@codex review

hubgeter (approved):
LGTM

PR #301, opened 2026-07-30
[fix](relation) Normalize ARRAY inputs for explode

### LINKED ISSUE #299: [P1][Bug] Relation explode on fixed-size ARRAY raises InternalException and invalidates the database
## Summary

On upstream `main` at `c0735b8b70b8978b86102d7de48e52dbe0cffba7`, `DuckDBPyRelation.explode()` accepts a fixed-size `ARRAY` column during binding and exposes the scalar element type, but executing the relation raises an `InternalException`:

```text
INTERNAL Error: Expected unified vector format of type LIST, but found type ARRAY
```

The internal error invalidates the database instance. A subsequent query on the same connection fails with `FatalException` and requires the database to be restarted.

The same Relation operation succeeds for `LIST`, and SQL `unnest()` succeeds for the same fixed-size `ARRAY` value.

## Affected area

DuckDB or SQL execution / Python Relation API

## Affected code

- [`UnnestRelation::Bind` accepts both `LIST` and `ARRAY`, then attaches the original typed column reference directly to `BoundUnnestExpression`](https://github.com/AstroVela/vane/blob/c0735b8b70b8978b86102d7de48e52dbe0cffba7/external/duckdb/src/main/relation/unnest_relation.cpp#L66-L81)
- [The normal SQL `UNNEST` binder normalizes an `ARRAY` child before constructing the bound unnest expression](https://github.com/AstroVela/vane/blob/c0735b8b70b8978b86102d7de48e52dbe0cffba7/external/duckdb/src/planner/binder/expression/bind_unnest_expression.cpp#L139-L148)
- [`PhysicalUnnest` reads the expression result through `ListVector` and `list_entry_t`](https://github.com/AstroVela/vane/blob/c0735b8b70b8978b86102d7de48e52dbe0cffba7/external/duckdb/src/execution/operator/projection/physical_unnest.cpp#L92-L125)
- [The unified-vector type check raises the observed `InternalException` when the requested LIST storage does not match ARRAY storage](https://github.com/AstroVela/vane/blob/c0735b8b70b8978b86102d7de48e52dbe0cffba7/external/duckdb/src/include/duckdb/common/types/vector.hpp#L52-L60)

## Reproduction

```python
import os

import duckdb

# Control: Relation explode on LIST succeeds.
list_con = duckdb.connect()
print(
    list_con.sql("SELECT [1, 2]::INTEGER[] AS a")
    .explode("a")
    .fetchall()
)

# Control: SQL UNNEST on the same fixed-size ARRAY succeeds.
sql_con = duckdb.connect()
print(
    sql_con.sql("SELECT unnest([1, 2]::INTEGER[2]) AS a")
    .fetchall()
)

# Failing direct Relation path.
con = duckdb.connect()
relation = con.sql("SELECT [1, 2]::INTEGER[2] AS a").explode("a")
print(relation.columns)
print(relation.types)

try:
    print(relation.fetchall())
except Exception as exc:
    print(type(exc).__name__, str(exc).splitlines()[0])

try:
    print(con.sql("SELECT 1").fetchall())
except Exception as exc:
    print(type(exc).__name__, str(exc).splitlines()[0])
```

Observed output:

```text
[(1,), (2,)]
[(1,), (2,)]
['a']
[INTEGER]
InternalException INTERNAL Error: Expected unified vector format of type LIST, but found type ARRAY
FatalException FATAL Error: Failed: database has been invalidated because of a previous fatal error. The database must be restarted prior to being used again.
```

Expected output for the Relation `ARRAY` case is `[(1,), (2,)]`, with the database remaining usable afterward.

## Confirmed scope

- Reproduces with both a constant `ARRAY` expression and an `INTEGER[2]` table column.
- Binding succeeds and reports the expected scalar output schema; failure occurs when rows are fetched.
- `relation.sql_query()` produces a serializable SQL query that executes successfully. The failure is specific to direct Relation binding/execution.
- `LIST` input is unaffected.
- SQL `unnest()` of fixed-size `ARRAY` input is unaffected.

## Environment

- OS: Linux 6.11.0-24-generic
- Architecture: x86-64
- Python: 3.12.13
- Vane commit: `c0735b8b70b8978b86102d7de48e52dbe0cffba7`
- DuckDB engine: `v1.6.0-dev2`, Source ID `bb37a8f949`
- Build: Release wheel built directly from the clean upstream commit
- `VANE_RUNNER`: `local`

## Validation status

Reproduced from an independent native wheel built from the exact current upstream `main` commit on 2026-07-30. Searches across open and closed repository issues found no report matching the exception, fixed-size `ARRAY` Relation execution, or this direct-binding failure.

## Related issue

- #276 concerns output column ordering across `explode()` binding paths. It does not cover this `ARRAY` execution error.

kaka11chen:
## Summary

- normalize fixed-size `ARRAY` children to `LIST` before constructing the directly bound `BoundUnnestExpression`
- reuse the same `BoundCastExpression::AddArrayCastToList` path as SQL `UNNEST`, while preserving Relation direct binding and output column order
- add a multi-row Relation API regression covering fixed arrays, NULL arrays, NULL elements, scalar output types, and continued connection usability

## Root cause

`UnnestRelation::Bind` accepted both `LIST` and `ARRAY`, but attached the original `ARRAY`-typed column reference directly to `BoundUnnestExpression`. `PhysicalUnnest` consumes LIST vector storage, so direct Relation execution raised an internal LIST/ARRAY format mismatch and invalidated the database. The normal SQL binder already normalizes ARRAY children before reaching that operator.

## Validation

- confirmed the new regression fails before the implementation with `Expected unified vector format of type LIST, but found type ARRAY`
- `VANE_RUNNER=local .venv/bin/python -m pytest -q tests/fast/relational_api/test_explode_array.py` — 1 passed
- `VANE_RUNNER=local .venv/bin/python -m pytest -q tests/fast/relational_api` — 394 passed, 1 skipped
- `scripts/run_release_tests.sh` — 455 passed, 1 skipped in the non-Ray shard; 10 passed in the real-Ray shard
- `pre-commit run --from-ref upstream/main --to-ref HEAD` — passed
- manual matrix: LIST, ARRAY, nested ARRAY, and NULL ARRAY all execute successfully and leave the connection usable
- three-way merge simulation with #298/#276 completed without conflicts

Closes #299

### DIFF
--- a/external/duckdb/src/main/relation/unnest_relation.cpp
+++ b/external/duckdb/src/main/relation/unnest_relation.cpp
@@ -10,6 +10,7 @@
 #include "duckdb/planner/binder.hpp"
 #include "duckdb/planner/operator/logical_unnest.hpp"
 #include "duckdb/planner/operator/logical_projection.hpp"
+#include "duckdb/planner/expression/bound_cast_expression.hpp"
 #include "duckdb/planner/expression/bound_unnest_expression.hpp"
 #include "duckdb/planner/expression/bound_columnref_expression.hpp"
 #include "duckdb/common/types.hpp"
@@ -76,7 +77,8 @@ BoundStatement UnnestRelation::Bind(Binder &binder) {
 	}
 
 	// 4. Create BoundUnnestExpression: unnest(col_ref) → element_type
-	auto col_ref = make_uniq<BoundColumnRefExpression>(column_name, list_type, child_bindings[unnest_col_idx]);
+	auto col_ref = BoundCastExpression::AddArrayCastToList(
+	    binder.context, make_uniq<BoundColumnRefExpression>(column_name, list_type, child_bindings[unnest_col_idx]));
 	auto unnest_expr = make_uniq<BoundUnnestExpression>(element_type);
 	unnest_expr->child = std::move(col_ref);
 
--- a/tests/fast/relational_api/test_explode_array.py
+++ b/tests/fast/relational_api/test_explode_array.py
@@ -1,0 +1,30 @@
+# SPDX-FileCopyrightText: 2026 Vane contributors
+# SPDX-License-Identifier: Apache-2.0
+
+import duckdb
+
+
+def test_explode_fixed_array(duckdb_cursor):
+    relation = duckdb_cursor.sql(
+        """
+        SELECT *
+        FROM (
+            VALUES
+                (10, [20, 21]::INTEGER[2], 30),
+                (11, NULL::INTEGER[2], 31),
+                (12, [22, NULL]::INTEGER[2], 32)
+        ) data(x, a, y)
+        """
+    )
+
+    exploded = relation.explode("a")
+
+    assert exploded.columns == ["x", "a", "y"]
+    assert exploded.types == [duckdb.sqltypes.INTEGER] * 3
+    assert exploded.fetchall() == [
+        (10, 20, 30),
+        (10, 21, 30),
+        (12, 22, 32),
+        (12, None, 32),
+    ]
+    assert duckdb_cursor.sql("SELECT 1").fetchall() == [(1,)]


kaka11chen:
@codex review

hubgeter (approved):
LGTM

kaka11chen:
Superseded by #298, merged in 693c661a6dd812743623c4d9bbf86860ada8f75d. The latest main now contains the same long-term ARRAY-to-LIST normalization via BoundCastExpression::AddArrayCastToList before PhysicalUnnest, together with fixed-size ARRAY regression coverage. Closing this PR to avoid duplicating the implementation. Its additional NULL-focused cases can still be considered separately as a test-only follow-up.

PR #303, opened 2026-07-30
[fix](executor) Preserve finalization state across flush re-entry

### LINKED ISSUE #292: [P1][Wrong result] Reentrant cache flush can call FinalExecute again after FINISHED
## Summary

The caching-operator flush state machine can call `FinalExecute` again after it already returned `FINISHED`.

The active `ExecutionBatch` implementation advances `flushing_idx` only when downstream pushing returns `NEED_MORE_INPUT`. If a completed final output is interrupted by a blocking or multi-output downstream stage, the completed state is not preserved across re-entry.

## Verified baseline

- Repository: `AstroVela/vane`
- Commit: `91bf7c1ec666ea75cd535775780957c1306d39be`
- Status: statically confirmed; no deterministic runtime reproducer was run as part of this report

## Evidence

- [Active batch flush implementation](https://github.com/AstroVela/vane/blob/91bf7c1ec666ea75cd535775780957c1306d39be/external/duckdb/src/parallel/pipeline_executor.cpp#L246-L321).
- [The equivalent non-batch implementation has the same state transition](https://github.com/AstroVela/vane/blob/91bf7c1ec666ea75cd535775780957c1306d39be/external/duckdb/src/parallel/pipeline_executor.cpp#L158-L244).
- [`PhysicalStreamingWindow::FinalExecute` can return `FINISHED` with a non-empty delayed output](https://github.com/AstroVela/vane/blob/91bf7c1ec666ea75cd535775780957c1306d39be/external/duckdb/src/execution/operator/aggregate/physical_streaming_window.cpp#L737-L758).

The problematic sequence is:

1. `FinalExecuteBatch` returns `FINISHED` with output.
2. `should_flush_current_idx` becomes `false`.
3. Pushing that output returns `BLOCKED` or `HAVE_MORE_OUTPUT`, so the function returns before incrementing `flushing_idx`.
4. On re-entry with no in-process operator, the code changes `should_flush_current_idx` back to `true`.
5. The same operator's `FinalExecuteBatch` is invoked again.

When re-entry first resumes an in-process downstream operator, the synthetic `HAVE_MORE_OUTPUT` finalize result can likewise overwrite the prior completed state.

## Impact

The second call can be harmless for implementations that clear their buffered state before returning `FINISHED`. It can duplicate final rows or repeat finalization side effects for implementations whose `FINISHED` call produced output but did not make repeated finalization idempotent.

The trigger window requires both a `RequiresFinalExecute` operator and downstream interruption/multi-output behavior, so ordinary non-blocking local sinks are less likely to expose it.

## Reproduction path

Use a finalizing operator that returns `FINISHED` with a non-empty chunk, followed by a sink or operator that blocks or returns `HAVE_MORE_OUTPUT` while that final chunk is being pushed.

kaka11chen:
## Summary

- persist whether a finalizing operator has more output or has already finished while downstream output drains
- advance after final output without invoking `FinalExecute`/`FinalExecuteBatch` again, while preserving `HAVE_MORE_OUTPUT` and `BLOCKED` retry behavior
- add a default-running partial-scheduler regression using streaming `LEAD` followed by `UNNEST`

## Root cause

`FinalExecute`/`FinalExecuteBatch` could return `FINISHED` with output, but a downstream `HAVE_MORE_OUTPUT` or `BLOCKED` re-entry lost that completed result and invoked the same finalizer again. The boolean state could not distinguish “drain before calling again” from “drain before advancing.”

The fix replaces that boolean with an explicit three-state transition:

- `NEEDS_FINALIZE`
- `DRAINING_MORE_OUTPUT`
- `DRAINING_FINAL_OUTPUT`

The saved state is resolved only after downstream operators and any blocked sink retry have fully drained the current output.

## Testing

- clean `upstream/main` (`c0735b8b70b8978b86102d7de48e52dbe0cffba7`) red proof: the new regression returned `(200001, 1400007)` instead of `(200000, 1400000)`
- patched regression: passed 10 consecutive runs and is discoverable in the default `[api]` suite
- batch and non-batch flush paths: regression passed in both modes
- related caching/flush tests: 3 test cases and 400,009 assertions passed in both modes
- multi-finalizer and multi-batch `HAVE_MORE_OUTPUT` diagnostics passed
- required incremental native install succeeded; installed extension returned `(200000, 1400000)`
- `scripts/run_release_tests.sh`: 455 non-Ray tests passed (1 skipped), then 10 real-Ray tests passed
- full native suite: 4,551 passed and 19 failed; the same 19 failures reproduce on the clean base commit (including the same empty-vector, ordering, and unavailable test-extension failures), so no new full-suite failure was introduced
- `scripts/format duckdb --changed` and `git diff --check` passed

Closes #292

### DIFF
--- a/external/duckdb/src/include/duckdb/parallel/pipeline_executor.hpp
+++ b/external/duckdb/src/include/duckdb/parallel/pipeline_executor.hpp
@@ -145,10 +145,19 @@ private:
 	//! source_chunk should be sent through the pipeline
 	bool next_batch_blocked = false;
 
+	enum class OperatorFlushState : uint8_t {
+		//! The current operator's FinalExecute callback should be invoked
+		NEEDS_FINALIZE,
+		//! The current output must be drained before invoking FinalExecute again
+		DRAINING_MORE_OUTPUT,
+		//! The current output must be drained before advancing to the next operator
+		DRAINING_FINAL_OUTPUT
+	};
+
 	//! Current operator being flushed
 	idx_t flushing_idx;
-	//! Whether the current flushing_idx should be flushed: this needs to be stored to make flushing code re-entrant
-	bool should_flush_current_idx = true;
+	//! Persist the current operator's finalization result while downstream operators drain its output
+	OperatorFlushState operator_flush_state = OperatorFlushState::NEEDS_FINALIZE;
 	//! Whether this executor should route source/operator/sink calls through ExecutionBatch callbacks
 	bool use_execution_batches = false;
 
--- a/external/duckdb/src/parallel/pipeline_executor.cpp
+++ b/external/duckdb/src/parallel/pipeline_executor.cpp
@@ -172,46 +172,53 @@ bool PipelineExecutor::TryFlushCachingOperators(ExecutionBudget &chunk_budget) {
 			continue;
 		}
 
-		// This slightly awkward way of increasing the flushing idx is to make the code re-entrant: We need to call this
-		// method again in the case of a Sink returning BLOCKED.
-		if (!should_flush_current_idx && in_process_operators.empty()) {
-			should_flush_current_idx = true;
+		// Resolve the persisted FinalExecute result only after downstream operators (and any blocked sink retry)
+		// have drained the current output.
+		if (in_process_operators.empty()) {
+			switch (operator_flush_state) {
+			case OperatorFlushState::NEEDS_FINALIZE:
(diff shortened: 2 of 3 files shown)


kaka11chen:
@codex review

QuakeWang (approved):
LGTM

hubgeter (approved):
LGTM

PR #304, opened 2026-07-30
[fix](vllm) Retire producers after downstream LIMIT

### LINKED ISSUE #274: [P1][Bug] Downstream LIMIT can strand the vLLM producer-finalization barrier
## Summary

On upstream `main` at `c143f8e0cdac2a7bd6d7987ebbf0644312619609`, a parallel `VLLM_PROJECT` query can remain parked indefinitely when a downstream `STREAMING_LIMIT` finishes one or more pipeline tasks before those tasks execute `PhysicalVLLM::FinalExecute`.

The global barrier records the scheduled producer count in `active_threads`, but a producer contributes to `finished_threads` only from `FinalExecute`. A producer whose finalization is skipped leaves the count permanently short. If another producer does enter `FinalExecute`, it can park waiting for `all_tasks_finished`, while the executor never receives the matching global `FinishedSubmitting` transition.

Current `main` uses one-shot scheduler wakeups in this path. The steady-state symptom is therefore a blocked/stalled query, not the empty-chunk CPU busy loop in the original characterization.

## Affected code

- [`PipelineMaxThreadsResolved` records the scheduled producer count](https://github.com/AstroVela/vane/blob/c143f8e0cdac2a7bd6d7987ebbf0644312619609/external/duckdb/src/execution/operator/projection/physical_vllm.cpp#L202-L209)
- [`ThreadFinishedSubmitting` requires an exact `finished_threads == active_threads` count](https://github.com/AstroVela/vane/blob/c143f8e0cdac2a7bd6d7987ebbf0644312619609/external/duckdb/src/execution/operator/projection/physical_vllm.cpp#L299-L311)
- [`PhysicalVLLM::FinalExecute` contributes that per-producer completion and parks through a one-shot wakeup](https://github.com/AstroVela/vane/blob/c143f8e0cdac2a7bd6d7987ebbf0644312619609/external/duckdb/src/execution/operator/projection/physical_vllm.cpp#L646-L699)
- [`PipelineExecutor` starts cache-operator flushing at `finished_processing_idx` after early completion](https://github.com/AstroVela/vane/blob/c143f8e0cdac2a7bd6d7987ebbf0644312619609/external/duckdb/src/parallel/pipeline_executor.cpp#L158-L169)
- [A downstream operator returning `FINISHED` sets that finished-processing index](https://github.com/AstroVela/vane/blob/c143f8e0cdac2a7bd6d7987ebbf0644312619609/external/duckdb/src/parallel/pipeline_executor.cpp#L893-L896)
- [`PhysicalStreamingLimit::Execute` returns `FINISHED` once the global limit has been reached](https://github.com/AstroVela/vane/blob/c143f8e0cdac2a7bd6d7987ebbf0644312619609/external/duckdb/src/execution/operator/helper/physical_streaming_limit.cpp#L50-L68)

## Failure mechanism

1. The pipeline schedules multiple producer tasks and publishes that number to the vLLM global operator state.
2. A downstream streaming limit reaches its global row limit and returns `FINISHED` for a task.
3. Early-finish flushing starts at the downstream finished-processing index, so the upstream vLLM operator is not finalized for that task.
4. The skipped producer never increments `finished_threads`.
5. A different producer reaches `PhysicalVLLM::FinalExecute`, contributes its completion, observes that the global count is still short, and eventually returns `BLOCKED` after arming a one-shot executor wakeup.
6. Once outstanding executor events are drained, no transition can make the missing producer completion appear. The global submission-complete state remains false and the parked finalizer has no further event to wake it.

## Trigger shape

A representative shape is a parallel source feeding vLLM with a downstream streaming limit, for example:

```sql
SELECT *
FROM (
    SELECT vllm(model, prompt) AS result
    FROM parallel_source
)
LIMIT 5;
```

The failure is scheduling-dependent. It requires a mixed task outcome in which at least one producer exits through the downstream limit before vLLM finalization and at least one other producer enters the vLLM finalization path.

## Impact

- A query can remain blocked after the requested limited output has already become available.
- vLLM executor resources remain attached to the stalled query until a higher-level teardown path runs.
- A normal client interrupt only sets the context interrupt flag; it does not itself reschedule an already parked pipeline task, so this state can also delay observable cancellation.

## Validation status

Confirmed by static control-flow and state-machine analysis against upstream `main` commit `c143f8e0cdac2a7bd6d7987ebbf0644312619609` on 2026-07-29. A deterministic end-to-end reproducer was not completed during this review because the mixed task outcome depends on scheduling.

## Related tracking

- #50 tracks vLLM execution and lifecycle races broadly.
- #122 is the latest-main audit backlog.

Neither item identifies this downstream early-termination versus producer-completion-count mismatch as a dedicated defect.

hubgeter:
## Summary

Downstream operators such as LIMIT can finish a pipeline before the vLLM operator reaches FinalExecute. The skipped completion notification leaves the global producer-finalization barrier waiting for a producer that will never arrive.

VLLMOperatorState now shares an idempotent FinishSubmitting transition between the normal FinalExecute path and local state finalization. Early downstream completion retires the producer without submitting its buffered work, so the barrier completes while LIMIT retains its short-circuit behavior.

A native regression covers the downstream LIMIT path and verifies both the exact-once producer notification and early input cutoff.

## Related issue

close: #274

## Documentation impact

- [ ]  A follow-up issue is required in AstroVela/vane-website.

This is an internal native lifecycle correction with no public API or configuration change.

## Validation

- PATH="$PWD/.venv/bin:$PATH" scripts/format workspace --changed — passed.
- Incremental Release reinstall with SKBUILD_BUILD_DIR="$PWD/build/python-release", SKBUILD_CMAKE_BUILD_TYPE=Release, and extension autoload/autoinstall disabled — passed.
- python -m pytest tests/fast/test_vllm_native_regressions.py — 26 passed.
- python -m pytest tests/fast/test_vllm_runtime_fixes.py tests/fast/test_vllm_sampling_param_fold.py — 82 passed.
- scripts/run_release_tests.sh — non-Ray: 455 passed, 1 skipped, 14 deselected; real-Ray: 10 passed, 460 deselected.
- Final focused LIMIT regression rerun — 1 passed.
- Runtime DuckDB SourceID 27b0637cd4 matched the source tree 27b0637cd4d5980fa24e16cfc9cfd7abd65c1ef0.
- git diff --check — passed.

No real-model or GPU-backed vLLM inference was run; native execution used the recording executor test double.

## Checklist

- [x] New dependencies, copied code, model assets, and datasets have compatible licenses and are recorded where required. None were added.
- [x] Security implications of UDFs, serialization, remote code, network access, and untrusted input have been considered. The change only adjusts native operator lifecycle bookkeeping and adds no new input or execution surface.

### DIFF
--- a/external/duckdb/src/execution/operator/projection/physical_vllm.cpp
+++ b/external/duckdb/src/execution/operator/projection/physical_vllm.cpp
@@ -341,6 +341,24 @@ struct VLLMOperatorState : public OperatorState {
 	VLLMOperatorState(ClientContext &context, const Expression &prompt_expr) : prompt_executor(context, prompt_expr) {
 	}
 
+	void FinishSubmitting(ClientContext &context, VLLMGlobalOperatorState &gstate) {
+		if (finished_submitting) {
+			return;
+		}
+		finished_submitting = true;
+		gstate.ThreadFinishedSubmitting(context);
+	}
+
+	void Finalize(const PhysicalOperator &op, ExecutionContext &context) override {
+		if (!op.op_state) {
+			throw InternalException("vllm operator state is missing during local finalization");
+		}
+		// A downstream operator can finish the pipeline before this operator's
+		// FinalExecute runs. Retire this producer without submitting buffered
+		// work so the global producer barrier can still complete.
+		FinishSubmitting(context.client, op.op_state->Cast<VLLMGlobalOperatorState>());
+	}
+
 	ExpressionExecutor prompt_executor;
 
 	vector<BufferedVLLMInput> buffer;
@@ -660,8 +678,7 @@ OperatorFinalizeResultType PhysicalVLLM::FinalExecute(ExecutionContext &context,
 			gstate.EnsureExecutor(context);
 			PopAndSubmitTasks(context, gstate, state, 0);
 		}
-		state.finished_submitting = true;
-		gstate.ThreadFinishedSubmitting(context.client);
+		state.FinishSubmitting(context.client, gstate);
 	}
 
 	if (!gstate.HasExecutor()) {
--- a/tests/fast/test_vllm_native_regressions.py
+++ b/tests/fast/test_vllm_native_regressions.py
@@ -99,7 +99,7 @@ class _DeferredWakeupExecutor(_RecordingExecutor):
             callback()
 
 
-def _run_recording_sql(monkeypatch, prompts, options, *, executor=None, threads=1):
+def _run_recording_sql(monkeypatch, prompts, options, *, executor=None, threads=1, query_suffix=""):
     import duckdb
     import duckdb.execution.vllm as vllm
 
@@ -119,7 +119,10 @@ def _run_recording_sql(monkeypatch, prompts, options, *, executor=None, threads=
         )
         encoded = json.dumps(options, separators=(",", ":"))
         rows = con.execute(
-            "SELECT id, prompt, vllm(prompt, 'recording-model', '" + encoded + "') AS generated FROM vllm_input"
+            "SELECT id, prompt, vllm(prompt, 'recording-model', '"
+            + encoded
+            + "') AS generated FROM vllm_input"
+            + query_suffix
         ).fetchall()
         return executor, rows
     finally:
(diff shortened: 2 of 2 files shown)


kaka11chen:
@codex review

kaka11chen (approved):
LGTM

PR #308, opened 2026-07-30
[fix](distributed) Preserve streaming UDF output schema

### LINKED ISSUE #283: [P1][Bug] Row-preserving multi-column streaming UDF fails across a LIMIT coordinator
## Affected area

A row-preserving streaming batch UDF that produces multiple physical output columns, followed by a distributed materialized coordinator such as `LIMIT`.

## What happens

The row-preserving streaming UDF node exposes a schema built only from `output_types[0]`, even though its physical output contains the passthrough columns plus the UDF result columns.

A following `LIMIT` inherits that schema when constructing its materialized coordinator. The exchange source is consequently configured for one type while the actual exchange partition contains multiple columns, and execution fails with a partition type mismatch.

## Trigger conditions

The verified scope is the row-preserving expression/batch UDF path, where the physical UDF output consists of passthrough columns and one or more UDF result columns. A result-only multi-column `Relation.map_batches` result is physically encoded as a single `STRUCT` in the tested path and is not the same case.

The child must also have multiple actual execution tasks so that `LIMIT` uses its distributed coordinator path.

## Reproduction

Use four Parquet inputs and force four scan tasks:

```text
VANE_RAY_SCAN_TASK_SIZE_GROUPING=0
VANE_RAY_SCAN_TASK_MIN_PARTITION_NUM=4
VANE_FTE_DYNAMIC_SCAN_MAX_SPLITS_PER_PARTITION=1
VANE_UDF_TARGET_MAX_BATCH_BYTES=16777216
```

A minimal Python shape is:

```python
import pyarrow as pa
import vane


def add_one(table):
    return pa.table({"y": [value + 1 for value in table["x"].to_pylist()]})

base = con.sql("SELECT x FROM read_parquet('data/*.parquet')")
expr = vane.func.batch(
    add_one,
    inputs={"x": vane.col("x")},
    schema={"y": "INTEGER"},
    row_preserving=True,
)
relation = base.select(vane.col("x"), expr.alias("y")).limit(5)
list(runner.run_iter_tables(relation))
```

The distributed plan has this shape:

```text
StreamingLimit [partitions=4]
+- StreamingUDF [partitions=4]
   +- Project
      +- ScanTaskSource [4 scan tasks]
```

Expected result: five rows with two columns, `x` and `y`.

Observed result: the FTE task retries and ultimately fails with:

```text
flight exchange partition types mismatch
```

## Source evidence

- [Streaming UDF passthrough node schema construction](https://github.com/AstroVela/vane/blob/main/src/data/physical_plan/nodes/streaming_udf_passthrough.hpp#L89-L95)
- [The node retains the full physical output type list](https://github.com/AstroVela/vane/blob/main/src/data/physical_plan/nodes/streaming_udf_passthrough.hpp#L19-L29)
- [Row-preserving UDF planning creates passthrough and result output types](https://github.com/AstroVela/vane/blob/main/src/data/physical_plan/plan_builder/plan_udf_project.cpp#L370-L401)
- [`LIMIT` copies the child configuration and enters the coordinator path](https://github.com/AstroVela/vane/blob/main/src/data/physical_plan/nodes/limit.cpp#L158-L216)
- [Materialized coordinator exchange construction](https://github.com/AstroVela/vane/blob/main/src/data/physical_plan/nodes/materialized_coordinator.cpp#L24-L39)

## Impact

A supported composition of a row-preserving multi-column UDF and `LIMIT` fails on distributed input instead of returning a result. Other downstream operators that derive exchange types from the node schema may be affected by the same mismatch.

## Environment

Verified against upstream `main` at `eb6c2bae4c4e7b521e655cf8fbefa0a56c94d0a5`. The focused runtime reproduction used a native build whose relevant source files are byte-identical to that revision, with a 2 GiB Ray object store.

kaka11chen:
## Summary

- build the distributed `StreamingUDF` node schema from every physical output type instead of only `output_types[0]`
- add a native regression test for multi-column streaming UDF schemas
- add a four-scan-task Ray regression covering a row-preserving batch UDF followed by a distributed `LIMIT` coordinator

## Root cause

The row-preserving streaming UDF emits passthrough columns together with its UDF result columns, but its pipeline-node configuration advertised only the first type. `StreamingLimit` copied that incomplete schema into the materialized coordinator, so the Flight exchange source expected one column while the exchange partitions contained multiple columns.

Using the shared `MakeSchemaRef` helper preserves the existing empty- and single-column behavior and represents multi-column output as a struct schema that the coordinator expands back into the complete physical type list.

## Testing

- `VCPKG_INSTALLED_DIR=/home/kaka/astrovela_vane/vcpkg_installed scripts/run_native_tests.sh "Streaming UDF passthrough schema preserves all output columns"`
- `VCPKG_INSTALLED_DIR=/home/kaka/astrovela_vane/vcpkg_installed scripts/run_native_tests.sh "[distributed]"` (140 cases, 4273 assertions)
- `python -m pytest tests/fast/test_ray_e2e.py::test_ray_row_preserving_batch_udf_limit_preserves_output_schema`
- `scripts/run_release_tests.sh` (459 non-Ray passed, 1 skipped; 10 real-Ray passed)
- `VANE_FAST_TEST_EXCLUDE_GPU=1 scripts/run_fast_tests.sh non-ray` (5943 passed)
- `VANE_FAST_TEST_EXCLUDE_GPU=1 scripts/run_fast_tests.sh shared-ray` (91 passed)
- `VANE_FAST_TEST_EXCLUDE_GPU=1 scripts/run_fast_tests.sh owner-ray` (7 independent owner tests passed)

Fixes #283

### DIFF
--- a/external/duckdb/src/include/duckdb/execution/distributed/pipeline_node/streaming_udf_passthrough.hpp
+++ b/external/duckdb/src/include/duckdb/execution/distributed/pipeline_node/streaming_udf_passthrough.hpp
@@ -21,7 +21,7 @@ public:
 	                            vector<column_t> projected_input, optional_idx ordinality_idx,
 	                            vector<LogicalType> output_types, idx_t estimated_cardinality)
 	    : ctx_(InheritPipelineNodeContext(child, node_id, "StreamingUDF")),
-	      config_(BuildSchema(output_types), child ? child->config().execution_config() : DuckDBExecutionConfigRef(),
+	      config_(MakeSchemaRef(output_types), child ? child->config().execution_config() : DuckDBExecutionConfigRef(),
 	              child ? child->config().clustering_spec() : ClusteringSpec::unknown_with_num_partitions(1)),
 	      child_(std::move(child)), function_(std::move(function)), bind_data_(std::move(bind_data)),
 	      column_ids_(std::move(column_ids)), projected_input_(std::move(projected_input)),
@@ -87,13 +87,6 @@ public:
 	}
 
 private:
-	static SchemaRef BuildSchema(const vector<LogicalType> &output_types) {
-		if (output_types.empty()) {
-			return nullptr;
-		}
-		return std::make_shared<duckdb::LogicalType>(output_types[0]);
-	}
-
 	PipelineNodeContext ctx_;
 	PipelineNodeConfig config_;
 	PipelineNodeRef child_;
--- a/external/duckdb/test/execution/distributed/test_physical_plan_translator.cpp
+++ b/external/duckdb/test/execution/distributed/test_physical_plan_translator.cpp
@@ -59,6 +59,7 @@
 #include "duckdb/execution/distributed/pipeline_node/expression_scan.hpp"
 #include "duckdb/execution/distributed/pipeline_node/shuffles/repartition.hpp"
 #include "duckdb/execution/distributed/pipeline_node/sort.hpp"
+#include "duckdb/execution/distributed/pipeline_node/streaming_udf_passthrough.hpp"
 #include "duckdb/execution/distributed/pipeline_node/window.hpp"
 #include "duckdb/planner/expression/bound_window_expression.hpp"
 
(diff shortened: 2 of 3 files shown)


kaka11chen:
@codex review

QuakeWang (approved):
LGTM

PR #322, opened 2026-07-31
[fix](execution) Preserve materialized result order

### LINKED ISSUE #320: [P1][Wrong result] Materialized order-preserving scans can reorder output when batch indexes are unavailable
## Affected area

DuckDB or SQL execution

## What happened?

On the latest `main` (`d938a8b6c846b321df3a644610c6aba542b9543f`), a materialized order-preserving query can return rows in completion-dependent order when its source is parallel but does not support batch-index partitioning.

`PhysicalResultCollector::GetResultCollector` reaches the order-preserving, no-batch-index branch, whose comment says to use a single-threaded result collector. The materialized path nevertheless constructs `PhysicalMaterializedCollector(data, true)`. The collector therefore permits a parallel sink, creates one local collection per pipeline task, and appends those collections to the global result in `Combine` arrival order without an ordering key.

For `datasource_scan`, tasks are registered in a defined sequence and the table function is parallel, but it does not provide `get_partition_data`. Consequently, `PhysicalTableScan::SupportsPartitioning` returns false for the batch-index request. With `preserve_insertion_order=true`, delayed tasks can therefore produce a silently reordered materialized result.

Expected behavior: when insertion-order preservation is enabled, the result follows datasource task registration order.

Actual behavior: the result contains all rows, but task blocks are ordered by pipeline completion/combination timing.

The confirmed scope is narrower than all sorted or distributed queries:

- materialized result collection is involved;
- the plan requires insertion-order or fixed-order preservation;
- at least one source cannot provide a batch index;
- the pipeline remains parallelizable.

Ordinary `ORDER BY` and TopN sources support batch-index partitioning and select the batch collector, so this report does not claim that those paths are generally reordered. The streaming result branch also selects a non-parallel buffered collector in the same no-batch-index case.

## Reproduction

```python
import os
import time
from collections.abc import Iterator

os.environ["VANE_RUNNER"] = "local-fast"

import duckdb
import pyarrow as pa
from duckdb.datasource import DataSource, DataSourceTask, read_datasource


class DelayedTask(DataSourceTask):
    def __init__(self, task_id: int) -> None:
        self.task_id = task_id

    def execute(self) -> Iterator[pa.RecordBatch]:
        # Tasks are registered 0 -> 3 and deliberately finish in reverse order.
        time.sleep((3 - self.task_id) * 0.2)
        yield pa.record_batch(
            {"task_id": pa.array([self.task_id], type=pa.int64())}
        )


class DelayedSource(DataSource):
    @property
    def schema(self) -> dict[str, str]:
        return {"task_id": "BIGINT"}

    def get_tasks(self) -> Iterator[DataSourceTask]:
        for task_id in range(4):
            yield DelayedTask(task_id)


for threads in (1, 4):
    con = duckdb.connect(
        config={
            "threads": threads,
            "preserve_insertion_order": True,
        }
    )
    values = [
        row[0]
        for row in read_datasource(DelayedSource(), con=con).fetchall()
    ]
    print(f"threads={threads}: {values}")
    con.close()
```

Example output:

```text
threads=1: [0, 1, 2, 3]
threads=4: [2, 3, 1, 0]
```

Repeated four-thread materialized runs also produced:

```text
[3, 2, 1, 0]
[3, 2, 1, 0]
[3, 2, 1, 0]
```

Adding an ordinary `ORDER BY task_id` control returned `[0, 1, 2, 3]` with four threads.

## Environment

- OS: Linux 6.17.0-29-generic
- Architecture: x86_64
- Python: 3.12.3
- PyArrow: 25.0.0
- Vane inspected: `d938a8b6c846b321df3a644610c6aba542b9543f`
- Runtime-tested engine tree: `ec43c81b6f958ba0e1da4674329502a1dcb6cc55`, from `c0a261e4e4a2b86684c1df224d3fc8db1388b02b`
- `VANE_RUNNER`: `local-fast`

The collector selection, materialized collector, datasource scan, table scan partitioning, physical-plan order checks, and pipeline scheduling paths are byte-identical between the runtime-tested engine tree and the inspected latest `main` revision.

## Additional context

Relevant paths on the inspected revision:

- `external/duckdb/src/execution/operator/helper/physical_result_collector.cpp`, lines 43-48
- `external/duckdb/src/execution/operator/helper/physical_materialized_collector.cpp`, lines 34-47 and 75-76
- `external/duckdb/src/include/duckdb/function/table/datasource_scan.hpp`, lines 70-84
- `external/duckdb/src/function/table/datasource_scan.cpp`, lines 276-285
- `external/duckdb/src/execution/operator/scan/physical_table_scan.cpp`, lines 264-269
- `external/duckdb/src/execution/physical_plan/plan_insert.cpp`, lines 36-68

This is a silent ordering-contract violation rather than row loss: all rows remain present, but their observable order can contradict `preserve_insertion_order=true`.

kaka11chen:
## Summary

- serialize materialized result collection when insertion order must be preserved but the plan cannot provide batch indexes
- leave unordered parallel collection and ordered batch collection unchanged
- add a deterministic native regression covering a multi-threaded, parallel, order-preserving source without batch-index partitioning

## Validation

- `scripts/run_native_tests.sh "Result collector serializes ordered plans without batch indexes" -s` (1 case, 7 assertions)
- negative control with the old constructor flag fails the regression at `collector.ParallelSink()`
- `scripts/run_native_tests.sh "[distributed][executor]"` (13 cases, 182 assertions)
- `scripts/run_native_tests.sh "[distributed]"` (150 cases, 4390 assertions)
- delayed datasource reproduction with 4 threads returned `[0, 1, 2, 3]` in 5 consecutive runs
- `scripts/run_release_tests.sh` (460 base tests and 10 real-Ray tests passed; 1 artifact-only test skipped)

Closes #320

### DIFF
--- a/external/duckdb/src/execution/operator/helper/physical_result_collector.cpp
+++ b/external/duckdb/src/execution/operator/helper/physical_result_collector.cpp
@@ -45,7 +45,7 @@ PhysicalOperator &PhysicalResultCollector::GetResultCollector(ClientContext &con
 		if (data.output_type == QueryResultOutputType::ALLOW_STREAMING) {
 			return physical_plan.Make<PhysicalBufferedCollector>(data, false);
 		}
-		return physical_plan.Make<PhysicalMaterializedCollector>(data, true);
+		return physical_plan.Make<PhysicalMaterializedCollector>(data, false);
 	}
 
 	// Order-preserving plan, and we can use the batch index: use a batch collector.
--- a/external/duckdb/test/execution/distributed/test_executor_initialize.cpp
+++ b/external/duckdb/test/execution/distributed/test_executor_initialize.cpp
@@ -6,12 +6,26 @@
 #include "duckdb/main/connection.hpp"
 #include "duckdb/execution/executor.hpp"
 #include "duckdb/execution/physical_plan.hpp"
+#include "duckdb/execution/physical_plan_generator.hpp"
 #include "duckdb/execution/operator/helper/physical_result_collector.hpp"
 #include "duckdb/execution/operator/scan/physical_dummy_scan.hpp"
 #include "duckdb/main/prepared_statement_data.hpp"
 #include "duckdb/parallel/task_scheduler.hpp"
 
 using namespace duckdb;
+
+namespace {
+
+class ParallelUnpartitionedDummyScan : public PhysicalDummyScan {
+public:
+	using PhysicalDummyScan::PhysicalDummyScan;
+
+	bool ParallelSource() const override {
+		return true;
+	}
+};
+
+} // namespace
 
 TEST_CASE("Executor: Initialize with dummy plan builds pipelines", "[distributed][executor]") {
 	DuckDB db(nullptr);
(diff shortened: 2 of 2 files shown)


kaka11chen:
@codex review

hubgeter (approved):
LGTM

PR #327, opened 2026-07-31
[fix](udf) Allow registered table UDFs to change cardinality

### LINKED ISSUE #326: [P2][Bug] create_table_function rejects valid row-changing table UDF output
## Affected area

DuckDB or SQL execution

## What happened?

`DuckDBPyConnection.create_table_function()` rejects a registered table UDF when the function validly returns a different number of rows from its input batch.

With five input rows and a filter that returns three rows, execution fails with:

```text
_duckdb.InvalidInputException: Invalid Input Error: streaming UDF async error: streaming UDF lazy output count (3) does not match input rows (5) for submit_id 1
```

Expected result:

```text
[(0,), (1,), (2,)]
```

This prevents filter/flat-map-style registered table UDFs from using `create_table_function()`.

## Reproduction

```python
import duckdb
import pyarrow as pa


def keep_first_three(table):
    values = table.column(0).to_pylist()
    return pa.table({"x": [value for value in values if value < 3]})


con = duckdb.connect()
con.create_table_function(
    "filter_rows",
    keep_first_three,
    schema={"x": duckdb.sqltypes.BIGINT},
)

print(
    con.execute(
        "SELECT * FROM filter_rows((SELECT i FROM range(5) t(i)))"
    ).fetchall()
)
```

Run with:

```bash
VANE_RUNNER=local-fast python reproduce.py
```

Observed result:

```text
_duckdb.InvalidInputException: Invalid Input Error: streaming UDF async error: streaming UDF lazy output count (3) does not match input rows (5) for submit_id 1
```

## Environment

- OS: Linux 6.11, x86_64
- Python: 3.12.13
- Vane: `upstream/main` at `fe299e130f42c79efe25d4bceef7a545e1feeeed`
- Package: `0.1.0a1`, non-editable Release build
- DuckDB engine: `v1.6.0-dev2`
- `VANE_RUNNER`: `local-fast`

## Additional context

- [`RegisterTableUDF` builds the payload through `BuildPythonUDFPayload`](https://github.com/AstroVela/vane/blob/fe299e130f42c79efe25d4bceef7a545e1feeeed/src/duckdb_py/pyconnection.cpp#L1185-L1206). The resulting registration payload has no `row_preserving` field.
- [`BuildPythonUDFPayload` emits a `map_batches` call mode and returns the payload without a `row_preserving` field](https://github.com/AstroVela/vane/blob/fe299e130f42c79efe25d4bceef7a545e1feeeed/src/duckdb_py/python_udf_utils.cpp#L406-L454).
- The relation APIs record non-row-preserving semantics explicitly for both [`map_batches`](https://github.com/AstroVela/vane/blob/fe299e130f42c79efe25d4bceef7a545e1feeeed/src/duckdb_py/pyrelation.cpp#L2214-L2238) and [`flat_map`](https://github.com/AstroVela/vane/blob/fe299e130f42c79efe25d4bceef7a545e1feeeed/src/duckdb_py/pyrelation.cpp#L2333-L2354).
- In streaming execution, a missing `row_preserving` field is treated as row-preserving, and completion then rejects mismatched output cardinality: [`UDFRequiresRowPreservingOutput`](https://github.com/AstroVela/vane/blob/fe299e130f42c79efe25d4bceef7a545e1feeeed/external/duckdb/src/execution/operator/projection/physical_udf_inout.cpp#L414-L427), [completion check](https://github.com/AstroVela/vane/blob/fe299e130f42c79efe25d4bceef7a545e1feeeed/external/duckdb/src/execution/operator/projection/physical_udf_inout.cpp#L2328-L2335).

kaka11chen:
## Summary

- declare result-only `map_batches` and `flat_map` payloads as non-row-preserving in the shared payload builder
- remove the duplicate relation-level declarations while preserving expression UDF overrides
- add a regression proving a registered table UDF can filter five input rows to three output rows

## Root cause

`create_table_function()` used `BuildPythonUDFPayload()` without adding the `row_preserving` field that relation `map_batches()` and `flat_map()` added separately. Streaming execution treats a missing field as row-preserving for compatibility, so it rejected otherwise valid row-changing output.

The shared builder now records the result-only cardinality contract explicitly. Expression batch UDF construction continues to override the field when `row_preserving=True`, and the execution-side fallback for legacy payloads remains unchanged.

## Validation

- negative control: the new regression failed before the fix with `lazy output count (3) does not match input rows (5)`
- non-editable incremental Release build with `uv pip install . --no-build-isolation`
- `python -m pytest tests/fast/test_map.py tests/fast/test_expression_udf_map_batches.py tests/fast/test_expression_udf_contracts.py -q` — 100 passed
- `python -m pytest tests/fast/test_flat_map.py tests/fast/test_udf.py::test_flat_map_basic -q` — 9 passed
- `python -m pytest tests/fast/test_expression_cls_batch.py -q` — 17 passed
- `scripts/run_release_tests.sh` — 460 passed and 1 artifact-only skip; 10 real-Ray tests passed

Closes #326

### DIFF
--- a/src/duckdb_py/pyrelation.cpp
+++ b/src/duckdb_py/pyrelation.cpp
@@ -2227,7 +2227,6 @@ unique_ptr<DuckDBPyRelation> DuckDBPyRelation::MapBatches(
 		input_name_values.emplace_back(Value(col.Name()));
 	}
 	new_children.emplace_back("input_names", Value::LIST(LogicalType::VARCHAR, std::move(input_name_values)));
-	new_children.emplace_back("row_preserving", Value::BOOLEAN(false));
 	if (uses_subprocess_backend) {
 		new_children.emplace_back("produce_ref_bundle_output", Value::BOOLEAN(true));
 		new_children.emplace_back("streaming_output_mode", Value("local_shm_ref_bundle"));
@@ -2343,7 +2342,6 @@ unique_ptr<DuckDBPyRelation> DuckDBPyRelation::FlatMap(
 			input_name_values.emplace_back(Value(col.Name()));
 		}
 		new_children.emplace_back("input_names", Value::LIST(LogicalType::VARCHAR, std::move(input_name_values)));
-		new_children.emplace_back("row_preserving", Value::BOOLEAN(false));
 		if (uses_subprocess_backend) {
 			new_children.emplace_back("produce_ref_bundle_output", Value::BOOLEAN(true));
 			new_children.emplace_back("streaming_output_mode", Value("local_shm_ref_bundle"));
--- a/src/duckdb_py/python_udf_utils.cpp
+++ b/src/duckdb_py/python_udf_utils.cpp
@@ -408,6 +408,7 @@ Value BuildPythonUDFPayload(
 	auto udf_display_name = PythonCallableDisplayName(udf);
 	children.emplace_back("udf_name", Value(udf_display_name));
 	children.emplace_back("call_mode", Value(flat_map ? "flat_map" : "map_batches"));
+	children.emplace_back("row_preserving", Value::BOOLEAN(false));
 	children.emplace_back("execution_backend", Value(execution_backend));
 	children.emplace_back("function_pickle", Value::BLOB_RAW(pickled_str));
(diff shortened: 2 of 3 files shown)


kaka11chen:
@codex review

QuakeWang (approved):
LGTM

hubgeter (approved):
LGTM

PR #328, opened 2026-07-31
[fix](vllm) Use a single atomic inflight counter

### LINKED ISSUE #324: [P2][Concurrency] vLLM inflight accounting can form an inconsistent snapshot
## Affected area

Native vLLM execution and prompt backpressure accounting.

## What happened?

On the latest `main` (`fe299e130f42c79efe25d4bceef7a545e1feeeed`), `VLLMGlobalOperatorState::InflightPrompts()` reads two independently updated atomic counters in this order:

```cpp
const auto submitted = submitted_prompts.load();
const auto completed = completed_prompts.load();
if (completed > submitted) {
    throw InternalException("vllm completed prompt count exceeded submitted prompt count");
}
```

Both loads use the default sequentially consistent ordering, but that does not make the pair a single consistent snapshot. `PhysicalVLLM` is a parallel operator, and `InflightPrompts()` does not hold `executor_call_mutex`, so submissions and result drains from other pipeline threads can occur between the two loads.

A valid interleaving is:

1. `submitted_prompts == completed_prompts == N`.
2. Thread A reads `submitted_prompts` as `N`.
3. Thread B submits new work and increments `submitted_prompts` to `N + 1`.
4. A pipeline thread drains a completed result and increments `completed_prompts` to `N + 1`.
5. Thread A reads `completed_prompts` as `N + 1`.
6. Thread A throws `vllm completed prompt count exceeded submitted prompt count`, although the actual counters satisfy the invariant.

The submission and completion updates are individually serialized through `executor_call_mutex`, but the reader is outside that critical section, so that mutex does not prevent this interleaving.

`CanSubmitMore()` calls `InflightPrompts()` before checking whether `inflight_limit` is zero. Consequently, the inconsistent-snapshot exception is not inherently limited to configurations with a positive backpressure limit.

## Expected behavior

Concurrent, valid prompt submission and completion must not produce an internal counter-invariant exception.

## Actual behavior

A reader can combine an older submitted count with a newer completed count and report `completed > submitted`, aborting the query with an `InternalException`.

## Scope

The window requires concurrent `PhysicalVLLM` pipeline activity where one thread checks inflight accounting while other threads submit work and drain results. It is a concurrency-dependent query failure; no row corruption or silent wrong-result path has been identified.

## Environment

- Repository: `AstroVela/vane`
- Revision: `fe299e130f42c79efe25d4bceef7a545e1feeeed`
- Relevant file: `external/duckdb/src/execution/operator/projection/physical_vllm.cpp`
- Relevant methods: `VLLMGlobalOperatorState::InflightPrompts`, `CanSubmitMore`, `SubmitPrompts`, and `TakeReadyResultOnce`

kaka11chen:
## Summary

- replace the independently sampled submitted/completed totals with one authoritative atomic inflight count
- update the count through checked CAS transitions so valid concurrent work cannot form a mixed snapshot, while real over-completion and overflow still fail without corrupting state
- bypass inflight accounting entirely when backpressure is disabled
- add deterministic native tests for normal transitions, invalid transitions, competing completions, and concurrent readers/writers

## Why

`PhysicalVLLM` is parallel, but its old inflight calculation loaded two separately updated atomics. A submission and completion between those loads could make a reader combine an older submitted total with a newer completed total and raise an internal invariant error even though the live state was valid.

A single inflight value gives every reader one linearizable snapshot and removes that class of failure rather than relying on a particular two-load ordering.

## Validation

- Release incremental package build and install
- `python -m pytest tests/fast/test_vllm_native_regressions.py` (26 passed)
- vLLM fast regressions (82 passed, 1 optional vLLM dependency skip)
- `scripts/run_native_tests.sh "[vllm]" -s` (4 cases, 15 assertions)
- focused concurrency test repeated 50 times normally and 20 times on one CPU
- `scripts/run_native_tests.sh "[distributed]"` (150 cases, 4390 assertions)
- `scripts/run_release_tests.sh` (460 passed/1 skipped, then 10 passed)
- DuckDB formatter check and diff checks

Closes #324

### DIFF
--- a/external/duckdb/src/execution/operator/projection/physical_vllm.cpp
+++ b/external/duckdb/src/execution/operator/projection/physical_vllm.cpp
@@ -4,9 +4,10 @@
 #include "duckdb/execution/operator/projection/physical_vllm.hpp"
 
 #include "duckdb/common/exception.hpp"
+#include "duckdb/common/serializer/serializer.hpp"
 #include "duckdb/execution/expression_executor.hpp"
 #include "duckdb/execution/vllm_executor.hpp"
-#include "duckdb/common/serializer/serializer.hpp"
+#include "duckdb/execution/vllm_inflight_counter.hpp"
 
 #include <algorithm>
 #include <atomic>
@@ -193,11 +194,9 @@ struct VLLMGlobalOperatorState : public GlobalOperatorState {
 	std::mutex executor_call_mutex;
 	shared_ptr<VLLMExecutor> executor;
 
-	//! Global backpressure tracking (atomic for lock-free access).
-	//! Signed to avoid underflow when concurrent executors share an actor pool
-	//! and concurrent executors receive results belonging to another executor.
-	std::atomic<int64_t> submitted_prompts {0};
-	std::atomic<int64_t> completed_prompts {0};
+	//! Global backpressure tracking. A single atomic counter prevents readers
+	//! from combining submitted and completed values from different moments.
+	VLLMInflightCounter inflight_prompts;
 
 	//! Thread coordination for FinalExecute.
 	std::atomic<idx_t> active_threads {1};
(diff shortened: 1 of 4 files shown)


kaka11chen:
@codex review

hubgeter (approved):
LGTM

PR #331, opened 2026-07-31
[fix](relation) Remove unreachable repartition validation

### LINKED ISSUE #330: [P3][Maintenance] Relation::Repartition contains an unreachable validation branch
## Affected area

DuckDB or SQL execution

## What happened?

On main at 2d65f776d010, Relation::Repartition in external/duckdb/src/main/relation.cpp contains this condition:

```cpp
if (num_partitions > 0 && false) {
    throw InvalidInputException("num_partitions must be greater than zero");
}
```

The literal false makes the condition permanently false, so the exception branch is unreachable for every num_partitions value.

This is dead code rather than a current public API failure. The repartition configuration contract uses 0 to mean automatic partition selection. The Python binding also distinguishes an omitted partition count from an explicit 0 and rejects an explicitly non-positive value before entering the Relation API.

## Reproduction

Inspect Relation::Repartition on current main. No runtime input can enter the branch because its condition contains a literal false operand.

## Impact

The unreachable branch is misleading during maintenance because it appears to enforce a constraint while having no runtime effect, and its condition does not reflect the surrounding partition-count contract.

Priority: P3.

kaka11chen:
## Summary

- remove the permanently disabled validation branch from Relation::Repartition
- preserve the internal zero-as-auto partition contract
- leave Python argument validation unchanged

## Validation

- scripts/format duckdb --changed
- tests not run, per requested scope
- local build not run, per requested scope

Closes #330

### DIFF
--- a/external/duckdb/src/main/relation.cpp
+++ b/external/duckdb/src/main/relation.cpp
@@ -123,9 +123,6 @@ shared_ptr<Relation> Relation::Limit(int64_t limit, int64_t offset) {
 }
 
 shared_ptr<Relation> Relation::Repartition(idx_t num_partitions, vector<unique_ptr<ParsedExpression>> partition_by) {
-	if (num_partitions > 0 && false) {
-		throw InvalidInputException("num_partitions must be greater than zero");
-	}
 	return make_shared_ptr<RepartitionRelation>(shared_from_this(), num_partitions, std::move(partition_by));
 }
 


kaka11chen:
@codex review

QuakeWang (approved):
+1

hubgeter (approved):
LGTM

PR #370, opened 2026-08-02
[fix](relation) Match explode columns case-insensitively

### LINKED ISSUE #369: [P2][Relation] explode() matches column names case-sensitively
## Affected area

Relation API; `Relation.explode()` / `UnnestRelation` column binding.

## What happened?

Verified against upstream `main` at `310fc0910877b21a0b1b7eb22abefcc5624eaf29`.

For a relation with one uniquely named column `a`, SQL identifier binding and expression-based Relation APIs accept `A`, but `relation.explode("A")` rejects it with `BinderException`. `UnnestRelation` searches the child names using exact string equality, and its constructor eagerly binds the relation, so the mismatch is rejected during `explode()` construction.

This is a consistent failure in both the direct and serialized Relation paths; it is not the previously reported path-dependent behavior.

## Reproduction

```python
import os

import duckdb

con = duckdb.connect()
base = con.sql("SELECT [1, 2]::INTEGER[] AS a")

print(con.sql('SELECT A FROM (SELECT [1, 2]::INTEGER[] AS a)').fetchall())
print(base.project("A").fetchall())
print(base.explode("A").fetchall())
```

Observed:

```text
[([1, 2],)]
[([1, 2],)]
BinderException: Binder Error: Column "A" not found in child relation for unnest/explode
```

Using the exact spelling succeeds:

```python
print(base.explode("a").fetchall())
```

```text
[(1,), (2,)]
```

The connection remains usable after the binding error.

## Code evidence

- [`FindUnnestColumnIndex` compares `names[i] == column_name`](https://github.com/AstroVela/vane/blob/310fc0910877b21a0b1b7eb22abefcc5624eaf29/external/duckdb/src/main/relation/unnest_relation.cpp#L20-L27).
- [`UnnestRelation` eagerly calls `TryBindRelation` in its constructor](https://github.com/AstroVela/vane/blob/310fc0910877b21a0b1b7eb22abefcc5624eaf29/external/duckdb/src/main/relation/unnest_relation.cpp#L29-L34).
- [Direct binding uses the same exact-match helper](https://github.com/AstroVela/vane/blob/310fc0910877b21a0b1b7eb22abefcc5624eaf29/external/duckdb/src/main/relation/unnest_relation.cpp#L67-L77).

## Impact

Callers that obtain a column spelling independently of the relation metadata can bind the same name successfully in SQL and other expression-based Relation operations but fail when composing `explode()`. The failure is loud and occurs at relation construction; it does not produce wrong rows or invalidate the connection.

## Scope and related work

PR #298 intentionally preserved exact target selection when a relation contains duplicate names that differ only by case. This issue is limited to a unique child column, where no case-only duplicate needs disambiguation. #298 fixed explode column order and ARRAY execution, but it did not record this compatibility discrepancy as a dedicated issue.

## Validation status

- Static trace against upstream `main` commit `310fc0910877b21a0b1b7eb22abefcc5624eaf29`.
- Dynamic reproduction with the local Release native build containing the same eager bind and exact-equality lookup.
- No dedicated open or closed issue was found by searches for `explode case sensitive column` and `UnnestRelation case insensitive`.

## Environment

```text
OS: Linux
Architecture: x86_64
Vane main commit inspected: 310fc0910877b21a0b1b7eb22abefcc5624eaf29
Runner: local
Ray: not required
```

kaka11chen:
## Summary

- resolve `Relation.explode()` targets with exact-name priority followed by a unique ASCII case-insensitive match, matching DuckDB identifier semantics without changing the existing exact selection of case-only duplicate names
- reject non-exact lookups that match multiple child columns as ambiguous instead of silently choosing one
- preserve the resolved child column's original spelling in direct-bound output so direct and serialized Relation schemas remain aligned
- add LIST/ARRAY regressions for case-insensitive lookup, original-name preservation, direct-versus-serialized parity, and ambiguous case-only duplicates

## Related issue

Closes #369

## Documentation impact

      `AstroVela/vane-website`.

This restores the documented DuckDB identifier behavior for an existing Relation API; the public API is unchanged.

## Validation

- Release incremental native install with `SKBUILD_BUILD_DIR="$PWD/build/python-release" SKBUILD_CMAKE_BUILD_TYPE=Release uv pip install . --no-build-isolation`
- `python -m pytest tests/fast/relational_api/test_explode.py -q`: 16 passed
- `python -m pytest tests/fast/relational_api -q`: 409 passed, 1 skipped
- `scripts/run_release_tests.sh`: 468 passed, 1 skipped in the non-Ray shard; 10 passed in the real-Ray shard
- `VANE_NATIVE_BUILD_JOBS=8 scripts/run_native_tests.sh "[relation_api]"`: 35 test cases and 651 assertions passed
- manual exact/case-variant/ambiguous/missing-name matrix, including serialized rebinding and connection reuse
- pre-commit hooks for both changed files
- `git diff --check`
- two consecutive clean review passes after addressing the only review finding (Python import ordering)

## Checklist

      licenses and are recorded where required. No such additions are included.
      and untrusted input have been considered. This changes local identifier resolution only.

### DIFF
--- a/external/duckdb/src/main/relation/unnest_relation.cpp
+++ b/external/duckdb/src/main/relation/unnest_relation.cpp
@@ -13,6 +13,7 @@
 #include "duckdb/planner/expression/bound_unnest_expression.hpp"
 #include "duckdb/planner/expression/bound_columnref_expression.hpp"
 #include "duckdb/planner/expression/bound_cast_expression.hpp"
+#include "duckdb/common/string_util.hpp"
 #include "duckdb/common/types.hpp"
 
 namespace duckdb {
@@ -22,6 +23,21 @@ static idx_t FindUnnestColumnIndex(const vector<string> &names, const string &co
 		if (names[i] == column_name) {
 			return i;
 		}
+	}
+
+	idx_t case_insensitive_match = DConstants::INVALID_INDEX;
+	for (idx_t i = 0; i < names.size(); i++) {
+		if (!StringUtil::CIEquals(names[i], column_name)) {
+			continue;
+		}
+		if (case_insensitive_match != DConstants::INVALID_INDEX) {
+			throw BinderException("Ambiguous reference to column name \"%s\" in child relation for unnest/explode",
+			                      column_name);
+		}
+		case_insensitive_match = i;
+	}
+	if (case_insensitive_match != DConstants::INVALID_INDEX) {
+		return case_insensitive_match;
 	}
 	throw BinderException("Column \"%s\" not found in child relation for unnest/explode", column_name);
 }
@@ -75,6 +91,7 @@ BoundStatement UnnestRelation::Bind(Binder &binder) {
 
 	// 2. Find the column to unnest
 	auto unnest_col_idx = FindUnnestColumnIndex(child_bound.names, column_name);
+	const auto &unnest_column_name = child_bound.names[unnest_col_idx];
 
 	// 3. Determine element type from the list/array column
 	auto &list_type = child_bound.types[unnest_col_idx];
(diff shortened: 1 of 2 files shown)


kaka11chen:
@codex review

QuakeWang (approved):
+1

PR #371, opened 2026-08-02
[fix](python) Reject empty shared_ptr try_cast results

### LINKED ISSUE #159: [Bug] None bypasses py::try_cast type guards across the Python bindings (13 call sites raise InternalException)
## What happened?

While fixing #37 (merged as #135) I audited every `py::try_cast<shared_ptr<...>>` call site in `src/duckdb_py`. That crash is one instance of a systemic pattern: **`py::try_cast` returns `true` with a null holder when the input is Python `None`**, so every `if (!try_cast) throw InvalidInputException(...)` guard silently passes `None` through and crashes on the subsequent dereference with:

```text
INTERNAL Error: Attempted to dereference shared_ptr that is NULL!
```

The irony is that every affected site already has a correct, typed error path written for bad input — passing `"abc"` produces a clear `InvalidInputException` — but `None` is the one bad input that walks straight through the guard and turns into an internal error.

Root cause: `py::try_cast` (`src/duckdb_py/include/duckdb_python/pybind11/pybind_wrapper.hpp:81`) wraps `cast<T>` in try/catch. For holder types (`PYBIND11_DECLARE_HOLDER_TYPE(T, duckdb::shared_ptr<T>)`, same file), pybind11 converts `None` to a **null holder without raising `cast_error`** — the function reports success and hands the caller a null `shared_ptr`. "Cast succeeded" and "result is usable" diverge exactly for `None`.

Relation to existing reports: #122 item 6.1 covers the `PicklePythonUDFCallable` instance (fixed in #135) and the non-atomic registration concern; this issue generalizes the remaining class of call sites. The registration-atomicity half of 6.1 is out of scope here.

## Reproduction

Each line is an independent trigger (one per affected category):

```python
import duckdb
from duckdb import FunctionExpression

con = duckdb.connect()

# 1. UDF parameter annotation (return annotations are guarded; parameters are not)
def f(x: None):
    return 1
con.create_function("f", f, [duckdb.typing.BIGINT], duckdb.typing.BIGINT)

# 2. Type creation
duckdb.struct_type([None])
duckdb.struct_type({"a": None})

# 3. Expression APIs
FunctionExpression("concat", None)
con.sql("select 1 as a").sort(None)
con.sql("select 1 as a").aggregate([None])
```

Expected: `InvalidInputException` with the message each site already defines (e.g. "Please provide arguments of type Expression!").
Observed: `INTERNAL Error: Attempted to dereference shared_ptr that is NULL!`

## Audit (17 sites, verified at current `main`)

Safe today (4) — each for a different reason:

| Site | Why safe |
|---|---|
| `pyconnection.cpp:113` (`PicklePythonUDFCallable`) | fixed in #135 with a local `&& pytype` guard |
| `python_udf.cpp:504` (return annotation) | guarded by `!py::none().is(return_annotation)` |
| `pyrelation.cpp:1885` (`Update` where-clause) | guarded by `!py::none().is(where)` |
| `typing/pytype.cpp:321` (`FromObject` TYPE case) | `GetTypeObjectType(None)` dispatches to INVALID and throws first |

Vulnerable (13) — each passes its guard on `None`, then dereferences the null holder:

| Site | Trigger |
|---|---|
| `python_udf.cpp:514` (parameter annotations) | `def f(x: None)` + `create_function` |
| `pyconnection/type_creation.cpp:57` (list) | `duckdb.struct_type([None])` |
| `pyconnection/type_creation.cpp:71` (dict) | `duckdb.struct_type({"a": None})` |
| `native/python_conversion.cpp:585` (Value fallback) | a `duckdb.Value` whose `type` is `None` (the Python-side `Value.__init__` stores it unvalidated) |
| `pyexpression.cpp:42,240,272` | `FunctionExpression("f", None)` and expression composition |
| `pyconnection.cpp:2450` (expression list) | expression APIs with a `None` element |
| `pyrelation.cpp:130,285,315,334,1912` | `rel.sort(None)`, `rel.aggregate([None])`, partition/order args, `rel.update(set={"col": None})` |

Note the inconsistency inside a single file: `python_udf.cpp` guards the return annotation against `None` but not parameter annotations — which suggests individual guarding doesn't scale and the helper itself should be safe.

Verification method: the None-to-(true, null holder) mechanism is empirically confirmed by the #37 crash and by reading the `try_cast`/holder declarations; each site above was classified by reading its guard and the dereference that follows. The acceptance tests below double as executable repros.

## Acceptance criteria

- One regression test per category: parameter annotation `None`, `struct_type([None])`, `struct_type({"a": None})`, and one expression API (`rel.sort(None)`), each asserting `InvalidInputException` (not `InternalException`).
- Existing suites unchanged; run `scripts/run_release_tests.sh`.

I'm happy to take this — the audit is done and the change is small; the work is mostly the regression tests.

kaka11chen:
## Summary

- Treat a pybind11 try_cast as unsuccessful when it produces an empty DuckDB shared_ptr holder, so existing invalid-input branches run before any dereference.
- Keep the generic result path unchanged, preserving false and zero-valued enums as well as None converting to a non-empty SQL NULL expression where implicit expression conversion is supported.
- Add regression coverage for statement, Value type, struct type, CSV dtype, UDF annotation, and expression conversion paths.
- Implement the result check with C++11-compatible function-template overloads; no C++20-only language feature is used.

## Related issue

close: #159

## Documentation impact

## Validation

- Native incremental Release install: `uv pip install . --no-build-isolation`
- C++11 wrapper syntax check: `/usr/bin/c++ -std=c++11 -fsyntax-only ...`
- New regression tests: 8 passed
- Affected Python tests: 249 passed, 1 skipped
- `scripts/run_release_tests.sh`: 468 passed, 1 skipped, 14 deselected; real-Ray shard 10 passed, 473 deselected
- `scripts/run_fast_tests.sh`: non-Ray shard 6216 passed, 28 skipped, 8 xfailed, 1 xpassed; shared-Ray shard 98 passed, 6 skipped; isolated Ray-cluster shards 7 passed; optional vLLM shard skipped because vLLM is unavailable

## Checklist

### DIFF
--- a/src/duckdb_py/include/duckdb_python/pybind11/pybind_wrapper.hpp
+++ b/src/duckdb_py/include/duckdb_python/pybind11/pybind_wrapper.hpp
@@ -78,13 +78,25 @@ inline bool isinstance(handle obj, handle type) {
 }
 
 template <class T>
+bool is_valid_try_cast_result(const T &) {
+	return true;
+}
+
+// Pybind11 accepts None for holder types and returns an empty holder. Callers use a successful
+// try_cast as permission to dereference the result, so an empty shared_ptr is not a valid result.
+template <class T, bool SAFE>
+bool is_valid_try_cast_result(const shared_ptr<T, SAFE> &result) {
+	return result != nullptr;
+}
+
+template <class T>
 bool try_cast(const handle &object, T &result) {
 	try {
 		result = cast<T>(object);
 	} catch (pybind11::cast_error &) {
 		return false;
 	}
-	return true;
+	return is_valid_try_cast_result(result);
 }
 
 } // namespace py
--- a/tests/fast/test_pybind_try_cast.py
+++ b/tests/fast/test_pybind_try_cast.py
@@ -1,0 +1,56 @@
+# SPDX-FileCopyrightText: 2018-2025 Stichting DuckDB Foundation
+# SPDX-FileCopyrightText: 2026 Vane contributors
+# SPDX-License-Identifier: MIT AND Apache-2.0
+#
+# Modified by Vane contributors.
+
+from pathlib import Path
+
+import pytest
+
+import duckdb
+from duckdb import FunctionExpression
+from duckdb.sqltypes import BIGINT
+
+
+@pytest.mark.parametrize("fields", [[None], {"a": None}])
+def test_struct_type_rejects_null_type_holder(fields):
+    with pytest.raises(duckdb.InvalidInputException, match="object has to be a list of DuckDBPyType"):
+        duckdb.struct_type(fields)
+
+
+def test_execute_rejects_null_statement_holder(duckdb_cursor):
+    with pytest.raises(
+        duckdb.InvalidInputException,
+        match="Please provide either a DuckDBPyStatement or a string representing the query",
+    ):
+        duckdb_cursor.execute(None)
+
+
+def test_value_rejects_null_type_holder(duckdb_cursor):
+    value = duckdb.Value(1, None)
+    with pytest.raises(
+        duckdb.InvalidInputException,
+        match="The 'type' of a Value should be of type DuckDBPyType",
+    ):
+        duckdb_cursor.execute("select ?", [value])
+
+
+@pytest.mark.parametrize("dtype", [[None], {"category_id": None}])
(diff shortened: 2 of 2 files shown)


kaka11chen:
@codex review

QuakeWang (approved):
+1

PR #398, opened 2026-08-03
[fix](distributed) Preserve nested-loop join output schemas

kaka11chen:
## Summary

- keep the translated `HashJoinNode` output types as the schema contract when a pure non-equality join is rebuilt as a `PhysicalNestedLoopJoin`
- centralize join output-shape semantics for left-only, right-only, MARK, and two-sided joins, and reuse them in metadata repair and output-name translation
- fail fast when translated NLJ output types disagree with the attached child-derived schema instead of silently replacing the contract
- cover all join-type shapes plus the SEMI, ANTI, and MARK non-equality reconstruction paths

## Root cause

`HashJoinNode::BuildHashJoinTask` copied the correct translated output types into the replacement NLJ, then unconditionally overwrote them with all left and right child types. That shape is valid for ordinary two-sided joins, but corrupts SEMI/ANTI (left only) and MARK (left plus BOOLEAN) output schemas.

## Validation

- `scripts/format duckdb --changed`
- `scripts/run_native_tests.sh "[distributed][join]"` — 15 test cases, 153 assertions passed
- local distributed execution with physical `NESTED_LOOP_JOIN` plans for MARK, SEMI, and ANTI — expected rows and schemas passed
- `scripts/run_release_tests.sh` — 471 passed, 1 skipped; real-Ray shard 10 passed
- full native suite — 4602/4619 passed; all 17 failures were reproduced on the unmodified baseline or came from the same missing loadable demo-extension artifacts in that build configuration

### DIFF
--- a/external/duckdb/src/execution/distributed/pipeline_node/join/hash_join.cpp
+++ b/external/duckdb/src/execution/distributed/pipeline_node/join/hash_join.cpp
@@ -6,8 +6,10 @@
 #include <algorithm>
 #include <functional>
 
+#include "duckdb/common/enum_util.hpp"
 #include "duckdb/common/exception.hpp"
 #include "duckdb/execution/distributed/pipeline_node/join/hash_join_metadata.hpp"
+#include "duckdb/execution/distributed/pipeline_node/join/join_output_types.hpp"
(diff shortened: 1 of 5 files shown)


kaka11chen:
@codex review

hubgeter (approved):
LGTM

PR #414, opened 2026-08-04
[fix](udf) Rebuild failed stream result collector

### LINKED ISSUE #409: Ray UDF collector fatal error permanently poisons subsequent queries
## Summary

`AsyncResultCollector` stores fatal scheduler and event-loop failures in `_thread_error`. Once populated, every later `track_generator_ref()` or `drain_results()` call raises the stored failure.

The native UDF dispatcher owns one collector across query-slot lifetimes. `EnsureAsyncCollector()` only constructs the collector when its pointer is null, so a failed but still-owned collector remains attached to the process.

## Affected code

- `duckdb/execution/udf_stream_result_collector.py`: `_thread_error` assignment sites and `_raise_if_stopped_locked()`
- `src/duckdb_py/udf_executor.cpp`: process-owned `async_collector` and `EnsureAsyncCollector()`

## Failure sequence

1. A distributed Ray UDF query triggers a collector-level invariant or scheduler failure, such as an unknown readiness handle, duplicate generator waitable, event-loop failure, or escaped callback exception.
2. The collector stores the exception in `_thread_error`.
3. The active query fails when the dispatcher next registers or drains collector work.
4. Query-slot cleanup does not clear or replace the process-owned collector.
5. Every later distributed Ray UDF query in the same process reuses the failed collector and raises the same stored error.

## Impact

A single collector-internal failure makes distributed Ray UDF execution permanently unavailable in that process. Unrelated later queries cannot recover without restarting the process. Ordinary non-Ray SQL execution and collectors in other processes are outside this failure scope.

## Reproduction

Run two distributed Ray UDF queries sequentially in one process. Inject a fatal collector error into the first query, allow that query to fail and retire, then execute an otherwise healthy second query. The second query fails with the first collector's stored error, and only one collector instance is constructed.

Observed on `upstream/main` at `64a776c07959552edddaadb8e446cb20bd8c6f50`.

kaka11chen:
## Summary

- expose fatal scheduler state to the native UDF dispatcher
- pause submissions, fail slots owned by a poisoned collector, and shut it down before lazily creating a replacement
- rename the collector API and native identifiers to `UDFStreamResultCollector` without a compatibility alias or fallback
- cover a failed first Ray UDF query followed by a successful query in the same process

## Tests

- `python -m pytest tests/fast/test_udf_process.py`
- `python -m pytest tests/fast/test_udf_stream_result_collector.py`
- `python -m pytest tests/fast/test_udf_executor_lifecycle.py`
- `scripts/run_release_tests.sh`

Closes #409

### DIFF
--- a/duckdb/execution/ray_stream_adapter.py
+++ b/duckdb/execution/ray_stream_adapter.py
@@ -145,7 +145,7 @@ class TaskLeaseObjectRefGenerator:
             admission.release()
             raise
         # Stream-contract validation belongs to RayStreamAdapter, which is
-        # constructed by AsyncResultCollector under its durable cleanup-ticket
+        # constructed by UDFStreamResultCollector under its durable cleanup-ticket
         # owner.  Once remote work exists, this constructor must not discover a
         # post-submit error that has no scheduler available to retain an
         # asynchronous driver handoff.
--- a/duckdb/execution/udf_stream_result_collector.py
+++ b/duckdb/execution/udf_stream_result_collector.py
@@ -212,7 +212,7 @@ class _CleanupTicket:
             return () if self._error is None else (self._error,)
 
 
-class AsyncResultCollector:
+class UDFStreamResultCollector:
     """Capacity-aware scheduler for lease-owned Ray generator streams.
 
     ``_cv`` owns every published stream transition. Registration first starts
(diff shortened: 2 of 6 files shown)


kaka11chen:
@codex review

hubgeter (approved):
LGTM

PR #485, opened 2026-08-05
[feat](udf) Make batch UDFs Arrow-column decorators

### LINKED ISSUE #482: [UDF] func.batch expression mode rejects multi-column output
### Problem

`vane.func.batch` can return a multi-column Arrow Table when called with Python values. However, when the input contains `vane.col(...)`, it enters expression mode and rejects any output schema with more than one field.

As a result, a single batch computation cannot produce multiple related fields in an expression. Users must split the computation into multiple single-column calls, causing duplicate work. For non-deterministic functions, the fields may also come from different executions.

### Proposed outcome

Entering expression mode should not change the number of fields that `func.batch` can return. Multiple fields produced by one batch execution should remain part of the same result, while existing single-column calls remain compatible.

The current workarounds are to use multiple single-column `func.batch` calls or switch to relation-level `map_batches`.

The first repeats the computation, while the second no longer supports expression-style composition.

### Additional context

```python
import pyarrow as pa
import vane


def enrich_batch(table: pa.Table) -> pa.Table:
    values = table.column("value").to_pylist()
    return pa.table(
        {
            "doubled": [value * 2 for value in values],
            "label": [f"value_{value}" for value in values],
        }
    )


schema = {
    "doubled": "INTEGER",
    "label": "VARCHAR",
}

# Returns two columns successfully.
result = vane.func.batch(
    enrich_batch,
    inputs={"value": [1, 2, 3]},
    schema=schema,
)

assert result.to_pydict() == {
    "doubled": [2, 4, 6],
    "label": ["value_1", "value_2", "value_3"],
}

# Fails in expression mode.
vane.func.batch(
    enrich_batch,
    inputs={"value": vane.col("value")},
    schema=schema,
    row_preserving=True,
)
```

```text
InvalidInputException:
map_batches expression requires exactly one output column
```

kaka11chen:
## Summary

- Redesign `@vane.func.batch(...)` and `@vane.cls.batch(...)` as decorator-only, row-preserving Arrow-column UDF APIs. User callables receive one `pyarrow.Array | pyarrow.ChunkedArray` per input and must return one Arrow column with matching cardinality.
- Represent related multi-field outputs as a declared Arrow Struct. `unnest=True` expands the fields into projected columns, while call-scoped `expression_id` metadata ensures that expression and SQL expansion execute the underlying UDF only once per logical call.
- Remove the legacy `inputs`, `schema`, and `row_preserving` batch-decorator surface without a compatibility or Table fallback. Cardinality-changing and Table-based transformations remain on Relation `map_batches`.
- Update SQL registration, Python/native bindings, optimizer extraction, stubs, affected call sites, tests, and the Unreleased changelog for the new contract.

## Related issue

close: #482

## Documentation impact

      `AstroVela/vane-website`.

The Unreleased section of `CHANGELOG.md` documents the breaking API change, Arrow-column protocol, Struct output, `unnest=True`, and the division of responsibility with Relation `map_batches`.

## Validation

- `scripts/format workspace --from-ref upstream/main --check` — passed for Vane-owned files and the DuckDB subtree.
- `pre-commit run --from-ref upstream/main --to-ref HEAD` — passed, including ruff, clang-format, and mypy.
- Targeted decorator, strict Arrow protocol, Struct/unnest execution sharing, separate-call identity, SQL, class-batch, and stub-sync tests — 264 passed; the final focused review selection passed 7/7.
- `scripts/run_release_tests.sh` — 507 passed, 4 skipped, 15 deselected; the real-Ray shard passed 11/11.
- `scripts/run_native_tests.sh "[distributed]"` — 184 test cases and 7,972 assertions passed.
- Changed slow and AI test selection — 8 passed.
- `scripts/run_fast_tests.sh` — the non-Ray shard passed 6,279 tests and the isolated owner-Ray shard passed 7 with 1 skip. The local shared-Ray environment had 15 unrelated Relation failures; all affected UDF tests passed.
- GitHub Required CI was green on the previous head; checks were re-triggered for the final review-fix commit.
- Review-comment regression coverage for Struct call identity, independent-call separation, batch-builder IDs, and synchronous decorator contracts — 7 passed; two consecutive final review rounds were clean.

## Checklist


kaka11chen:
@codex review

kaka11chen:
@codex review

vane/_expression_udf.py
```diff
@@ -60,6 +61,13 @@ def _arrow_to_duckdb_type(dtype: Any, *, original: Any) -> Any:
     if pa.types.is_fixed_size_list(dtype):
         child_type = _arrow_to_duckdb_type(dtype.value_type, original=original)
         return duckdb.array_type(child_type, dtype.list_size)
+    if pa.types.is_struct(dtype):
```
QuakeWang, line 64: This enables Struct outputs for the shared dtype path used by `vane.func` and `vane.cls`, but their expression calls do not carry a call identity. Expanding one Struct with `unnest(...)` creates multiple UDF nodes, so fields from one logical result can come from different executions for non-deterministic or stateful UDFs.

kaka11chen:
@codex review

vane/_expression_udf.py
```diff
@@ -1413,23 +1923,15 @@ def attach_function(
...
-    Python semantics.
+    Decorated UDFs require ``parameters`` and infer SQL argument names from
+    their Python signatures. Raw scalar callables additionally require an
+    effective ``return_dtype``. Raw batch callables retain the low-level
+    ``pa.Table`` protocol and require ``input_names`` plus ``schema``.
+
+    Decorator return types, batch sizes, and GPU/actor settings cannot be
+    overridden at attach time. Row-oriented class expressions propagate SQL
+    NULL without invoking user code and accept only timezone-naive
+    ``TIMESTAMP`` output; eager calls retain ordinary Python semantics.
```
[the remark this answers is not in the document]
kaka11chen: Fixed. The `attach_function` docstring now distinguishes decorator-configured batch sizes from the supported attach-time `batch_size` override for an instantiated row-oriented `vane.cls`. Regression coverage confirms that exception.

vane/_expression_udf.py
kaka11chen: Fixed in 86719cc4. Each direct expression call now carries a unique call-scoped `expression_id`; registered SQL calls receive one during lowering. The projection rewriter reuses an extracted UDF binding only for clones with the same ID, so expanding one Struct executes the logical call once while separate calls remain distinct. Added regression tests for nondeterministic/stateful `vane.func` and `vane.cls` Struct expansion and for distinct-call identity; the affected 264-test selection passes.

QuakeWang (approved):
LGTM

Issue #243, opened 2026-07-28
[P1][Security] Keep plaintext credentials out of serialized plans


hubgeter:
## Summary

Provider credential redaction currently protects display surfaces, but not serialized plan bytes.

`Secret` hides values from `repr`, logs, exception text, and EXPLAIN output, while its pickle representation preserves the wrapped plaintext so a worker can reconstruct the provider. AI descriptors are captured by Python UDF callables, and those callables are stored in the plan as a `function_pickle` BLOB. Anyone who can read the logical or physical plan bytes can therefore recover inline credentials.

This is the same security boundary exposed by the native vLLM review in PR #137, but it is not vLLM-specific.

Related:

- PR review: https://github.com/AstroVela/vane/pull/137#discussion_r3631469867
- #105 fixed descriptor `repr` and error rendering, but did not remove plaintext from pickle bytes.

## Confirmed affected paths

The built-in Python AI providers accept inline credentials in their relation/expression APIs and place their descriptors inside serialized UDF payloads:

- OpenAI prompt and embedding
- Anthropic prompt
- Google prompt and embedding
- Transformers embedding and classification

A byte-level probe with unique sentinel values produced this result:

| Provider | EXPLAIN contains plaintext | Logical plan pickle contains plaintext | Physical plan pickle contains plaintext |
| --- | --- | --- | --- |
| OpenAI | No | Yes | Yes |
| Anthropic | No | Yes | Yes |
| Google | No | Yes | Yes |
| Transformers | No | Yes | Yes |

## Serialization path

The plaintext reaches serialized plans through the following path:

1. Provider descriptors call `wrap_sensitive_options(...)`.
2. `Secret.__reduce__()` serializes the wrapped plaintext.
3. `_PromptBatch`, `_EmbedTextBatch`, or `_ClassifyTextBatch` retains the descriptor.
4. `BuildExpressionMapBatchesUDFPayload()` pickles the complete callable into the `function_pickle` BLOB.
5. The BLOB is copied into logical and physical plans.

The current pickle tests only check that `repr(pickle.loads(...))` remains redacted. They do not check whether the original sentinel is present in the raw pickle bytes.

## Additional serialized credential paths

Connection session snapshots also capture credential-bearing `AWS_*`, `DUCKDB_*`, `S3FS_*`, and selected `VANE_*` environment values and preserve them through physical-plan serialization.

Arbitrary Python UDF closures can similarly capture secrets in `function_pickle`.

[...]

Issue #151, opened 2026-07-21
[P2][AI] Transformers provider: hidden_size is not the sentence-transformers output dimension; GPU detection runs on the driver


caomaocao:
## Problem

1. **Wrong dimension source** (`vane/ai/providers/transformers.py:94-110`): `get_dimensions` reads `AutoConfig.hidden_size`, which differs from the actual sentence-transformers output dimension for models with Dense/projection pooling (e.g. `distiluse-base-multilingual-cased-v1`: hidden 768, actual output 512) → wrong declared dims, wrong zero-fill width, wrong schema for consumers. CLIP-style configs have no top-level `hidden_size` → `AttributeError`.
2. **GPU allocation decided on the driver** (`transformers.py:112-123`): `torch.cuda.is_available()` runs on the driver to set worker `num_gpus`; heterogeneous driver/worker GPU setups mis-allocate. The classifier descriptor never requests a GPU at all and its pipeline defaults to CPU — asymmetric with the embedder.
3. Minor: UDF batch is 64 but `model.encode` is called without `batch_size` (ST default 32) — perf only (`transformers.py:162-167`).

## Suggested fix

Use `SentenceTransformer.get_sentence_embedding_dimension()` (or the model's `modules.json`/pooling config) instead of `hidden_size`; decide GPU placement from descriptor options rather than driver-side probing, and align embedder/classifier GPU semantics.

Reviewed commit: c43bd706.

caomaocao:
I'll take this one (filed from my provider-layer review; assignment welcome — no triage rights to self-assign).

kaka11chen:
Consolidating the overlapping resource-descriptor report from #101 here. In addition to the embedding-dimension fix already described, the canonical acceptance scope should retain: preserve explicit caller actor/GPU constraints; determine devices from worker/Ray allocation rather than driver visibility; validate impossible combinations before actor creation; and cover CPU, fractional-GPU, and no-GPU-driver/GPU-worker configurations. #101 is being closed as absorbed by this issue.

kaka11chen:
Scope update after merged #397: the replacement Transformers embedder no longer derives dimensions from `AutoConfig.hidden_size`; dimensions are trusted/explicit, and GPU resources are selected from explicit descriptor device options rather than a driver-side CUDA probe.

Keeping this open because classifier GPU/resource semantics are still asymmetric and `SentenceTransformer.encode` still does not receive the UDF batch size. The remaining issue/PR should be scoped to those pieces.

[...]

PR #487, opened 2026-08-05
[fix](ai) Remove legacy text classification

### LINKED ISSUE #106: [P1][Bug] AI retries treat deterministic errors like transient failures and zero-fill embeddings
## Problem

The shared AI retry layer does not consistently distinguish authentication, invalid request, rate limit, and transient server/network failures. SDK-level and Vane-level retries can multiply.

Some embedding error modes return correctly shaped all-zero vectors, which look valid and can silently contaminate similarity search.

kaka11chen:
## Summary

- remove the legacy classify_text function, Relation method, batch wrapper, Provider protocol, and Transformers classifier runtime
- remove the obsolete classifier error-policy and provider-resolution fallbacks instead of retaining compatibility shims
- make the shared sync and async retry core intrinsically retry only classified transient provider failures and reject invalid on_error values before execution
- update the changelog and retain negative contract tests for every removed entry point

## Why

PR #397 resolved the zero-fill and retry behavior for the replacement Prompt and Embed paths. The only scope left in #106 was the legacy _ClassifyTextBatch path, which still used the broad historical retry behavior. Removing that obsolete vertical slice and its fallbacks completes the contract.

Closes #106

## Related issues

- Partially addresses #100 and #148 by removing the non-logging log mode and invalid-policy fallback from the remaining shared Python retry surface.
- Partially addresses #151 by removing the classifier side of the Transformers GPU-allocation asymmetry.
- Related to #152 and #243 because the legacy classifier protocol and serialized classifier descriptor path are removed; their broader remaining scopes stay open.

## Validation

- scripts/format root --changed
- pre-commit on all 16 changed files
- 432 directly affected AI and release-contract tests passed
- scripts/run_release_tests.sh: 508 passed and 1 skipped in the non-Ray shard; 11 passed in the real-Ray shard
- full fast launcher: non-Ray and shared-Ray phases passed; owner-Ray phase passed on a complete phase rerun with 7 passed and 1 dependency skip after one fault-injection test flaked once and passed immediately in isolation
- two consecutive clean review rounds on the final unchanged diff

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,10 @@ All notable user-visible changes are documented here. Vane is currently in alpha
 - Positioned the current project as the Vane Data developer preview.
 - Prompt uses ordered `messages`, first-class call parameters, and the closed
   `PromptOptions` keyword surface. OpenAI Responses is the default endpoint.
+- Removed the legacy Python-only zero-shot classification API, Provider
+  protocol, and Transformers runtime. Prompt and Embed are now the complete AI
+  surface, and their shared retry layer retries only classified transient
+  failures.
 - Defined `DuckDBPyRelation.map` exclusively as a row-wise scalar UDF with a
   required `return_type`; batch transforms use `map_batches` with an explicit
   output `schema`. The inherited pandas DataFrame-style DuckDB `map` contract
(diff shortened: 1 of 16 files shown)


kaka11chen:
@codex review

QuakeWang (approved):
LGTM

PR #502, opened 2026-08-06
[fix](distributed) Stabilize Flight shutdown lifecycle test

### LINKED ISSUE #501: [Flaky][Distributed] Flight shutdown test can miss the active shutdown window
## Affected area

Distributed execution or Ray

## What happened?

The native distributed test named Exchange: process-local Flight shutdown is bounded and releases its service lock failed intermittently in the Python 3.12 CI matrix.

The test expects a service start issued during shutdown to fail. In the failing run, the endpoint lookup had already observed the unpublished port as zero, but the subsequent start completed successfully:

    REQUIRE( concurrent_start_result.is_err() )
    with expansion:
      false

This failed PR #500 Required CI even though that PR changes only AI Python files and does not modify the Flight implementation or its native tests.

## Reproduction

Run the native test:

    scripts/run_native_tests.sh "Exchange: process-local Flight shutdown is bounded and releases its service lock"

Observed failure:

- CI run: https://github.com/AstroVela/vane/actions/runs/31106479839/job/92632950411?pr=500
- Test case 142 of 184 failed at external/duckdb/test/execution/distributed/test_exchange.cpp:1141.
- The same source passed in the Python 3.12 native jobs for PR #498 and PR #499.
- The same test binary passed 30 consecutive local repetitions, so the failure is not deterministic.

## Environment

- OS: GitHub-hosted Ubuntu runner
- Architecture: x86_64
- Python: 3.12
- Vane commit: f5c1689f5def053b4c5a82a847ddf072d821cfcc
- Execution: native distributed and Flight tests enabled
- Ray: not involved in this native test
- VANE_RUNNER: not applicable

## Additional context

The test launches shutdown asynchronously, waits for a fixed 100 ms, and then launches the service start whose result is asserted. The observed successful start means the expected active-shutdown overlap was not present when that start acquired the service mutex.

Related history:

- #258 reported unbounded Flight server shutdown and is closed.
- #263 added the bounded shutdown behavior and this regression test.
- The current failure is different from #258: shutdown did not remain blocked indefinitely; the test instead observed that shutdown had completed before the asserted concurrent start.

kaka11chen:
## Summary

- treat both legal process-local Flight lifecycle linearizations as valid: a start that overlaps shutdown must fail with `shutting down`, while a start that reaches the mutex after shutdown completes may succeed
- keep the bounded-shutdown and lifecycle-lock responsiveness assertions, and explicitly stop any service restarted by the successful path
- rename the regression test to describe the lifecycle responsiveness contract instead of requiring a particular scheduling outcome

The failure was a test race rather than a Flight implementation regression. The test waited a fixed 100 ms before starting the concurrent operation, but the bounded shutdown could finish within that interval. Requiring the later start to fail therefore rejected a valid completed-shutdown/restart ordering. This change models the two valid orderings directly; it does not add a compatibility path or fallback.

## Related issue

close: #501

## Documentation impact

      `AstroVela/vane-website`.

This is a native regression-test correction with no public API or behavior change.

## Validation

- controlled local reproduction: with the stream reduced to one byte, the old assertion failed because the concurrent start succeeded after shutdown completed; the updated test passed the same forced ordering (temporary reproduction change reverted)
- `scripts/run_native_tests.sh "Exchange: process-local Flight shutdown is bounded and keeps its service lifecycle responsive"` — 1 case, 81 assertions passed
- updated target repeated 30 consecutive times — 30/30 passed
- `scripts/run_native_tests.sh "[distributed]"` — 184 cases, 7,970 assertions passed
- incremental native package build with `SKBUILD_BUILD_DIR=build/python-release`, Release configuration, and `uv pip install . --no-build-isolation`
- `scripts/run_release_tests.sh` — 510 non-Ray tests passed, 1 skipped; 11 real-Ray tests passed
- two consecutive clean code-review passes after rebasing onto current `upstream/main`

## Checklist

### DIFF
--- a/external/duckdb/test/execution/distributed/test_exchange.cpp
+++ b/external/duckdb/test/execution/distributed/test_exchange.cpp
@@ -1047,7 +1047,7 @@ TEST_CASE("Exchange: Flight service isolates published attempts and rejects rele
 	registry.RemoveAndCleanupByPrefix(prefix);
 }
 
-TEST_CASE("Exchange: process-local Flight shutdown is bounded and releases its service lock",
+TEST_CASE("Exchange: process-local Flight shutdown is bounded and keeps its service lifecycle responsive",
           "[distributed][exchange]") {
 	struct ServiceGuard {
 		~ServiceGuard() {
@@ -1127,7 +1127,7 @@ TEST_CASE("Exchange: process-local Flight shutdown is bounded and releases its s
 		reader->Cancel();
 	}
 	auto stop_result = stop_future.get();
-	const auto observed_port = port_future.get();
+	(void)port_future.get();
 	auto concurrent_start_result = start_future.get();
 	reader->Cancel();
 	if (concurrent_start_result.is_ok()) {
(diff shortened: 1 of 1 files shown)


kaka11chen:
@codex review

hubgeter (approved):
LGTM

PR #506, opened 2026-08-06
[fix](ray) Make result ownership teardown-safe

### LINKED ISSUE #436: [P2][Race] RayWorkerManagerBackend shares result-handle state without synchronization
## Affected area

`RayWorkerManagerBackend._handles_by_query`.

## What happens

The backend appends, pops, and removes lists in a shared `defaultdict` from task submission, input-exhaustion, query wait, and query drop paths without a lock. These entry points can run from scheduler, completion, and teardown threads.

## Source evidence

`duckdb/runners/fte/backends/ray/backend.py:200-280`:
- appends at lines 229-248;
- pop/extend at lines 259-275;
- removal at lines 278-280.

## Impact

Interleavings can lose a just-appended result handle, return it after query teardown, or mutate a list while another path is consuming it. Lost handles can also delay output release.

## Verification scope

Static shared-state audit of `upstream/main` at `64a776c07959552edddaadb8e446cb20bd8c6f50`.

### LINKED ISSUE #439: [P2][Ray UDF] Generation-flip compensation can release an output already handed downstream
## Affected area

Native UDF dispatcher output delivery during slot teardown.

## What happens

`DeliverOutputEvent` passes an event, including its output-lease release callback, into the downstream consumer. If the slot generation flips while `accept_event` runs, the post-call compensation path invokes a saved copy of that release callback. The consumer may already own the event and still need the referenced object.

## Source evidence

`src/duckdb_py/udf_executor.cpp:2393-2415`: `accept_event(std::move(event))` runs first; a later failed generation check calls `ReleaseOutputLeaseCallback(release_on_consumer_error)`.

## Impact

Teardown racing successful delivery can release the Ray object before the downstream reader consumes it, turning a cancellation window into missing-object errors or corrupted query output.

## Verification scope

Static ownership audit of `upstream/main` at `64a776c07959552edddaadb8e446cb20bd8c6f50`.

kaka11chen:
## Summary

- Replace Ray FTE result-handle lists with a synchronized query-owner lifecycle that fences registration, active operations, drop, outer replay/data-source cleanup, and shutdown as one teardown boundary.
- Preserve owner state after failed drop or shutdown, serialize retries and concurrent teardown, reject late work and stale query-ID reuse, and require the explicit lifecycle/resource-owner contracts without compatibility fallbacks.
- Replace copyable UDF output callbacks with a move-only RAII lease so ownership transfers exactly once and moved-from producers cannot release output already handed downstream.
- Add Python/C++ concurrency regressions for same-owner re-registration, active-submit teardown failures, nested execution owners, shutdown/drop races, retry state, strict contracts, and late result handles, plus native lease tests.

## Related issue

close: #436
close: #439

## Documentation impact

      `AstroVela/vane-website`.

This intentionally tightens the internal worker-backend contract: lifecycle registration, explicit resource ownership, drop, and shutdown are required; there is no compatibility fallback.

## Validation

- `SKBUILD_BUILD_DIR="$PWD/build/python-release" SKBUILD_CMAKE_BUILD_TYPE=Release uv pip install . --no-build-isolation`
- `build/native-cxx20/test/unittest "[output-lease]"` (4 test cases, 13 assertions)
- `python -m pytest tests/fast/test_fte_native_backend.py tests/fast/test_fte_ray_backend_adapter.py tests/fast/test_ray_cpp_bindings.py tests/fast/test_udf_process.py` (276 passed, 3 skipped, 4 deselected)
- `scripts/run_release_tests.sh` (507 passed, 4 skipped, 15 deselected; real-Ray shard: 11 passed, 515 deselected)
- `pre-commit run --from-ref 23ada9229312e7b8787d0e61a22061b5e7d5456c --to-ref HEAD`
- `scripts/format workspace --from-ref 23ada9229312e7b8787d0e61a22061b5e7d5456c --check`
- `git diff --check 23ada9229312e7b8787d0e61a22061b5e7d5456c..HEAD`

After the final implementation, two consecutive review passes found no further issues. The second pass independently checked API contracts, ownership/release paths, lock ordering, failure retries, and the merge with current `main`.

## Checklist


kaka11chen:
@codex review

src/duckdb_py/ray/distributed_plan_bindings.cpp
```diff
@@ -1059,6 +1075,61 @@ class PyBackendWorkerManager : public duckdb::distributed::WorkerManager {
...
+		}
+		{
+			lock_guard<mutex> guard(mutex_);
+			auto existing_owner = result_handle_owner_by_query_.find(query_id);
+			if (existing_owner != result_handle_owner_by_query_.end()) {
+				if (existing_owner->second != owner_query_id) {
+					throw std::runtime_error("FTE query result owner changed while active: query=" + query_id +
+					                         " existing=" + existing_owner->second + " requested=" + owner_query_id);
+				}
+				return;
```
hubgeter: **Blocking:** The same-owner early return bypasses the close, drop, and shutdown fence below it. If an old `query_id -> owner_query_id` mapping is still present while `drop_query_fragments` has marked the owner closed/dropping, a concurrent reuse of the same IDs returns success here without calling `backend.register_query_owner`. The old drop can then erase the mapping and backend generation after `run_plan` already considers the new generation registered, so its first operation is silently rejected or canceled.

Please check the closing states before this idempotent return, matching the Python backend behavior, and add a test proving that registration is rejected during an in-progress drop/cleanup and succeeds only after that drop finishes.

src/duckdb_py/ray/distributed_plan_bindings.cpp
```diff
@@ -1295,45 +1433,121 @@ class PyBackendWorkerManager : public duckdb::distributed::WorkerManager {
...
 				}
 			} catch (...) {
 				backend_drop_error = std::current_exception();
 			}
 		}
-		if (result_cleanup_error && backend_drop_error) {
-			auto exception_message = [](const std::exception_ptr &error) {
+		std::exception_ptr initial_cleanup_error;
+		std::exception_ptr final_cleanup_error;
+		if (!backend_drop_error) {
```
hubgeter, line 1523: **Blocking:** This skips `WaitForResultHandleOperations` whenever the backend drop raises, but `PyPhysicalPlanWrapperRunner::drop_query_fragments` catches that exception and still unconditionally runs `ReleaseDataSourceFactoriesForQuery` and `CleanupQueryPythonReplayState`. An already-active submit can therefore resume after the failed drop while the corresponding replay/data-source state has been removed. A transient teardown failure can turn into an execution failure, and the advertised cleanup retry is no longer teardown-safe.

Please retain the outer owner-scoped state until the inner lifecycle reaches successful quiescence, or until a successful retry/shutdown, and cover the sequence active operation + backend drop error + retry.

duckdb/runners/fte/backends/ray/backend.py
```diff
@@ -232,76 +245,433 @@ def worker_snapshots(self) -> Sequence[Mapping[str, Any]]:
...
+        try:
+            _required_method(self._coordinator, "wait_fte_query")(query_id, float(timeout_s))
+            raw_handles = []
+            pop_handles = getattr(self._coordinator, "pop_fte_result_handles", None)
+            if callable(pop_handles):
+                raw_handles = list(pop_handles(query_id) or [])
+            handles = self._accept_handles(active_owner, raw_handles)
+            if allowed is not None:
+                selected_handles = []
+                discarded_handles = []
```
hubgeter: The current Code Quality job fails mypy because these empty lists need explicit element types.

```suggestion
                selected_handles: list[RayTaskResultHandleAdapter] = []
                discarded_handles: list[RayTaskResultHandleAdapter] = []
```

hubgeter (changes requested):
Requesting changes for two lifecycle correctness issues:

1. C++ same-owner registration can bypass an active drop or shutdown fence.
2. A failed backend drop skips quiescing active operations while the outer replay and data-source state is still destroyed.

There is also a straightforward mypy failure in the current Code Quality check. The move-only UDFOutputLease direction and the broader owner-scoped lifecycle architecture look good; the blockers are the edge cases called out inline.

Review scope: static code analysis only, as requested. I did not run local builds or tests.

kaka11chen:
@codex review

src/duckdb_py/ray/distributed_plan_bindings.cpp
kaka11chen: Fixed in dc404a05e5. Owner replay/data-source cleanup now runs inside the C++ lifecycle fence only after backend teardown, active-operation quiescence, and result-handle cleanup succeed. Failed drop/shutdown keeps owner state for retry, and concurrent drop waits for shutdown. The active-submit drop-failure/retry, nested-owner cleanup, shutdown/drop race, and shutdown-retry tests cover these paths.

src/duckdb_py/ray/distributed_plan_bindings.cpp
kaka11chen: Fixed in dc404a05e5. Same-owner registration now checks shutdown, closed-owner, and active-drop fences before the idempotent return. test_cxx_backend_re_registering_dropping_owner_is_not_treated_as_idempotent_success covers rejection during drop and successful reuse only after teardown finishes.

duckdb/runners/fte/backends/ray/backend.py
kaka11chen: Fixed in dc404a05e5 with explicit RayTaskResultHandleAdapter list types. The changed-file mypy hook and the complete pre-commit diff both pass.

hubgeter (approved):

PR #508, opened 2026-08-07
[fix](ai) Reject empty nested objects and case-colliding schema fields

### LINKED ISSUE #466: [P2][AI] Nested empty object schemas compile to invalid DuckDB STRUCT types
## Affected area

Prompt structured-output schema compilation.

## What happens

At PR #397 HEAD 8f3f8af5, an object below the schema root may declare an empty properties mapping. Vane accepts the schema and derives STRUCT("meta" STRUCT()), but DuckDB rejects the zero-child nested STRUCT when the type is bound.

This lets a statically invalid result type pass schema compilation and fail later at the DuckDB boundary.

## Reproduction

~~~python
from vane.ai._schema import compile_return_format

spec = compile_return_format({
    "type": "object",
    "properties": {
        "meta": {"type": "object", "properties": {}},
    },
})
print(spec.duckdb_type)
~~~

Observed derived type:

~~~text
STRUCT("meta" STRUCT())
~~~

Binding that type raises a DuckDB ParserException near the inner closing parenthesis.

## Existing issue check

#381 and #383 are the parent phase/release issues and are scheduled to be closed by #397; neither separately tracks this failure. Repository issue searches for the schema shape, derived type, and compiler symbol found no equivalent open or closed issue.

## Source

Follow-up from #397: https://github.com/AstroVela/vane/pull/397#discussion_r3710038351

### LINKED ISSUE #472: [P2][AI] Case-insensitive schema field collisions fail during DuckDB binding
## Affected area

Prompt structured-output schema compilation.

## What happens

At PR #397 HEAD 8f3f8af5, Vane accepts distinct JSON Schema property spellings that DuckDB treats as the same STRUCT member name. A schema containing answer and ANSWER compiles successfully, then fails when DuckDB binds the generated type.

The error is therefore deferred past Vane schema validation.

## Reproduction

~~~python
from vane.ai._schema import compile_return_format

spec = compile_return_format({
    "type": "object",
    "properties": {
        "answer": {"type": "string"},
        "ANSWER": {"type": "integer"},
    },
})
print(spec.duckdb_type)
~~~

Observed derived type:

~~~text
STRUCT("answer" VARCHAR, "ANSWER" BIGINT)
~~~

DuckDB raises BinderException: Duplicate STRUCT type argument name "ANSWER".

## Existing issue check

#381 and #383 are the parent phase/release issues and are scheduled to be closed by #397. #369 and #448 concern Relation column resolution rather than structured-output schema compilation. Repository searches found no equivalent issue for schema property collisions.

## Source

Follow-up from #397: https://github.com/AstroVela/vane/pull/397#discussion_r3710038352

kaka11chen:
## Summary

The Prompt structured-output compiler accepted two schema shapes that derive
DuckDB types DuckDB itself rejects at bind time, deferring the failure past
Vane's schema validation:

- An object below the root may declare an empty `properties` mapping, deriving
  `STRUCT(...)` with zero children, which DuckDB cannot bind
  (`STRUCT("meta" STRUCT())` raises a ParserException).
- Distinct property spellings that differ only by case derive STRUCT members
  DuckDB treats as the same name (`answer`/`ANSWER` raises
  `BinderException: Duplicate STRUCT type argument name`).

Both are now rejected at compile time in `_compile_object`, the single
recursive compilation point, so nested objects, array items, and `$defs`
references are all covered. The root-only empty-object check in `compile()`
was folded into the same check.

Long-term hardening: a parametrized corpus test compiles every supported
schema shape and asserts the derived `duckdb_type` parses with
`duckdb.sqltype()` and binds in `CAST(NULL AS ...)`, so any future construct
that maps to an unparseable DuckDB type fails at compile/test time instead of
at the DuckDB boundary.

## Related issue

close: #466
close: #472

## Documentation impact

Both schemas previously compiled and then failed at the DuckDB boundary, so
the rejection only tightens validation of behavior that never worked.

## Validation

- `python -m pytest tests/ai/test_prompt_schema.py` — 35 passed, 2 skipped
  (pydantic not installed in the environment).
- Regression tests fail on the pre-fix `_schema.py` (verified by reverting the
  module change: 5 new tests fail) and pass with the fix.
- `ruff check` and `ruff format --check` clean on both changed files.
- `scripts/format root --changed` run.
- Full `tests/ai` run: 505 passed; the 14 failures are all pre-existing
  `ModuleNotFoundError: No module named 'openai'` in this environment,
  unrelated to this change.

## Checklist

### DIFF
--- a/tests/ai/test_prompt_schema.py
+++ b/tests/ai/test_prompt_schema.py
@@ -8,6 +8,7 @@ import json
 import pyarrow as pa
 import pytest
 
+import duckdb
 from vane.ai._schema import OutputValidationError, SchemaValidationError, compile_return_format
 
 PORTABLE_SCHEMA = {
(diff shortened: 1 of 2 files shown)


hubgeter (approved):
LGTM

PR #513, opened 2026-08-07
[fix](execution) Reject zero partitions in RepartitionNode

### LINKED ISSUE #427: [P2][Liveness] Repartition accepts zero partitions and creates a channel that can never send
## Affected area

Bounded channel construction and `RepartitionNode`.

## What happens

A bounded channel with capacity zero waits for `queue.size() < capacity`; that predicate can never become true while a receiver remains attached. `RepartitionNode` does not reject `num_partitions == 0` and uses the value directly as the result-channel capacity.

## Source evidence

- `external/duckdb/src/include/duckdb/execution/distributed/utils/channel.hpp:48-67` permits zero capacity and blocks `send`.
- `external/duckdb/src/execution/distributed/pipeline_node/shuffles/repartition.cpp:208-240` stores the unvalidated count.
- The channel is created with `num_partitions_` at lines 275-290.

## Impact

A malformed or propagated zero-partition plan can hang permanently on the first send. Later source-task calculations also divide work using the zero count, so the state is not a meaningful executable plan.

## Verification scope

Static validation and liveness audit of `upstream/main` at `64a776c07959552edddaadb8e446cb20bd8c6f50`.

kaka11chen:
## Summary

RepartitionNode previously accepted `num_partitions == 0`, which created a bounded channel with capacity 0. Since `ChannelState::send()` waits for `queue_.size() < capacity`, that predicate can never be true when capacity is zero — causing a permanent deadlock on the first send.

The fix adds an `InvalidInputException` guard in both the private constructor and the static factory `create()`, matching the existing validation pattern in `PhysicalRemoteExchangeSink` (`physical_remote_exchange_sink.cpp:130`).

## Related issue

close: #427

## Documentation impact

## Validation

- Static validation: the guard fires at plan construction time, before any channel or exchange resources are allocated.
- The factory check acts as a fast path (rejects before PipelineNodeContext / ClusteringSpec allocation); the constructor check is the structural invariant against future bypass.

## Checklist

### DIFF
--- a/external/duckdb/src/execution/distributed/pipeline_node/shuffles/repartition.cpp
+++ b/external/duckdb/src/execution/distributed/pipeline_node/shuffles/repartition.cpp
@@ -212,6 +212,9 @@ RepartitionNode::RepartitionNode(PipelineNodeConfig config, PipelineNodeContext 
                                  std::shared_ptr<ExchangeManager> exchange_mgr)
     : config_(std::move(config)), context_(std::move(context)), repartition_spec_(std::move(repartition_spec)),
       num_partitions_(num_partitions), child_(std::move(child)), exchange_mgr_(std::move(exchange_mgr)) {
+	if (num_partitions_ == 0) {
+		throw InvalidInputException("RepartitionNode requires at least one partition");
+	}
 }
 
 // Static factory method.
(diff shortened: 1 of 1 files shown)


QuakeWang (approved):
+1

PR #518, opened 2026-08-08
[fix](ai) Guard malformed Retry-After delays before retry sleeps

### LINKED ISSUE #469: [P2][AI] Malformed Retry-After delays fail inside retry sleeps
## Affected area

Prompt and Embed retry handling.

## What happens

At PR #397 HEAD 8f3f8af5, RetryAfterError accepts negative and non-finite delays. The synchronous retry path passes those values through min() and into time.sleep().

Observed outcomes:

~~~text
RetryAfterError(-1.0) -> ValueError: sleep length must be non-negative
RetryAfterError(NaN)  -> ValueError: Invalid value NaN (not a number)
~~~

Google's Retry-After header parsing uses float(), so parseable negative and NaN header values can reach the same path. The sleep failure replaces the provider retry/error-policy behavior.

## Existing issue check

#148 covers broader provider retry and on_error drift but does not track invalid Retry-After delay values or the resulting sleep exception. #382 and #383 are parent phase issues scheduled to be closed by #397. Searches for RetryAfterError with negative, NaN, and finite-value terms found no equivalent issue.

## Source

Follow-up from #397: https://github.com/AstroVela/vane/pull/397#discussion_r3711116179

caomaocao:
## Summary

A `RetryAfterError` carrying a negative, NaN, or infinite `retry_after` reached `time.sleep()`/`asyncio.sleep()` unchecked. The sync Embed retry path raised `ValueError` (`sleep length must be non-negative` / `Invalid value NaN (not a number)`) after the **first** attempt, bypassing the remaining retries and the `on_error="ignore"` policy and replacing the original 429/503 with a spurious sleep failure. The async Prompt path was worse than the issue describes: `asyncio.sleep(nan)` creates a timer that never fires on Python ≤ 3.12, hanging the retry forever (reproduced locally on 3.12.2; 3.13+ raises `ValueError` instead).

Both values are reachable from a misbehaving custom provider (`RetryAfterError` is public API) and from Google's `Retry-After` header, which is parsed with bare `float()` and therefore accepts `"-1"`, `"nan"`, and `"1e999"` (→ `inf`).

The fix guards both ends:

- **Consumption** (`vane/ai/functions.py`): the wait computation duplicated across `_retry_call` / `_retry_call_async` is extracted into `_retry_wait_seconds`, which honours a finite non-negative `retry_after` (capped at 120s, as before) and falls back to the normal exponential backoff for anything else, so no malformed delay can reach a sleep. An int beyond float range is treated as infinite rather than letting the conversion itself raise `OverflowError`.
- **Construction** (`vane/ai/providers/google.py`): a parsed header value that is negative or non-finite is treated the same as a missing/unparseable header and falls back to the documented 5-second default.

**Design note:** the originating review comment (#397, r3711116179) suggested *clamping* the delay to a finite non-negative range. This PR instead **discards** malformed values (5s default on the Google parse side, exponential backoff on the consumer side), on the grounds that a garbage value should not be partially trusted — e.g. clamping `Retry-After: -1` to 0 would produce an immediate hot retry. Happy to switch to clamping if you prefer the letter of the suggestion.

Valid inputs are unchanged: header `"0"`, `retry_after=0.0`, fractional values, and the 120s cap all behave exactly as before.

## Related issue

close: #469

## Documentation impact

Behaviour is documented in the `RetryAfterError` and `_retry_wait_seconds` docstrings; internal retry mechanics are not part of the public docs.

## Validation

- Reproduced all three failure modes on the pre-fix code via the real `_retry_call` / `_retry_call_async` paths (Python 3.12.2): sync `retry_after=-1.0` → `ValueError` after 1 attempt; sync NaN → `ValueError`; async NaN → hang (confirmed via 10s `wait_for` timeout).
- `pytest tests/ai/test_ai_functions.py -k "TestRetryAfterError or TestGoogleRetryHandling"` — **27 passed** (16 new: helper unit cases incl. `±10**400` overflow pins, sync/async end-to-end with monkeypatched sleeps asserting `[1, 2, 4]` backoff and `on_error` honoured, Google header cases for `-1`/`-0.5`/`nan`/`NaN`/`inf`/`1e999`/`3.5`). The new tests fail on the pre-fix code.
- Local environment cannot build the native `_duckdb` extension, so the suite was run with the `vane`-package import stubbed; engine-dependent tests are covered by CI.
- `ruff check` / `ruff format` clean via pre-commit.

## Checklist

      and untrusted input have been considered. (This change hardens handling of
      an untrusted HTTP header.)

https://claude.ai/code/session_01UyPevr7aJRqRcuxhfxb56q

### DIFF
--- a/tests/ai/test_ai_functions.py
+++ b/tests/ai/test_ai_functions.py
@@ -1737,6 +1737,72 @@ class TestRetryAfterError:
         assert exc_info.value.__cause__ is None
         assert exc_info.value.__context__ is None
 
+    @pytest.mark.parametrize("bad", [-1.0, -0.001, float("nan"), float("inf"), 10**400, -(10**400)])
+    def test_retry_wait_seconds_rejects_malformed_retry_after(self, bad):
+        """A non-finite or negative retry_after never becomes the sleep value (issue #469)."""
+        from vane.ai.functions import RetryAfterError, _retry_wait_seconds
+
+        # Falls back to exponential backoff (1, 2, 4, ... capped at 30) rather
+        # than passing a value that time.sleep/asyncio.sleep would choke on.
+        # The 10**400 cases also pin that the guard itself must not raise
+        # OverflowError converting an int beyond float range.
+        assert _retry_wait_seconds(RetryAfterError(retry_after=bad), attempt=0) == 1
+        assert _retry_wait_seconds(RetryAfterError(retry_after=bad), attempt=2) == 4
+
+    def test_retry_wait_seconds_honors_valid_retry_after(self):
+        from vane.ai.functions import RetryAfterError, _retry_wait_seconds
+
+        assert _retry_wait_seconds(RetryAfterError(retry_after=7.0), attempt=0) == 7.0
+        assert _retry_wait_seconds(RetryAfterError(retry_after=0.0), attempt=0) == 0.0
+        assert _retry_wait_seconds(RetryAfterError(retry_after=999.0), attempt=0) == 120  # capped
+
+    @pytest.mark.parametrize("bad", [-1.0, float("nan"), float("inf")])
+    def test_retry_call_malformed_retry_after_does_not_break_sleep(self, monkeypatch, bad):
+        """Sync retry with a malformed delay retries and honours on_error, never ValueError."""
+        from vane.ai import functions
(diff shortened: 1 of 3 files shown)


kaka11chen:
@codex review

kaka11chen (approved):
LGTM

PR #527, opened 2026-08-09
[feat](udf) Define multi-actor class UDF semantics

### LINKED ISSUE #526: [UDF] Define multi-actor semantics for vane.cls and vane.cls.batch
### Problem

`vane.cls` and `vane.cls.batch` expose an `actor_number` parameter, but currently accept only the integer `1`. Values greater than `1` raise:

```text
actor_number must be exactly 1 for stateful vane.cls UDFs;
multi-actor state semantics are not defined
```

This restriction is enforced in the Python API, C++ payload construction, and Actor Pool runtime. The underlying non-stateful Actor backends already support multiple instances, so this is a state-semantics limitation rather than an Actor Pool limitation.

Each Actor creates an independent instance of the user class. If `actor_number > 1` is allowed, several behaviors are currently undefined:

- How data is assigned to Actors and whether affinity is required.
- Whether instance state is entirely local to each Actor.
- Whether an instance or batch may be restarted or retried after Actor loss.
- Whether calls with side effects may be replayed.

The current implementation uses one Actor, serial execution, and no restart or retry after Actor loss. This prevents silent state resets and duplicate side effects, but limits safely replicable workloads such as models, tokenizers, and read-only caches.

### Proposed outcome

Define the public state, routing, and failure contract for multiple class UDF instances before allowing `actor_number > 1`.

The decision should clarify:

- Whether `actor_number > 1` will be supported.
- What state guarantees multiple instances provide.
- How data is assigned to Actors.
- Whether Actor reconstruction or batch retries are allowed after failures.
- Whether `stateful` should continue to imply `side_effects`.
- If only one Actor remains supported, whether the parameter or documentation should be adjusted.

Local and Ray execution paths should follow the same public contract, with corresponding documentation and tests. The existing validation should remain in place until these semantics are defined.

Removing the validation without defining these behaviors could silently reset state or duplicate side effects. Keeping the current single-Actor contract remains a valid option, but would retain the throughput limit for safely replicable class UDFs and should be reflected clearly in the API and documentation.

Key-based routing, checkpointing, and state recovery across Actors are broader concerns and are out of scope for this issue.

### Additional context

Relevant implementation points include:

- `vane/_expression_udf.py::_validate_actor_number`
- `src/duckdb_py/python_udf_utils.cpp::ValidateStatefulActorContract`
- `duckdb/execution/udf_ray_actor_pool.py::_validate_stateful_actor_pool_contract`
- `tests/slow/test_expression_udf_stateful_fault.py::test_stateful_ray_actor_loss_fails_without_state_reset_or_batch_replay`

kaka11chen:
## Summary

- Allow class UDFs to use any positive `actor_number` and define the public
  contract: every Actor owns an independent ephemeral instance, batches are
  assigned without affinity or global ordering, and Actor reconstruction may
  reset instance-local state.
- Define replay semantics for every distributed UDF: retrying Task and Actor
  backends may replay function, class, or batch calls after failures. No
  distributed UDF provides exactly-once execution, so external effects must be
  idempotent. `VOLATILE` constrains optimizer rewrites but does not suppress
  runtime replay.
- Apply the class-instance contract consistently to Expression and Relation UDFs
  on local and Ray runners. Remove the distributed `stateful`/`side_effects`
  payload fields, keep Ray's standard Actor restart/task-retry policy, and
  default omitted Actor GPUs to zero.
- Route aliases registered with `vane.attach_function` through the same physical
  backend resolution and Task/Actor runtime as direct Expression UDFs. Keep
  direct and attached UDF operators `VOLATILE`, and add local replacement,
  real-Ray reconstruction/replay, multi-Actor, payload, GPU, SQL alias, and
  cross-API contract coverage.
- Disable the public module-level and connection-level `create_function` and
  `create_table_function` bindings while retaining their internal
  implementations. Use Vane decorators, Relation UDF methods, and
  `vane.attach_function` instead.

## Related issue

close: #526

## Documentation impact

      `AstroVela/vane-website`.

The public state, routing, retry, replay, volatility, GPU-default, and API
removal contracts are documented in the decorator/Relation docstrings and
`CHANGELOG.md`.

## Validation

- Incremental Release native build and non-editable install with
  `SKBUILD_BUILD_DIR=build/python-release` and `uv pip install . --no-build-isolation`
- `scripts/run_release_tests.sh`
  - non-Ray: 521 passed, 15 deselected
  - real Ray: 11 passed, 525 deselected
- `scripts/run_fast_tests.sh`
  - non-Ray: 6178 passed, 602 skipped, 118 deselected, 7 xfailed, 1 xpassed
  - shared Ray: 98 passed, 2 skipped, 6807 deselected
  - test-owned Ray cluster shard: passed
- Targeted UDF contract/lifecycle/SQL matrix: 264 passed
- Expression/Relation unified contract suite: 71 passed
- Real-Ray Actor reconstruction and replay fault test: 4 isolated runs passed;
  the result asserts all 4,097 values and both Actor indices
- `pre-commit run --all-files`
- `python scripts/check_source_license_headers.py --repo root`

Validation used a CPU-only local Ray cluster; the reconstruction test allocates
8 logical CPUs and does not require CUDA.

## Checklist

      licenses and are recorded where required. No new dependencies or assets
      are introduced.


kaka11chen:
@codex review

tests/slow/test_expression_udf_actor_retry_fault.py
```diff
@@ -0,0 +1,219 @@
...
+                break
+            time.sleep(0.02)
+        else:
+            raise AssertionError(f"interrupted actor call was not reconstructed and replayed: {snapshot}")
+
+        print("ACTOR_REPLAY_OBSERVED", snapshot, flush=True)
+        ray.get(control.release.remote(), timeout=5)
+        print("ACTOR_RELEASED", flush=True)
+        batches = future.result(timeout=45.0)
+        assert batches
```
QuakeWang: **Non-blocking:** This only verifies that the result is non-empty, so the test could still pass if Actor replay drops or duplicates rows. Please assert that the final result contains all 4,097 rows, ideally also validating the `value` column and the expected `actor_index` values.

src/vane_py/pyrelation/initialize.cpp
```diff
@@ -422,7 +428,10 @@ void DuckDBPyRelation::Initialize(py::handle &m) {
 	    .def("map", &DuckDBPyRelation::Map, py::arg("map_function"), py::kw_only(), py::arg("return_type").none(false),
 	         py::arg("batch_size") = py::none(), py::arg("cpus") = py::none(), py::arg("gpus") = py::none(),
 	         py::arg("execution_backend") = py::none(), py::arg("actor_number") = py::none(),
-	         py::arg("side_effects") = false, "Apply a row-wise scalar Python UDF through the unified UDF executor")
+	         "Apply a row-wise scalar Python UDF through the unified UDF executor. Callable classes use Actor "
+	         "backends: actor_number creates independent ephemeral instances, work has no Actor affinity or global "
+	         "ordering, and failures may reconstruct an Actor and replay a call. Exactly-once execution is not "
+	         "provided; external effects must be idempotent.")
```
QuakeWang: **Nit:** This idempotency guidance appears limited to callable classes, but regular `ray_task` UDFs may also be retried after system failures, and this PR removes `side_effects`. Please clarify whether this requirement applies to all distributed UDFs so users do not assume function backends provide exactly-once execution.

QuakeWang (commented):
No blocking issues found. Two non-blocking suggestions:

src/vane_py/pyrelation/initialize.cpp
kaka11chen: Addressed in ac574c883e. The contract now explicitly covers every retrying distributed Task and Actor backend across the changelog, Relation map/map_batches/flat_map docstrings, Expression function/class decorators, and attach_function: calls may be replayed, exactly-once execution is not provided, and external effects must be idempotent. It also clarifies that VOLATILE constrains optimizer rewrites but does not prevent failure replay.

tests/slow/test_expression_udf_actor_retry_fault.py
kaka11chen: Addressed in ac574c883e. The test now concatenates every returned partition and asserts exactly 4,097 rows, the complete sorted value range 0..4096 (so drops and duplicates both fail), and actor_index values {0, 1}. I also replaced the high-frequency release RPC with a unique local marker so the reconstruction observation is not starved by control traffic; the isolated real-Ray test passed four runs.

kaka11chen:
@codex review

QuakeWang (approved):
+1

PR #531, opened 2026-08-10
[chore](release) Harden package release process

kaka11chen:
## Summary

- remove the repository-maintained `CHANGELOG.md` and its sdist inclusion
- use the reviewed GitHub Release body as the canonical public change record, without `RELEASE_NOTES.md` templates
- document a draft-first, immutable release process and the required protected environments
- make one manually dispatched workflow build the distributions once, promote the exact files through TestPyPI and PyPI, attach attestations and supplemental assets, then publish the draft GitHub Release
- require a matching unused version tag, a non-empty draft release, and a release commit reachable from `main`

## Validation

- `scripts/format root --from-ref upstream/main --check`
- pre-commit `check-yaml`, `check-github-workflows`, and `check-toml`
- syntax validation of all 22 workflow shell blocks
- built and validated a real sdist from the rebased branch with `scripts/check_release_artifacts.py`
- confirmed the sdist contains neither `CHANGELOG.md` nor either `RELEASE_NOTES` filename

## Release notes

This is release-infrastructure work only. It does not bump the package version and does not introduce a repository release-notes file. Notes for the next package version will be reviewed in that release PR and copied into the GitHub Draft Release body.

## Future release branches

Formal publication always runs from an immutable `v*` tag, not directly from a branch. The current preflight requires the tagged commit to be reachable from `main` because the repository currently protects only its default branch; this is a present-day security allowlist, not a permanent restriction on the branching model.

If the project later introduces maintained release lines such as `release/0.2`, support should be added together with the corresponding controls:

- define and document the release-branch naming convention;
- protect `release/*` with pull-request review, required CI, conversation resolution, deletion protection, and non-fast-forward protection equivalent to `main`;
- update release preflight to accept a tagged commit reachable from either `main` or an allowed protected release branch;
- ensure each release branch contains a reviewed, trusted version of the release workflow; and
- continue to publish only from the matching immutable `v<version>` tag through the protected TestPyPI and PyPI environments.

The workflow should not pre-authorize `release/*` before those branches receive equivalent protection, because doing so would create a path around the review and CI guarantees enforced on `main`.


kaka11chen:
@codex review

RELEASE.md
```diff
@@ -1,70 +1,150 @@
...
+   scanning.
+3. Configure the `RELEASE_ARTIFACT_CONTENT_RULES` repository secret used by
+   trusted artifact validation.
+4. Create protected `testpypi` and `pypi` GitHub environments. Both accept only
+   `v*` tags, require maintainer approval, and disallow administrator bypass.
+5. Register `.github/workflows/release.yml` as a trusted publisher for the
+   `vane-ai` project on TestPyPI and PyPI, using the matching `testpypi` and
+   `pypi` environment names. Publishing intentionally has no API-token fallback.
+6. After the draft-first workflow has passed a build-only dry run, enable
+   GitHub immutable releases. Draft releases must remain mutable so the
```
QuakeWang, line 43: Since immutable releases lock the tag only after publication, could we also document a `v*` tag ruleset that prevents updates and deletions during the draft/publish window? This would keep the release tag aligned with the commit built by the workflow.

kaka11chen: Addressed in 4f82404de2. `RELEASE.md` now requires an active, no-bypass tag ruleset for `refs/tags/v*` that allows initial creation but blocks deletion and force-push updates throughout the draft/publish window. It also requires checking the ruleset before tag creation, verifying the remote tag resolves to the reviewed commit, and never moving/deleting/reusing a tag after creation. The matching repository ruleset, [Release Tag Immutability v1](https://github.com/AstroVela/vane/rules/20623932), is now active with `deletion` and `non_fast_forward` rules.

QuakeWang (approved):
+1

PR #576, opened 2026-08-14
[fix](ai) Reject model in vLLM engine args

### LINKED ISSUE #443: [P2][AI/vLLM] Explicit model leaves a duplicate model key in native options
## Affected area

Native vLLM provider option normalization.

## What happens

When the caller supplies `model` both as the explicit `get_prompter(model=...)` argument and in provider/options state, the explicit argument wins for `model_name`, but the options copy still retains its own `model` key. That retained key reaches the native engine argument map.

The engine is then constructed with `AsyncEngineArgs(model=self.model, **self.engine_args)`, so the same keyword can be supplied twice and fail during engine initialization.

## Source evidence

At commit `64a776c07959552edddaadb8e446cb20bd8c6f50`:

- `vane/ai/providers/vllm.py:191-205` builds `merged` and evaluates `model or merged.pop("model", ...)`. Python short-circuiting means the `pop` is skipped when the explicit argument is truthy.
- `duckdb/execution/vllm.py:338-356` passes the selected model explicitly and expands the retained engine arguments.

## Impact

A valid-looking combination of provider defaults and per-call overrides can turn into a runtime initialization error instead of deterministic override behavior.

## Verification scope

This report concerns the native vLLM path and the case where both model sources are populated.

QuakeWang:
## Summary

vLLM Prompt planning passes the public `model` separately from `engine_args`. Allowing `engine_args.model` therefore supplies `model` twice to `AsyncEngineArgs`, raising `TypeError` before engine initialization.

Reject direct `engine_args.model` during planning and direct users to the top-level `model` argument. Preserve valid nested fields such as `engine_args.speculative_config.model`, with Python and SQL regression coverage.

## Related issue

close: #443

## Documentation impact

      `AstroVela/vane-website`.

## Validation

- Pre-commit on both changed files, including Ruff and mypy
- `scripts/run_installed_pytest.sh tests/fast/test_vllm_sampling_param_fold.py`: 53 passed
- `scripts/run_release_tests.sh`: 526 passed, 4 skipped; real-Ray: 11 passed

## Checklist

### DIFF
--- a/tests/fast/test_vllm_sampling_param_fold.py
+++ b/tests/fast/test_vllm_sampling_param_fold.py
@@ -162,6 +162,21 @@ def test_public_vllm_rejects_guided_decoding_alias_mapping():
         _plan_from_prompt_options({"generate_args": {"sampling_params": {"guided_decoding": {"regex": ".*"}}}})
 
 
+@pytest.mark.parametrize(
+    "factory",
+    [
+        pytest.param(_plan_from_prompt_options, id="prompt-entry-point-default-model"),
+        pytest.param(
+            lambda options: VLLMProvider().get_prompter(model="selected-model", options=options),
+            id="provider-factory-explicit-model",
+        ),
+    ],
+)
+def test_public_vllm_rejects_model_in_engine_args_during_planning(factory):
+    with pytest.raises(ValueError, match=r"engine_args\.model.*top-level 'model' argument"):
+        factory({"engine_args": {"model": "engine-model"}})
+
+
 @pytest.mark.parametrize("sampling_params", [{"temperature": -0.1}, '{"temperature":-0.1}'])
 @pytest.mark.parametrize(
     "factory",
@@ -291,6 +306,16 @@ class TestNoOpCases:
 
 class TestSQLPathFolds:
     """SQL struct_pack options flow through get_prompter into the same fold."""
+
+    def test_sql_rejects_model_in_engine_args_during_planning(self):
+        from vane.ai._sql import build_ai_prompt_sql_spec
+
+        with pytest.raises(ValueError, match=r"engine_args\.model.*top-level 'model' argument"):
+            build_ai_prompt_sql_spec(
+                provider="vllm",
+                model="selected-model",
+                options={"engine_args": {"model": "engine-model"}},
+            )
 
     def test_sql_top_level_options_fold(self):
         from vane.ai._sql import _normalize_sql_options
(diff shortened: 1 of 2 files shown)


kaka11chen (approved):
LGTM

caomaocao:
Reviewed this against #443 by tracing the option flow end to end.

### Correctness

- **Type safety of the check**: `"model" in engine_args` runs after the existing validation (`vane/ai/providers/vllm.py` lines 136–141) has already enforced that `engine_args` is a Mapping with string keys, so the membership test can't misfire on a non-mapping.
- **All public entry points are covered**: Python `prompt()`, direct `VLLMProvider().get_prompter()`, and SQL (`build_ai_prompt_sql_spec`) all funnel through `_prepare_prompt_call` → `get_prompter` → `_validate_vllm_prompt_options`. The three added test cases map to exactly these paths, including the default-model path (`model=None` → `DEFAULT_MODEL`), which also collided before.
- **`model` is exactly the right key set**: the executor passes only `model=` explicitly to `AsyncEngineArgs` (`vane/execution/vllm.py:353,372`); no other engine-arg key can collide, and executor defaults use `setdefault` without touching `model`. So rejecting this one key is complete, neither under- nor over-scoped.
- **Not over-broad**: the check is non-recursive, so `engine_args.speculative_config.model` still passes. The pre-existing positive test `test_public_vllm_remote_code_accepts_a_commit_sha` asserts it survives planning, which also guards against a future recursive rewrite.
- **No behavior break**: this combination never worked — it always crashed at engine init with a duplicate-keyword `TypeError`. Converting that into an early planning `ValueError` breaks no working code.

### Contract note (informational)

#443's impact statement asks for "deterministic override behavior instead of a runtime initialization error," which admits two readings: silently let the explicit `model` win, or reject early. Rejection seems like the better call — silently dropping a user-specified model hides a real configuration mistake, and it matches this file's existing precedent (`guided_decoding` → `ValueError` pointing at `return_format`). Also worth noting the issue references an older code shape (`model or merged.pop("model", ...)` at `64a776c0`) that no longer exists; today a top-level `options["model"]` is already rejected as an unknown option, so `engine_args.model` was the only remaining duplicate-model vector — the PR maps the issue onto current code correctly.

### Minor observations (no change requested)

- `NativeVLLMPromptPlan.__post_init__` doesn't re-run validation, so constructing the dataclass directly bypasses the guard. All public paths go through `get_prompter`, and only tests construct it directly today — just a known seam.
- The low-level executor would still surface the raw `TypeError` if something ever reached it with a `model` key; defense-in-depth there is possible, but it's outside this issue's planning scope and I wouldn't expand this PR for it.

Error style, check ordering, and the parametrized-factory test pattern all match the file's existing conventions.

LGTM.

PR #599, opened 2026-08-14
[fix](ai) Distinguish provider import and construction errors

### LINKED ISSUE #100: [P2][Bug] AI on_error values, logging, and provider ImportError handling are inconsistent
## Problem

AI error policy is not one contract:

- `on_error="log"` can return a fallback without writing a log;
- invalid values can fall through like ignore/null;
- VLLM options and runtime accept different value sets (`ignore` versus `null`);
- the provider loader can wrap an ImportError raised inside provider construction as a misleading missing-extra message.

QuakeWang:
## Summary

Provider loaders classified any `ImportError` from importing an internal
`vane.ai.providers.*` module as a missing optional dependency. This could hide
Vane regressions behind an incorrect `pip install vane-ai[...]` hint.

Let internal provider import and construction errors propagate unchanged.
Translate only matching `ModuleNotFoundError` values at the actual optional-SDK
import sites, and preserve the install hint across worker-safe error wrapping
and serialization. Add regression coverage for all five built-in providers.

## Related issue

close: #100

## Documentation impact

- [ ] A follow-up issue is required in `AstroVela/vane-website`.

## Validation

- Pre-commit on all changed files, including Ruff and mypy
- Full AI suite: 651 passed, 6 deselected
- vLLM runtime suite: 70 passed
- `scripts/run_release_tests.sh`: 527 passed, 4 skipped; real-Ray: 11 passed

## Checklist


tests/ai/test_descriptors.py
```diff
@@ -23,6 +25,63 @@ def test_load_unknown_provider_raises(self):
...
+        ("provider_name", "module_name"),
+        [
+            ("transformers", "vane.ai.providers.transformers"),
+            ("openai", "vane.ai.providers.openai"),
+            ("vllm", "vane.ai.providers.vllm"),
+            ("anthropic", "vane.ai.providers.anthropic"),
+            ("google", "vane.ai.providers.google"),
+        ],
+    )
+    def test_provider_module_import_error_reports_missing_extra(self, monkeypatch, provider_name, module_name):
```
kaka11chen: Non-blocking: this test currently classifies every `ImportError` raised while importing a bundled `vane.ai.providers.*` module as a missing provider extra. These modules do not import their optional SDKs at module scope; those imports happen later (for example `openai`, `anthropic`, `google.genai`, `sentence_transformers`, and `vllm`). An internal Vane import regression or a broken base installation can therefore still receive an actionable-looking but incorrect `pip install vane-ai[...]` hint, and this test would lock that behavior in.

Could we avoid diagnosing errors at this internal-module boundary? A more precise contract would be:

1. Let errors from importing `vane.ai.providers.*` propagate unchanged.
2. At the actual optional-SDK import sites, translate only `ModuleNotFoundError` whose `exc.name` matches the expected optional module or namespace prefix.
3. Preserve every other `ImportError` / `ModuleNotFoundError` unchanged.
4. Change this test to assert that an internal provider-module error remains original, and add focused tests for genuinely missing optional modules.

Because the real SDK imports may execute on workers, the friendly error also needs to survive serialization and the safe execution-error wrapper (for example by making `ProviderImportError` pickle-safe and preserving a fixed, library-owned install hint). If that runtime handling is intentionally out of scope here, I suggest not making the broad classification a regression contract and tracking it in a focused follow-up. The constructor-outside-`try` change itself looks correct.

kaka11chen (commented):
Non-blocking review note on the remaining import-error classification.

tests/ai/test_descriptors.py
QuakeWang: @kaka11chen I have addressed this, PTAL

kaka11chen:
@codex review

vane/ai/provider.py
```diff
@@ -81,8 +105,10 @@ def _safe_provider_execution_error(
     model: str,
     operation: str,
     original_error: Exception,
-) -> RuntimeError:
+) -> ProviderImportError | RuntimeError:
     """Return a public-safe final error after provider retry handling."""
+    if isinstance(original_error, ProviderImportError):
+        return ProviderImportError(original_error.extra, function=original_error.function)
```
kaka11chen: Blocking: this bypasses the credential-safe error boundary for any `ProviderImportError` instance. Both `extra` and `function` are arbitrary, unbounded strings, and a custom provider/descriptor can construct and raise this importable exception. I reproduced the real `_EmbedTextBatch` initialization path with `ProviderImportError("sk-review-sentinel", function="sk-review-sentinel")`; the sentinel is returned verbatim in the worker/driver error, whereas current `main` sanitizes the same input to the exception type only. This conflicts with the fixed, library-owned hint requirement and the policy above that arbitrary upstream text must not be inspected or stringified. Please reconstruct only from canonical, allowlisted provider/function keys (or otherwise distinguish internally minted errors), fall back to the generic sanitized `RuntimeError` for unknown values, and add a sentinel regression test through the batch wrapper plus pickle/cloudpickle.

kaka11chen (commented):
Found one error-sanitization issue that should be addressed before merge.

kaka11chen:
@codex review

kaka11chen (approved):
LGTM

PR #606, opened 2026-08-16
[feat](ai) Introduce SGLang backend behind a unified engine executor framework

suxiaogang223:
## Summary

Add SGLang 0.5.17 as a native inference backend and generalize the existing vLLM path behind a shared engine-agnostic executor framework. `provider="sglang"` now lowers end to end through prompt planning, SQL execution, local execution, and Ray actor pools.

## Changes

- **Per-engine executor dispatch** — replace the C++ single-slot vLLM factory with `RegisterEngineExecutorFactory`, dispatching through the versioned native options envelope while retaining `engine="vllm"` as the compatibility default.
- **Unified executor hierarchy** — introduce the shared `LLMExecutor`, `LocalEngineExecutor`, and `RayActorExecutorMixin` lifecycle and scheduling layers, with vLLM and SGLang supplying backend-specific engine hooks.
- **Native async SGLang runtime** — create `sglang.Engine` on its executor event loop and call the public `Engine.async_generate` API on that same loop. Sampling parameters remain plain dictionaries and additional generation arguments are forwarded unchanged.
- **Coordinated engine lifecycle** — reject new requests during shutdown, wait for initialization, cancel pending generation, run backend shutdown on the owning event loop, stop the loop, and join the executor thread. SGLang and vLLM both use explicit backend teardown.
- **Ray integration** — support SGLang actor classes in shared named pools, driver-side pool precreation, background-loop controls inside generic Ray actors, native execution controls, and sealed-secret restoration before local or named-pool engine construction.
- **Prompt semantics and validation** — lower `max_tokens`, `max_new_tokens`, `temperature`, structured output schemas, batch size, and actor count into the SGLang physical plan. Validate sampling, resource, timeout, and engine/generation argument values during planning.
- **Versioned dependency** — add the Linux x86-64 `sglang` extra pinned to `sglang==0.5.17`. The README distinguishes the upstream default CUDA 13 dependency set from the separately installed CUDA 12.9 wheel overrides.

No signal suppression, `signal.signal` monkey patch, synchronous generation wrapper, or thread-offloaded generation workaround is used.

## Validation

- CPU tests with an injected public-shape `sglang.Engine` cover SQL/provider round trips, concurrent async generation, dictionary sampling parameters, generation argument forwarding, event-loop affinity, explicit teardown, shutdown during initialization, cancellation ordering, Ray actor construction, execution-control lowering, and sealed-secret restoration.
- The relevant installed-package suites pass: 35 focused checks and 505 AI/executor lifecycle tests.
- A real Ray + SGLang GPU smoke test passed twice using NVIDIA Ada (compute capability 8.9), driver 570.207, SGLang 0.5.17, PyTorch 2.11.0+cu129, `sglang-kernel` 0.4.5+cu129, and Ray 2.58.0. Runs completed in 98.16 s and 63.56 s and produced `DuckDB is a database`.
- Both real-GPU runs shut down cleanly, released GPU memory, and left no Ray/SGLang processes. Logs contained no signal-thread, CUDA, undefined-symbol, or crash errors.

## Compatibility and operational notes

- A normal `vane-ai[sglang]` installation follows the upstream SGLang 0.5.17 CUDA 13 dependency set.
- CUDA 12.9 installations must explicitly reinstall the documented cu129 PyTorch and SGLang kernel wheels. Python package metadata cannot select a host-specific wheel index, and a later unconstrained dependency sync may replace those wheels with CUDA 13 variants.
- SGLang 0.5.17 shutdown owns its engine subprocess tree. In-process deployments therefore assume that those child processes are dedicated to SGLang; stronger production isolation can place the engine in a dedicated worker process.


vane/execution/sglang.py
```diff
@@ -0,0 +1,227 @@
...
+            return SamplingParams()
+        if isinstance(sampling_params, SamplingParams):
+            return sampling_params
+        if isinstance(sampling_params, str):
+            try:
+                sampling_params = json.loads(sampling_params)
+            except json.JSONDecodeError as exc:
+                raise ValueError("sglang sampling_params JSON could not be parsed") from exc
+        if isinstance(sampling_params, dict):
+            return SamplingParams(**sampling_params)
```
kaka11chen: **[P1] Keep sampling parameters in the shape accepted by `Engine.generate`**

This converts every dict (and the default) into SGLang's internal `SamplingParams` object, but the released offline Engine API accepts `Dict | List[Dict]`; `GenerateReqInput._handle_parallel_sampling()` treats any non-dict as a list and calls `.get()` on its first element. Real SGLang generation can therefore fail during request normalization even though `_FakeEngine` passes because it discards this argument. Please keep/validate a dict (or explicitly convert back to one) and pin/test the supported SGLang version.

References: https://github.com/sgl-project/sglang/blob/v0.5.17/python/sglang/srt/entrypoints/engine.py#L343-L426 and https://github.com/sgl-project/sglang/blob/v0.5.17/python/sglang/srt/managers/io_struct.py#L440-L450

vane/execution/_llm_executor.py
```diff
@@ -0,0 +1,825 @@
...
+    def shutdown(self) -> None:
+        if self._shutdown_called:
+            return
+        self._shutdown_called = True
+        self._finished_submitting = True
+        self._notify_state_change(force=True)
+        loop = getattr(self, "loop", None)
+        if loop is not None and loop.is_running():
+            loop.call_soon_threadsafe(loop.stop)
+        self._shutdown_engine()
```
kaka11chen: **[P1] Close engines that finish initialization after shutdown starts**

Construction waits only for `loop_ready`, not `engine_ready`. If shutdown happens while `_create_engine()` is still blocking, this call sees `self.llm is None` and does nothing; the initializer can then assign a live SGLang Engine after shutdown, while the event-loop thread exits. A deterministic slow-Engine repro leaves `llm` alive with `shutdown_called == False`, leaking scheduler subprocesses/GPU resources. Please coordinate initialization and shutdown (for example, close a newly created engine when `_shutdown_called` is already set) and join the loop thread.

vane/execution/sglang.py
```diff
@@ -0,0 +1,227 @@
...
+
+        return RemoteVLLMExecutor(llm_actors, pool_name=pool_name)
+
+    opts = _restore_native_vllm_secrets(opts)
+    return SGLangLocalExecutor(
+        model,
+        opts["engine_args"],
+        opts["generate_args"],
+        on_error=opts["on_error"],
+        engine_init_timeout_s=opts.get("engine_init_timeout_s"),
```
kaka11chen, line 240: **[P1] Forward the forced-background-loop options to the local executor**

`NativeSGLangPromptPlan` deliberately writes `use_threading=True` and `_force_background_thread=True`, but this constructor call drops both. Inside a generic Ray actor outside the named-pool path, `_detect_ray_actor()` then selects actor mode and no `self.loop` is created; the synchronous native bridge later calls `submit()`, which schedules work on `self.loop` and fails. Please forward these options the same way the vLLM builder does.

vane/execution/vllm.py
```diff
@@ -2509,17 +1807,32 @@ def ensure_named_vllm_pools_for_plan(
...
-            opts = _restore_native_vllm_secrets(normalize_options(raw_opts))
+            engine = raw_opts.get("engine") if isinstance(raw_opts, dict) else None
+            if engine not in ("vllm", "sglang"):
+                raise ValueError("distributed inference node options are missing a valid 'engine' field")
+
+            if engine == "sglang":
+                from vane.execution.sglang import SGLangRayLocalExecutor
+                from vane.execution.sglang import normalize_options as normalize_sglang_options
+
+                opts = normalize_sglang_options(raw_opts)
```
kaka11chen: **[P2] Restore sealed SGLang options before creating the named actor pool**

`normalize_sglang_options()` deliberately leaves secret-reference markers plus the opaque payload in place, but this branch immediately extracts `engine_args`/`generate_args`. The adjacent vLLM branch calls `_restore_native_vllm_secrets()` first. With a sealed value, the SGLang actor currently receives `{'__vane_vllm_secret_ref': 0}` instead of the credential, so authenticated engine creation fails. Please restore before extracting the actor arguments and cover this branch with a sealed-secret test.

vane/ai/functions.py
```diff
@@ -1724,7 +1725,7 @@ def _prepare_prompt_call(
         provider_name = getattr(resolved_provider, "name", type(resolved_provider).__name__)
         raise ValueError(f"Provider {provider_name!r} is not a prompt provider") from exc
 
-    if isinstance(descriptor, NativeVLLMPromptPlan):
+    if isinstance(descriptor, NativeInferencePlan):
```
kaka11chen, line 1728: **[P2] Preserve native execution controls for SGLang plans**

Before reaching this branch, `_prepare_prompt_call` pops `batch_size`, `actor_number`, and `max_retries`, then restores them only when `family == 'vllm'`. SGLang declares these as accepted native prompt options, so `batch_size=7` and `actor_number=3` are silently lost; execution falls back to batch size 128 and concurrency 1. Please generalize the native-plan lowering and map `actor_number` to SGLang `concurrency` as the vLLM plan does.

kaka11chen (commented):
Reviewed the current head (`8647dc8`). The shared executor abstraction is a good direction, but the inline comments below cover remaining runtime, lifecycle, distributed-option, packaging, and compatibility gaps. Required CI is also currently red in the shared-Ray shard: https://github.com/AstroVela/vane/actions/runs/32140982178/job/95740562886

vane/ai/provider.py
```diff
@@ -216,6 +216,15 @@ def _load_vllm(name: str | None = None) -> Provider:
     return VLLMProvider(name)
 
 
+def _load_sglang(name: str | None = None) -> Provider:
```
kaka11chen, line 219: **[P2] Add an installable, versioned SGLang extra before registering this provider**

The missing-dependency path raises `ProviderImportError('sglang')`, whose public message tells users to run `pip install 'vane-ai[sglang]'`, but `pyproject.toml` defines no `sglang` extra or supported version/platform constraint. That command installs only the base package and leaves the provider unusable. Please add the optional dependency (plus typing/lock metadata as appropriate) so the runtime API has an explicit compatibility target.

kaka11chen:
@codex review

kaka11chen:
@codex review

kaka11chen:
@codex review

vane/ai/providers/sglang.py
```diff
@@ -0,0 +1,238 @@
...
+        if name not in copied:
+            continue
+        value = copied[name]
+        if not isinstance(value, Mapping):
+            raise TypeError(f"Prompt option {name!r} must be a mapping")
+        _validate_sglang_json(value, name)
+
+    engine_args = copied.get("engine_args", {})
+    if "model_path" in engine_args:
+        raise ValueError("Prompt option 'engine_args.model_path' is not supported; use the top-level 'model' argument")
```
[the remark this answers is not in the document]
kaka11chen: Tracked as follow-up issue #669: https://github.com/AstroVela/vane/issues/669. Resolving this PR thread so the remaining remote-code hardening is handled explicitly after #606.

vane/execution/sglang.py
```diff
@@ -0,0 +1,242 @@
...
+            from sglang import Engine  # type: ignore[import-not-found, import-untyped, unused-ignore]
+
+        self.llm = Engine(model_path=self.model, **self.engine_args)
+
+    async def _run_generate(self, prompt: str, request_id: str) -> str:
+        del request_id  # SGLang's offline Engine does not require Vane's request id.
+        output = await self.llm.async_generate(
+            prompt=prompt,
+            sampling_params=self.sampling_params,
+            **self.generate_args,
```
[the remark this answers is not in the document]
kaka11chen: Tracked together with the executor-owned `rid` collision as follow-up issue #672: https://github.com/AstroVela/vane/issues/672. Resolving this PR thread and keeping the deferred validation work visible in the issue.

kaka11chen (approved):
LGTM

Issue #610, opened 2026-08-16
[macOS arm64] Source build fails and no binary wheels are published


figurant:
### Affected area

Build or installation; distributed execution / Ray.

### What happened?

`pip install vane-ai` on an Apple Silicon Mac falls back to the 0.1.0 sdist because PyPI does not currently contain macOS arm64 wheels. The sdist build cannot produce the full distributed/Arrow Flight extension in a clean environment: it first lacks the Arrow C++ package, and after the native dependencies are provisioned it hits several macOS/libc++ portability errors in Vane and Arrow.

The expected user experience is that a supported Apple Silicon Mac can install a prebuilt wheel and import Vane without installing Homebrew, CMake, vcpkg, Arrow, or a compiler:

```bash
python -m pip install --only-binary=:all: vane-ai
python -c 'import vane; print(vane.__version__, vane.__engine_version__)'
```

The wheel must retain the full distributed implementation, including Arrow Flight exchange; publishing a build with the Flight stub would not resolve the issue.

### Reproduction

On an arm64 Mac:

```bash
uv venv --python 3.12
source .venv/bin/activate
uv pip install --no-cache vane-ai
```

The initial sdist build reaches CMake and fails because `ArrowConfig.cmake` is unavailable:

```text
Could not find a package configuration file provided by "Arrow" with any of
the following names:

  ArrowConfig.cmake
  arrow-config.cmake
```

After installing the manifest dependencies with vcpkg and building with `BUILD_DISTRIBUTED_EXCHANGE=ON`, the following independent portability failures are reproducible:

1. Arrow 24 selects `std::log2p1` with the older Apple libc++ feature set:

   ```text
   arrow/util/bit_util.h:146:15: error: no member named 'log2p1' in namespace 'std'
   ```

2. `ExchangeSourceFile::file_size` is serialized as `size_t`. On macOS arm64 this is `unsigned long`, while DuckDB's serializer has no matching overload:

   ```text
   no matching member function for call to 'Read'
   requirement 'std::is_same<unsigned long, unsigned long long>::value' was not satisfied
   ```

3. `shuffle_cache.cpp` calls `getpid()` without including its POSIX declaration:

   ```text
   error: use of undeclared identifier 'getpid'
   ```

4. The Ray binding uses `std::atomic<std::shared_ptr<ColumnDataCollection>>`, which is unavailable in the tested Apple libc++:

   ```text
   error: _Atomic cannot be applied to type 'std::shared_ptr<...>'
[...]

PR #623, opened 2026-08-20
[fix](macos) Compile distributed sources with libc++

figurant:
## Summary

- Let the active macOS toolchain supply the deployment target instead of forcing macOS 10.7, so libc++ APIs used by the C++20 build remain available.
- Serialize exchange file sizes with a fixed `uint64_t` wire type and explicitly include the POSIX declaration for `getpid()`.
- Preserve thread-safe Ray result materialization with the libc++-supported `std::atomic_load` / `std::atomic_store` shared-pointer API.
- Add a native regression assertion that round-trips a file size larger than 32 bits.

This PR is intentionally limited to source compatibility. macOS wheel/release automation, dependency provisioning, and installation acceptance scripts remain follow-up work under #610.

## Related issue

Refs #610

## Documentation impact

      `AstroVela/vane-website`.

The changes only make existing distributed functionality compile on macOS; they do not add or change a public API.

## Validation

- `scripts/run_native_tests.sh "[serialization][physical_plan][exchange]"` — 183 assertions in 12 test cases passed.
- `uv build --wheel --out-dir /tmp/vane-issue-610-dist --python 3.12 .` — full CPython 3.12 native wheel built successfully on Linux x86_64, including `task.cpp` and `ray_module.cpp`.
- Installed the wheel into an isolated virtual environment and verified `import vane`, engine/version metadata, and `SELECT 42` with `VANE_RUNNER=local-fast`.
- Ran `tests/fast/test_ray_cpp_bindings.py::test_ray_backed_result_partition_concurrent_materialization` against the installed wheel — 2 parameterized cases passed (success and cached failure across eight concurrent callers).
- The equivalent source compatibility changes were previously validated on arm64 macOS 26 with CPython 3.12, including a two-node local Ray plan with Arrow Flight shuffle; this branch was independently compiled and regression-tested in the Linux environment above.

## Checklist

### DIFF
--- a/cmake/duckdb_loader.cmake
+++ b/cmake/duckdb_loader.cmake
@@ -576,11 +576,8 @@ function(_duckdb_create_interface_target target_name)
                 /utf-8 # treat source files as UTF-8 encoded
     )
   elseif(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
-    target_compile_options(
-      ${target_name}
-      INTERFACE -stdlib=libc++ # for libc++ in favor of older libstdc++
-                -mmacosx-version-min=10.7 # minimum osx version compatibility
-    )
+    # Use libc++ on macOS; the deployment target is supplied by the toolchain.
+    target_compile_options(${target_name} INTERFACE -stdlib=libc++)
   endif()
 
   # Link to the DuckDB static library
--- a/external/duckdb/src/execution/distributed/exchange/shuffle_cache.cpp
+++ b/external/duckdb/src/execution/distributed/exchange/shuffle_cache.cpp
@@ -33,6 +33,7 @@
 #include <sstream>
 #include <unordered_map>
 #include <unordered_set>
+#include <unistd.h>
 
 namespace duckdb {
 namespace distributed {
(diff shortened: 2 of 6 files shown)


caomaocao (approved):
Reviewed on real Apple Silicon hardware (macOS 26.5 arm64, AppleClang 21, CPython 3.12). LGTM.

Verified locally at this branch: `exchange_source_task.cpp` and `task.cpp` pass `clang++ -fsyntax-only -std=c++20`; without the fix, `exchange_source_task.cpp` reproduces exactly the reported serializer failure on my machine. `std::atomic_load`/`std::atomic_store` on `shared_ptr` produce no warnings or errors on AppleClang 21 in `-std=c++20` (nor `-std=c++2c`), and compile clean under both `-mmacosx-version-min=10.7` and `=12.0` — no objection to this over a mutex. Grepped the branch for remaining `Property<size_t>` / `std::atomic<std::shared_ptr>` sites: none left — nice catch on the one in `test_physical_plan_serialization.cpp`, which I had missed. `scripts/format duckdb --from-ref main --check` passes.

One toolchain caveat: AppleClang 21's libc++ accepts even the old constructs under the 10.7 flag, so I can't re-reproduce the original AppleClang 14 failures locally — my validation covers "no regression on the current toolchain"; your original arm64 run covers the older one.

One question: with the hard-coded flag gone, nothing in-tree sets `CMAKE_OSX_DEPLOYMENT_TARGET`, so source builds target the host OS. I assume the macOS 12 floor lands with the cibuildwheel config in the follow-up? Happy to take that follow-up — I have the hardware to test wheel installs on a clean arm64 environment.

kaka11chen (approved):
LGTM

PR #641, opened 2026-08-22
[fix](ray) Execute UNION plans

### LINKED ISSUE #628: [bug](ray) UNION and UNION ALL relations cannot execute through Ray Runner
### Affected area

Distributed execution or Ray

### What happened?

Relations containing `UNION ALL` or `UNION` cannot be executed through the Ray runner. DuckDB produces a `UNION` physical operator, which distributed-plan translation rejects before execution.

Both forms execute successfully with `VANE_RUNNER=local-fast` and should return the same rows through Ray.

`INTERSECT` and `EXCEPT` do not show this failure in the tested cases because DuckDB rewrites them to supported hash join and aggregate operators.

### Reproduction

```python
import ray
import vane
from vane import runners

ray.init(num_cpus=2)
connection = vane.connect()

try:
    relation = connection.sql("""
        SELECT x FROM (VALUES (1), (2)) AS left_input(x)
        UNION ALL
        SELECT x FROM (VALUES (3), (4)) AS right_input(x)
    """)
    runners.set_runner_ray(noop_if_initialized=True)
    runner = runners.get_or_create_runner()
    print(list(runner.run_iter_tables(relation)))
finally:
    connection.close()
    ray.shutdown()
```

Observed error:

```text
Invalid Input Error: Ray runner cannot execute this query: DuckDBError::ValueError failed to translate physical plan: Distributed pipeline does not support operator type: UNION
```

Replacing `UNION ALL` with `UNION` produces the same translation error.

### Environment

```text
OS: Linux 6.17.0-29-generic, glibc 2.39
Architecture: x86_64
Python: 3.12.3
Vane: 0.2.0.dev564; upstream/main inspected at 08152e15b921d2fb62f2efff64696df3ad5e13e4
Engine: v1.5.5-vane.9d91c17ba6
Ray: 2.57.0, local cluster with 2 CPUs
VANE_RUNNER: ray
```

### Additional context

The failure also occurs when each branch reads Parquet files. `UNION DISTINCT` adds a hash aggregate above the union, but translation still fails when it reaches the `UNION` child. The relevant translator source is unchanged between the tested engine source and the inspected `upstream/main` commit.

kaka11chen:
## Summary

- add a first-class N-way distributed `UnionNode` that fans in independently translated branch task streams instead of placing `PhysicalUnion` in worker plans
- preserve each UNION branch's task, fragment, resource, and lineage identity, including optimizer-pruned and statically empty branches, without a local execution fallback
- propagate the effective downstream order requirement from the result consumer through physical operators before constructing distributed nodes; LIMIT, OFFSET, streaming order-sensitive operators, and order-dependent sinks force branch-ordered UNION fan-in, while order-insensitive materialization barriers retain unordered fan-in
- make the serialized remote exchange sink plan purely static: it carries exchange layout, query/output metadata, partitioning expressions, and range metadata, but no task or attempt ownership
- use one scheduler-owned task partition identity for each logical exchange-producing task; allocate it before submission, retain it across retries, and bind only the current attempt ID when an attempt is submitted
- enforce the worker-plan invariant that a task contains at most one remote exchange sink across scheduler inspection, runtime binding, and worker execution instead of silently selecting the first sink
- apply the same task identity contract to UNION, repartition, broadcast join, materialized exchange, ORDER BY/range exchange, and source-free tasks, with worker-side validation of the submitted binding
- add native and Ray coverage for plan serialization, strict binding, retry stability, collision detection, registration-order independence, UNION variants, ordered LIMIT/OFFSET consumption, empty branches, Parquet scans, ORDER BY, and broadcast joins

Ray Runner can now execute `UNION` and `UNION ALL` plans directly while preserving DuckDB's effective insertion-order contract when downstream semantics depend on it. Exchange sink ownership follows a Trino-style boundary: the plan describes static work, the scheduler owns the sole logical task identity, and each submission supplies only the attempt-specific binding. This intentionally introduces no compatibility or fallback path.

## Related issue

close: #628

## Documentation impact

      `AstroVela/vane-website`.

This implements existing SQL behavior in Ray Runner and changes an internal execution contract without adding a user-facing API.

## Validation

Linux x86-64, Python 3.12, local CPU Ray. Native build concurrency was set to 16 with `CMAKE_BUILD_PARALLEL_LEVEL=16`, `VANE_NATIVE_BUILD_JOBS=16`, and `cmake --build ... --parallel 16`.

- two consecutive code and semantic review passes completed without findings after the review fixes
- `scripts/format workspace --from-ref upstream/main --check`
- `pre-commit run --files <all changed files>`
- `git diff --check`
- `uv pip install . --no-build-isolation`
- `cmake --build build/native-cxx20 --target unittest --parallel 16`
- `build/native-cxx20/test/unittest "[distributed][union]"` — 74 assertions in 4 test cases passed
- `build/native-cxx20/test/unittest "[serialization][physical_plan][exchange]"` — 181 assertions in 13 test cases passed
- `build/native-cxx20/test/unittest "[distributed][repartition]"` — 18 assertions in 1 test case passed
- `build/native-cxx20/test/unittest "Remote exchange sink: MARK build summary records empty and NULL-bearing inputs"` — 6 assertions in 1 test case passed
- `build/native-cxx20/test/unittest "[union]"` — 723 assertions in 18 test cases passed; 1 test case skipped because the JSON extension was unavailable
- focused exchange-sink/FTE/native/Ray fragment pytest selection — 21 passed, 846 deselected
- `scripts/run_installed_pytest.sh -vv tests/fast/test_ray_e2e.py -k "union_limit_and_offset or ray_union_"` — 8 passed, 78 deselected
- `scripts/run_release_tests.sh` — 554 base tests passed, 15 deselected; 11 real-Ray tests passed, 558 deselected

## Checklist


kaka11chen:
@codex review

kaka11chen:
@codex review

hubgeter (changes requested):
## Review summary\n\nI reviewed commit  against base .\n\nThe distributed UNION fan-in and scheduler-owned exchange identity changes are otherwise coherent. I found one correctness issue that should be fixed before merging.\n\n## Required change\n\nThe new  receives only . Upstream  also derives  from downstream pipeline and sink order dependencies. On the distributed path, this information is lost, so a downstream order-sensitive / can consume UNION branches in a different order under the default  behavior. Because LIMIT selects rows, this can change the returned values, not merely their presentation order.\n\nFor example,  may select a different branch in distributed execution.\n\nPlease propagate the effective order requirement into  (or conservatively force ordered fan-in whenever the parent pipeline/sink is order-dependent), and add a Ray regression test covering  with /.\n\nIf distributed execution intentionally does not preserve DuckDB's insertion-order contract, please document that behavior explicitly and add a test for the intended semantics.\n\n### Non-blocking follow-up\n\n currently reports the first exchange sink while the C++ binding path recursively applies a runtime handle. The current topology appears to use one sink per task, but an explicit invariant or multi-sink regression test would make that assumption safe.

external/duckdb/src/execution/distributed/pipeline_node/translator_union.cpp
```diff
@@ -0,0 +1,48 @@
...
+	auto output_names = GetSchemaNames(children.front()->config().schema());
+	SchemaRef schema;
+	if (output_names.size() == op.GetTypes().size()) {
+		schema = MakeSchemaRef(op.GetTypes(), output_names);
+	} else {
+		schema = MakeSchemaRef(op.GetTypes());
+	}
+
+	return std::make_shared<UnionNode>(get_next_pipeline_node_id(), plan_config_, children, std::move(schema),
+	                                   op.allow_out_of_order);
```
hubgeter: The translator forwards only op.allow_out_of_order. Upstream PhysicalUnion::BuildPipelines also computes order_matters from downstream pipeline and sink dependencies. Please carry that effective requirement into UnionNode; otherwise UNION ALL with LIMIT can select a different branch under preserve_insertion_order=true.

hubgeter (changes requested):
## Follow-up clarification\n\nThe earlier review body was incomplete because shell quoting stripped several identifiers. Please use this review as the authoritative version.\n\n## Review summary\n\nI reviewed commit 3ccad0a4f211a92b288d83a5fcc91405269fd1bb against base a31be62d45c27e03c564193c03ea4e5d385a173b.\n\nThe distributed UNION fan-in and scheduler-owned exchange identity changes are otherwise coherent. I found one correctness issue that should be fixed before merging.\n\n## Required change: preserve downstream order dependencies\n\nThe new UnionNode receives only PhysicalUnion::allow_out_of_order. Upstream PhysicalUnion::BuildPipelines also derives order_matters from downstream pipeline and sink order dependencies. On the distributed path, this information is lost, so a downstream order-sensitive LIMIT or OFFSET can consume UNION branches in a different order under the default preserve_insertion_order behavior. Because LIMIT selects rows, this can change the returned values, not merely their presentation order.\n\nFor example, SELECT * FROM (SELECT 3 AS x UNION ALL SELECT 1 AS x) LIMIT 1 may select a different branch in distributed execution.\n\nPlease propagate the effective order requirement into UnionNode, or conservatively force ordered fan-in whenever the parent pipeline or sink is order-dependent. Please also add a Ray regression test covering UNION ALL with LIMIT and OFFSET.\n\nIf distributed execution intentionally does not preserve DuckDB's insertion-order contract, please document that behavior explicitly and add a test for the intended semantics.\n\n### Non-blocking follow-up\n\nRayWorkerTask::ExchangeSinkConfig currently reports the first exchange sink while the C++ binding path recursively applies a runtime handle. The current topology appears to use one sink per task, but an explicit invariant or multi-sink regression test would make that assumption safe.

external/duckdb/src/execution/distributed/pipeline_node/translator_union.cpp
kaka11chen: Addressed in `89d425f99a`. The translator now performs an explicit top-down output-order requirement pass before its existing post-order node construction. Result order is propagated through insertion-order-preserving operators; order-sensitive sinks use `SinkOrderDependent`, batch-index requirements, and sink parallelism; `FIXED_ORDER` operators such as streaming LIMIT force ordered input; materialization barriers such as aggregate and ORDER BY reset the requirement according to their input contract. `UnionNode` therefore receives the effective `allow_out_of_order` value. Native coverage exercises root UNION, LIMIT, OFFSET, an order-insensitive aggregate, and `preserve_insertion_order=false`; Ray regressions cover both UNION ALL with LIMIT and OFFSET. The focused Ray E2E selection passed 8/8.

kaka11chen:
Also addressed the non-blocking multi-sink follow-up in `89d425f99a`. `TryGetUniqueRemoteExchangeSink` is now the shared inspection path used by `RayWorkerTask::ExchangeSinkConfig`, runtime task binding, and native worker execution. A worker plan with more than one remote exchange sink is rejected before any handle is applied; the new native regression verifies both sinks remain unbound. The exchange serialization/binding suite passed 181 assertions in 13 cases.

kaka11chen:
@codex review

hubgeter (approved):

PR #675, opened 2026-08-25
[fix](ai) Reject invalid text chunk sizes and overlaps

### LINKED ISSUE #70: [P2][Bug] AI text chunking accepts invalid size and overlap values

Mr-Neutr0n:
Fixes #70

Added parameter and weight-sum validation to `chunk_text` and `_weighted_average_embeddings` so invalid `max_chars`/overlap combinations and zero-weight aggregation raise clear `ValueError`s instead of producing empty chunks or NaN embeddings.

Could not run the suite locally: . Fork CI needs a maintainer approval to run, so this branch has no test signal yet.

---
This change was prepared with AI assistance under human direction and review.

### DIFF
--- a/vane/ai/functions.py
+++ b/vane/ai/functions.py
@@ -716,8 +716,16 @@ def chunk_text(
     Returns:
         List of text chunks. Returns ``[text]`` if text fits in one chunk.
     """
+    if isinstance(max_chars, bool) or not isinstance(max_chars, int) or max_chars <= 0:
+        raise ValueError("max_chars must be a positive integer")
+    if isinstance(overlap_chars, bool) or not isinstance(overlap_chars, int) or overlap_chars < 0:
+        raise ValueError("overlap_chars must be an integer >= 0")
+
     if len(text) <= max_chars:
         return [text]
+
+    if overlap_chars >= max_chars:
+        raise ValueError("overlap_chars must be smaller than max_chars")
 
     chunks: list[str] = []
     start = 0
@@ -738,7 +746,10 @@ def _weighted_average_embeddings(
     """Compute length-weighted average of embeddings."""
     arr = np.array(embeddings, dtype=np.float64)
     w = np.array(weights, dtype=np.float64)
-    w /= w.sum()
+    total = float(w.sum())
+    if not np.isfinite(total) or total <= 0:
+        raise ValueError("weights must sum to a positive finite value")
+    w /= total
     averaged = (arr * w[:, np.newaxis]).sum(axis=0)
     norm = np.linalg.norm(averaged)
     if norm > 0:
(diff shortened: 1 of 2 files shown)


kaka11chen:
@Mr-Neutr0n Thanks, please resolve the comments.

Mr-Neutr0n:
Fixed — moved the `overlap_chars < max_chars` validation before the short-text early return (8f4556a). Now `chunk_text("x", max_chars=1, overlap_chars=1)` correctly raises instead of returning `['x']`, so the invalid config is caught consistently regardless of input length.

vane/ai/functions.py
```diff
@@ -716,6 +716,13 @@ def chunk_text(
     Returns:
         List of text chunks. Returns ``[text]`` if text fits in one chunk.
     """
+    if isinstance(max_chars, bool) or not isinstance(max_chars, int) or max_chars <= 0:
+        raise ValueError("max_chars must be a positive integer")
+    if isinstance(overlap_chars, bool) or not isinstance(overlap_chars, int) or overlap_chars < 0:
+        raise ValueError("overlap_chars must be an integer >= 0")
+    if overlap_chars >= max_chars:
```
kaka11chen, line 723: **[P1] Update the existing `max_chars`-only cases for the new invariant**

Because `overlap_chars` defaults to 200, moving this check ahead of the early return makes both `test_chunk_text_short` and `test_chunk_text_exact_boundary` fail: they call `chunk_text(..., max_chars=100)` without an explicit overlap and now raise here. I reproduced 2 failures and 8 passes in `TestChunking`.

If rejecting this implicit combination is intentional, please update those two tests to pass an explicit valid overlap (for example, `0`) and add regression cases for invalid max/overlap values and non-positive or non-finite weight totals. Otherwise, adjust the default/API so a `max_chars`-only call remains valid.

kaka11chen (commented):
I reproduced a deterministic regression in the existing `TestChunking` tests: 2 of its 10 tests fail on this head. Please address the inline comment before merge.

The PR title check is also failing. This repository expects `[type](scope) Subject`; for example: `[fix](ai) Reject invalid text chunk sizes and overlaps`.

vane/ai/functions.py
Mr-Neutr0n: Updated in 2e32074f. The two existing max_chars-only tests now pass overlap_chars=0 explicitly, and the suite covers invalid max sizes, overlaps, and non-positive or non-finite weight totals. All 259 tests in test_ai_functions.py pass. I also corrected the PR title, and the new title check is green.

kaka11chen (approved):
LGTM

kaka11chen:
@Mr-Neutr0n thank you!

PR #679, opened 2026-08-25
[feat](datasink) Allow sink-defined delivery semantics

kaka11chen:
## Summary

- Remove the `CommitProtocol`, `RetryMode`, and `DataSinkCapabilities` gate so custom sinks define their own delivery semantics, including non-idempotent append behavior.
- Add `DataSinkExecutionOptions(max_retries=0)`. The default performs no retry; configured retries rerun the full input after non-cancellation `UNKNOWN` outcomes with one stable operation ID and an explicit duplicate-write warning.
- Keep cancellation- and query-deadline-derived `UNKNOWN` outcomes non-retryable, including interruption observed during Ray waits, execution finalization, or teardown.
- Reject retry-enabled relations backed by potentially single-use Arrow streams, including `RecordBatchReader`, bare stream capsules, and scanners; the default no-retry path remains permissive.
- Disable hidden Ray actor-task and FTE fragment replay inside each DataSink attempt while preserving the existing retry policy for ordinary UDFs.
- Document that Vane provides no exactly-once, rollback, or cross-sink transaction guarantee.

## Related issue

Follow-up to #652.

## Documentation impact

- [ ] No user-facing documentation update is required for this change.
- [x] Documentation is updated as part of this PR.
- [ ] A docs follow-up is required and tracked separately.

Public DataSink API and lifecycle docstrings now describe sink-defined semantics, default-off retries, cancellation behavior, replayable-input requirements, duplicate-write risk, and the absence of framework transaction guarantees.

## Validation

- Full changed-file pre-commit hooks, including Ruff, clang-format, and mypy
- Incremental release package build with `uv pip install . --no-build-isolation`
- `tests/fast/test_datasink.py`: 104 passed
- Affected DataSink, UDF lifecycle, and Ray result-contract suites: 727 passed
- `scripts/run_release_tests.sh`: 697 non-Ray tests and 14 real-Ray tests passed

## Checklist

- [x] I added or updated focused tests for the changed behavior.
- [x] Public behavior and intentional compatibility changes are documented and tested.
- [x] This change adds no new dependency or license requirement.
- [x] This change contains no credentials, secrets, or private endpoints.
- [x] Security-sensitive failure and retry behavior was reviewed.
- [x] Native changes were compiled and exercised by the Python release gates.


vane/datasink/__init__.py
```diff
@@ -278,23 +264,18 @@ def resolve(self) -> str:
...
+    """Execution requests for the internal Arrow operator.
+
+    ``max_retries`` is the number of full-operation retries after an UNKNOWN
+    outcome. It defaults to zero. Each retry re-executes the complete input
+    with the same operation ID, so the sink owns deduplication and all delivery
+    guarantees.
+    """
 
     worker_count: int = 1
+    max_retries: int = 0
```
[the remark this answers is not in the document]
kaka11chen: Addressed in 0dd9fc2228. `max_retries` is now declared after the existing fields and is keyword-only, so the prior positional order remains intact and a retry budget cannot be enabled accidentally through an extra positional argument. Added a regression test covering `DataSinkExecutionOptions(3, 20)` and rejection of a positional retry value.

kaka11chen:
@codex review

vane/datasink/__init__.py
```diff
@@ -1216,20 +1192,101 @@ def write_datasink(
...
+            try:
+                summary = _execute_datasink_once(
+                    prepared_relation,
+                    batch_actor,
+                    context,
+                    options,
+                    runner_type,
+                )
+            except DataSinkWriteError as error:
+                if error.outcome is WriteOutcome.UNKNOWN and retries_performed < options.max_retries:
```
hubgeter: **[P1] Do not retry an operation after driver cancellation**

`_execute_datasink_once()` catches every `BaseException` and converts `KeyboardInterrupt`, `SystemExit`, and cancellation-class control flow into `DataSinkWriteError(UNKNOWN)`. This condition then retries that error whenever budget remains. On the exact-head local-FTE artifact, a probe with `max_retries=1` that interrupted after the first external append ran two complete attempts and appended twice, so Ctrl-C can create additional non-idempotent side effects after the user asked the operation to stop. Preserve the `UNKNOWN` report, but distinguish cancellation-derived outcomes as non-retryable (or propagate a distinct terminal signal), and add a `max_retries > 0` interrupt regression that asserts only one attempt is started.

hubgeter (changes requested):
I found one merge-blocking correctness issue in the configured DataSink retry path. The PR correctly defaults `max_retries` to zero and suppresses hidden Ray actor-task and FTE fragment replay, but the new driver retry loop currently treats control-flow cancellation as an ordinary retryable `UNKNOWN` outcome. This can issue new non-idempotent external writes after the operator has explicitly cancelled the query.

Validated against exact head `0dd9fc2228` with its CI wheel: a local-FTE probe using `max_retries=1` and raising `KeyboardInterrupt` after the first external append executed two complete attempts and produced two external writes. The focused exact-head tests otherwise passed (96 passed, 365 deselected).

vane/datasink/__init__.py
kaka11chen: Addressed in 0d205887f4. Cancellation-derived UNKNOWN outcomes now carry a private driver-only terminal marker and never consume another framework retry. This covers interpreter control flow plus asyncio, future, and Vane execution cancellation while normalizing the escaped exception back to the public `DataSinkWriteError`. The local-FTE regression uses `max_retries=1`, waits until an external append is visible, triggers `KeyboardInterrupt`, and verifies exactly one attempt and one append. Added parameterized coverage for six cancellation/control-flow types. Validation: 94/94 DataSink tests, full changed-file hooks, and `scripts/run_release_tests.sh` (684 non-Ray + 14 real-Ray) passed. @hubgeter please re-review.

kaka11chen:
@codex review

vane/datasink/__init__.py
```diff
@@ -1212,24 +1214,120 @@ def write_datasink(
...
+                    context,
+                    options,
+                    runner_type,
+                )
+            except DataSinkWriteError as error:
+                interrupted = isinstance(error, _InterruptedDataSinkWriteError)
+                if (
+                    not interrupted
+                    and error.outcome is WriteOutcome.UNKNOWN
+                    and retries_performed < options.max_retries
```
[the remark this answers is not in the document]
kaka11chen: Addressed in `67ef617343`. Native and driver results now carry an explicit internal `outcome_cancelled` marker. The Ray driver observes cancellation across startup, fragment execution, result finalization, and teardown (including a cancellation racing with a cleanup diagnostic); an `UNKNOWN` result with that marker is converted to the existing terminal interruption path and cannot consume another framework retry. Known `APPLIED`/`ABORTED` results remain known. Added driver race tests plus a `max_retries=3` regression asserting exactly one attempt. Validation: 102/102 DataSink tests, focused Ray contracts, changed-file hooks, and the release gate (695 non-Ray + 14 real-Ray) passed.

vane/datasink/__init__.py
```diff
@@ -1212,24 +1214,120 @@ def write_datasink(
...
+    except BaseException as error:
+        raise TypeError("DataSink actor class must be cloudpickle-serializable before execution") from error
+    retries_performed = 0
+    last_unknown_error: DataSinkWriteError | None = None
+    try:
+        while True:
+            try:
+                summary = _execute_datasink_once(
+                    prepared_relation,
+                    batch_actor,
```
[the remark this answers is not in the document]
kaka11chen: Addressed in `67ef617343`. Before any retry-enabled write starts, Vane now binds the prepared relation inside a valid DuckDB transaction and recursively inspects the complete logical plan. Arrow dependencies that may be single-use (`RecordBatchReader`, bare stream capsules, and `Scanner`) are rejected without consuming the input; replayable `pyarrow.Table` input remains accepted. The default `max_retries=0` path stays permissive. Regressions cover direct readers, a reader hidden below a join, capsules, scanners, replayable tables, and default no-retry behavior. Validation: 102/102 DataSink tests, changed-file hooks, and the release gate (695 non-Ray + 14 real-Ray) passed.

kaka11chen:
@codex review

vane/datasink/__init__.py
```diff
@@ -1084,10 +1081,25 @@ def _unknown_error(
...
+    return _InterruptedDataSinkWriteError(
         _summary(context, WriteOutcome.UNKNOWN, results, warnings),
-        detail,
-        safe_to_retry=True,
+        _safe_error_summary(error),
+    )
+
+
+def _is_execution_interruption(error: BaseException) -> bool:
+    return not isinstance(error, Exception) or isinstance(
```
hubgeter: **[P1] Treat query-deadline expiry as a terminal interruption**

QueryDeadlineExceeded subclasses TimeoutError, so this predicate returns false. _execute_datasink_once() therefore wraps it as an ordinary UNKNOWN outcome, and the outer max_retries loop replays the complete input. On the Ray path, run_datasink_plan.remote() is submitted before progress.resolve() checks the deadline; with max_retries=3, an exact-wheel probe whose runner raises QueryDeadlineExceeded produced four calls. If the deadline expires during safe_get, _resolve_future() also rethrows a bare TimeoutError and loses the query-deadline marker.

This can submit additional non-idempotent external writes after query-wide cancellation. Please preserve QueryDeadlineExceeded when a deadline-limited wait expires, classify it as terminal and non-retryable here, and add a regression asserting only one attempt or submission when max_retries is greater than zero.

hubgeter (changes requested):
The previously reported cancellation propagation and single-use Arrow replay issues are addressed at this head, and the focused DataSink/Ray regressions pass.

One P1 remains: query-wide deadline expiry is still classified as a retryable UNKNOWN outcome, allowing full DataSink attempts to be resubmitted after cancellation. See the inline comment.

Validation performed against the current PR merge artifact c34a67809d:
- DataSink suite: 102 passed
- Ray cancellation and cleanup contracts: 4 passed
- Ray actor and FTE replay-isolation contracts: 8 passed

Please address the deadline retry path before merge.

vane/datasink/__init__.py
kaka11chen: Addressed in `dbb3c48d16`. Ray ObjectRef waits now retain whether the query deadline supplied the effective timeout and raise `QueryDeadlineExceeded` when that bound expires, both with and without progress callbacks. A completed task that itself raises `TimeoutError`, and a shorter configured ObjectRef timeout, remain ordinary timeouts. DataSink classifies direct and explicitly cause-wrapped query-deadline failures as terminal interruptions, so `max_retries=3` starts exactly one attempt even when teardown also fails. Regressions cover all of those paths. Validation: 104/104 DataSink tests, 727/727 affected tests, full changed-file hooks, and the release gate (697 non-Ray + 14 real-Ray) passed. @hubgeter please re-review.

kaka11chen:
@codex review

hubgeter (approved):

PR #685, opened 2026-08-26
[fix](ray) Serialize JSON scan plans

### LINKED ISSUE #632: [bug](ray) CSV and JSON scan plans cannot be serialized for Ray execution
### Affected area

Distributed execution or Ray

### What happened?

Relations backed by `read_csv`, `read_csv_auto`, or the `read_json*` / `read_ndjson*` table-function families cannot be executed through the Ray runner. Logical-plan serialization fails before distributed physical-plan translation begins.

The same files can be read with `VANE_RUNNER=local-fast`. File-backed relations using Parquet, `read_text`, and `read_blob` do not show this serialization failure.

### Reproduction

```python
import tempfile
from pathlib import Path

import ray
import vane
from vane import runners

with tempfile.TemporaryDirectory() as directory:
    root = Path(directory)
    csv_path = root / "input.csv"
    json_path = root / "input.json"
    csv_path.write_text("id,value\n1,a\n2,b\n", encoding="utf-8")
    json_path.write_text(
        '{"id":1,"value":"a"}\n{"id":2,"value":"b"}\n',
        encoding="utf-8",
    )

    ray.init(num_cpus=2)
    connection = vane.connect()
    try:
        runners.set_runner_ray(noop_if_initialized=True)
        runner = runners.get_or_create_runner()

        # Run each relation separately.
        relation = connection.sql(f"SELECT * FROM read_csv_auto('{csv_path}')")
        print(list(runner.run_iter_tables(relation)))

        # relation = connection.sql(f"SELECT * FROM read_json_auto('{json_path}')")
        # print(list(runner.run_iter_tables(relation)))
    finally:
        connection.close()
        ray.shutdown()
```

Observed core errors:

```text
CSVReaderSerialize not implemented
JSONScan Serialize not implemented
```

### Environment

```text
OS: Linux 6.17.0-29-generic, glibc 2.39
Architecture: x86_64
Python: 3.12.3
Vane: 0.2.0.dev564; upstream/main inspected at 08152e15b921d2fb62f2efff64696df3ad5e13e4
Engine: v1.5.5-vane.9d91c17ba6
Ray: 2.57.0, local cluster with 2 CPUs
VANE_RUNNER: ray
```

### Additional context

The failure is independent of the number of input files: both single-file and multi-file CSV/JSON relations fail while serializing the logical plan. The relevant CSV and JSON serialization callbacks are unchanged between the tested engine source and the inspected `upstream/main` commit.

kaka11chen:
## Summary

- Serialize the complete bound state for JSON and NDJSON scans so their logical plans can be translated and executed by Ray.
- Reconstruct worker-side scan state without reopening, rebinding, or resampling input files, while preserving multi-file metadata and file options.
- Preserve each coordinator-side file ordinal through JSON serde and singleton Ray split execution, including `file_index` projections and filters.
- Add coverage for every JSON/NDJSON table-function alias and for real-Ray multi-file execution with schema union, filenames, explicit date/timestamp formats, and stable file indexes.

## Related issue

close: #632

## Documentation impact

      `AstroVela/vane-website`.

The change enables an existing execution path and does not change user-facing syntax or configuration.

## Validation

- Two consecutive clean code-review rounds before the initial build, plus two consecutive clean rounds after addressing review feedback
- `scripts/format workspace --changed --check`
- `git diff --check upstream/main`
- JSON serialization generator determinism check
- `SKBUILD_BUILD_DIR="$PWD/build/python-release" SKBUILD_CMAKE_BUILD_TYPE=Release CMAKE_BUILD_PARALLEL_LEVEL=12 uv pip install --python .venv/bin/python . --no-build-isolation`
- `scripts/run_installed_pytest.sh tests/fast/test_ray_e2e_serialization.py::test_json_scan_family_plans_serialize_for_ray`
- `scripts/run_installed_pytest.sh tests/fast/test_ray_e2e_serialization.py::test_multi_file_json_bind_state_survives_real_ray_serde tests/fast/test_ray_e2e_serialization.py::test_multi_file_json_preserves_file_index_through_real_ray`
- `scripts/run_release_tests.sh` (`692 passed, 5 skipped, 18 deselected`; real Ray: `14 passed, 701 deselected`)

The release suite used a short symlink to the same worktree and installed native artifact. The full worktree path made three unrelated negative-path tests exceed the native 4096-byte exception-detail cap; those tests and the complete suite pass with the shorter path.

## Checklist

### DIFF
--- a/external/duckdb/extension/json/json_scan.cpp
+++ b/external/duckdb/extension/json/json_scan.cpp
@@ -2,6 +2,7 @@
 
 #include "duckdb/common/enum_util.hpp"
 #include "duckdb/common/multi_file/multi_file_reader.hpp"
+#include "duckdb/common/set.hpp"
 #include "duckdb/common/serializer/deserializer.hpp"
 #include "duckdb/common/serializer/serializer.hpp"
 #include "duckdb/main/extension_helper.hpp"
@@ -10,6 +11,25 @@
 #include "json_multi_file_info.hpp"
 
 namespace duckdb {
+
+JSONFileSnapshot::JSONFileSnapshot(const OpenFileInfo &file) : path(file.path) {
+	if (file.extended_info) {
+		for (const auto &entry : file.extended_info->options) {
+			options.emplace(entry.first, entry.second);
+		}
+	}
+}
+
+OpenFileInfo JSONFileSnapshot::ToOpenFileInfo() const {
+	OpenFileInfo result(path);
+	if (!options.empty()) {
+		result.extended_info = make_shared_ptr<ExtendedOpenFileInfo>();
+		for (const auto &entry : options) {
+			result.extended_info->options.emplace(entry.first, entry.second);
+		}
+	}
+	return result;
+}
 
 JSONScanData::JSONScanData() {
 }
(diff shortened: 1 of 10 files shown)


kaka11chen:
@codex review

external/duckdb/extension/json/json_scan.cpp
```diff
@@ -104,12 +168,220 @@ void JSONScanLocalState::AddTransformError(idx_t object_index, const string &err
...
+	auto &json_data = bind_data.bind_data->Cast<JSONScanData>();
+
+	SerializedJSONScanData serialized_data;
+	auto files = bind_data.file_list->GetAllFiles();
+	serialized_data.files.reserve(files.size());
+	for (const auto &file : files) {
+		if (file.path.empty()) {
+			throw SerializationException("Cannot serialize %s with an empty JSON file path", function.name);
+		}
+		serialized_data.files.emplace_back(file);
```
hubgeter: [P1] Preserve the coordinator file ordinal during distributed execution

This snapshot keeps only the path and extended options. When Ray materializes a per-file split, `ApplyScanSplitBatches` rebuilds the worker file list from that one path, and `MultiFileFunction` derives `file_list_idx` from the worker-local index. With one split per batch, every file therefore reports `file_index = 0`.

I reproduced this at `7be31e03b88b4116acd7da0f31c3cf2236da34ce`: two local files return `(1, 0), (2, 1)` locally but `(1, 0), (2, 0)` through Ray; `WHERE file_index = 1` returns the second row locally and no rows through Ray. This is silent wrong-result/data-loss behavior.

Please carry a stable coordinator ordinal through the serialized JSON snapshot/split into `JSONReader::file_list_idx` (the CSV path has an ordinal mechanism), and add a real-Ray regression that both projects and filters `file_index` with one split per batch.

hubgeter (changes requested):
Thanks for adding serialized bind state for the JSON scan family. I verified the existing focused tests and exercised all seven JSON/NDJSON aliases on the exact head. However, distributed multi-file execution still loses the coordinator's file ordinal, so the public `file_index` virtual column can return silently incorrect results. I have left the blocking details inline. Please preserve the original ordinal end to end and add a real-Ray regression that projects and filters `file_index` with one split per batch.

external/duckdb/extension/json/json_scan.cpp
kaka11chen: Addressed in a0e149bba2. JSONFileSnapshot now owns and serializes the stable coordinator ordinal, rehydrates it into private OpenFileInfo metadata that travels with each file split, and JSONReader restores file_list_idx before MultiFileFunction can assign a worker-local index. The union reader path now also retains the complete OpenFileInfo.

Added a real-Ray regression with one split per batch that verifies both projection, yielding (1, 0) and (2, 1), and filtering on file_index = 1, yielding only (2, 1). The focused tests and full release suite pass.

kaka11chen:
@codex review

hubgeter (approved):

PR #686, opened 2026-08-26
[feat](qdrant) Add a keyed-upsert DataSink

### LINKED ISSUE #648: [feat](qdrant) Add a keyed-upsert DataSink
### Problem

Vane has no supported way to write distributed relation batches directly to Qdrant. Calling `qdrant-client` from a generic batch callback leaves point-ID generation, retry safety, acknowledgement, bounded task results, and credential handling outside Vane's write contract.

### Proposed outcome

After #646 is available, add a Qdrant adapter with replay-safe point upserts:

- Add `qdrant-client` as an optional dependency/extra with no base-package import requirement.
- Provide a `QdrantSink` built on the generic keyed-upsert DataSink helper.
- Require an explicit point-ID column using Qdrant-supported unsigned integer or UUID IDs. Reject client-generated/random IDs, null IDs, and duplicate IDs before distributed side effects begin.
- Support configured single-vector and named-vector mappings plus payload-field mappings, with Arrow type, vector dimension, and collection-schema validation.
- Use Qdrant point upsert as an immediate idempotent write and wait for application acknowledgement (`wait=True`) before reporting a batch as applied.
- Make full point-replacement behavior explicit and document that applied batches may remain visible if a later batch or the Vane job fails.
- Enforce configurable batch row/byte limits and return only bounded counts, provider status, and warnings through the generic result contract.
- Resolve URL/API-key credentials from worker environment or opaque secret references without serializing secret values.
- Add SDK-mocked tests for ID validation, vector/payload mapping, batching, replay, acknowledgement failure, result bounds, and redaction. Keep live Qdrant tests behind an external-service marker.
- Document installation and a representative relation-to-collection example.

- Generate UUIDs inside each worker. This is convenient but retrying an attempt would create different point IDs and duplicate data.
- Set `wait=False` and treat receipt as success. This improves latency but weakens the meaning of an applied `WriteResult`.
- Collect and upload from the driver. This avoids a distributed connector but sacrifices scalability and creates a driver memory bottleneck.

### Additional context

- Blocked by https://github.com/AstroVela/vane/issues/646.
- Qdrant documents point-loading APIs as idempotent: re-uploading a point with the same ID overwrites it. Explicit stable IDs are therefore the adapter's replay boundary: https://qdrant.tech/documentation/concepts/points/
- This issue does not provide cross-batch transactions, exactly-once guarantees, point deletion, or partial-update operations.

kaka11chen:
## Summary

- Add a public QdrantSink built on BoundKeyedUpsertSink with explicit Arrow uint64 or normalized UUID point IDs and global null/duplicate rejection before workers open.
- Support one unnamed dense vector or complete named-vector mappings, explicit JSON-safe payload mappings, collection/dimension validation, bounded batches/results, full-point replacement, and wait=True COMPLETED acknowledgement.
- Resolve endpoint/API-key EnvironmentSecret references only on workers, keep qdrant-client optional, and avoid compatibility or fallback paths.
- Add SDK-mocked coverage, documentation, packaging metadata, and a digest-pinned Qdrant v1.19.0 external-service CI job that verifies real full-point replacement.

qdrant-client is Apache-2.0 licensed and remains an optional, non-vendored dependency.

## Related issue

close: #648

## Documentation impact

      AstroVela/vane-website.

README documents installation, mappings, supported Arrow values, credentials, batching, replay, replacement, and partial-visibility behavior.

## Validation

- Two consecutive clean code-review rounds after the final change.
- scripts/format root --changed --check
- pre-commit run on all changed files (YAML/TOML, Ruff, workflow schema, mypy, shfmt)
- SKBUILD_BUILD_DIR=$PWD/build/python-release SKBUILD_CMAKE_BUILD_TYPE=Release uv pip install . --no-build-isolation
- scripts/run_installed_pytest.sh tests/fast/test_qdrant_datasink.py: 52 passed, 1 deselected
- scripts/run_release_tests.sh: 810 passed / 20 deselected, then 14 passed / 816 deselected
- Real Qdrant v1.19.0 (official Linux binary, build 74f3e85b) with qdrant-client 1.19.0: external-service live test passed; verified acknowledged upsert, raw vector mapping, full payload replacement, and cleanup

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -498,6 +498,84 @@ jobs:
         continue-on-error: true
         run: docker rm --force vane-milvus
 
+  qdrant-tests:
+    name: Qdrant integration test
+    needs:
+      - changes
+      - native-fast-tests
+    if: needs.changes.outputs.run_full == 'true'
+    runs-on: ubuntu-24.04
+    timeout-minutes: 15
+    env:
+      PIP_DISABLE_PIP_VERSION_CHECK: "1"
+      VANE_TEST_QDRANT_URL: http://127.0.0.1:6333
+    steps:
+      - name: Checkout
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+
+      - name: Set up Python
+        uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
+        with:
+          python-version: "3.12"
+
+      - name: Download the built wheel
+        uses: actions/download-artifact@37930b1c2abaa49bbe596cd826c3c89aef350131 # v7.0.0
+        with:
+          name: ci-python-artifacts-3.12
+          path: dist
+
+      - name: Install the wheel and Qdrant test dependencies
+        run: |
+          python -m pip install --upgrade pip
+          wheel_path="$(find dist -maxdepth 1 -type f -name '*.whl' -print -quit)"
+          test -n "$wheel_path"
+          python -m pip install --force-reinstall "${wheel_path}[qdrant]" pytest
+          python -m pip check
+          python -c 'import qdrant_client'
+
+      - name: Start Qdrant
+        run: |
+          docker run --detach \
+            --name vane-qdrant \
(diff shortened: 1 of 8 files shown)


kaka11chen:
@codex review

kaka11chen:
@codex review

hubgeter (approved):

Issue #713, opened 2026-08-30
[AudioFile] Metadata, decoding, and resampling execution


kaka11chen:
### Problem

Encoded audio references need a consistent metadata, waveform, and resampling API with defined sample layout, NULL behavior, resource limits, and errors.

The accepted stage tracked here uses an explicit decoded-audio STRUCT for expression/SQL results. It does not use the variable-shape Tensor result described in the original broader proposal.

### Proposed outcome

AudioFile metadata is available independently of full waveform decoding. Python waveform and resampling calls return detached float64 arrays with shape `(frames, channels)`, including `(frames, 1)` for mono audio.

The accepted `audio_resample()` expression/SQL result is:

```text
STRUCT(samples DOUBLE[], sample_rate BIGINT, frames BIGINT, channels BIGINT)
```

Samples are frame-major. NULL behavior, strict FILE ranges, resource limits, and media versus system-error behavior are defined for these calls.

### Additional context

This issue remains closed for its accepted scope:

- [x] AUDIOFILE logical-type foundation: #717.
- [x] Metadata and bounded Python waveform decoding: #734.
- [x] `AudioFile.resample()` and `audio_resample()` with the accepted STRUCT result: #747.

The current fixed-shape Tensor limitation was part of the original scope decision. The difference from the broader `resample -> Tensor` proposal is tracked separately in #755.

Parent: #715. Cross-cutting issues: #750/#752 for Flight shuffle and #661 for governed credentials.

[Current expression/SQL contract test](https://github.com/AstroVela/vane/blob/33da8668e3fffa35a39cb074448577e616b63f8a/tests/fast/test_audio_file.py#L198)

PR #734, opened 2026-09-01
[feat](file) Add AudioFile metadata and decoding

kaka11chen:
## Summary

- Add the native SQL audio_metadata(AUDIOFILE [, max_bytes]) scalar and Python expression facade.
- Add AudioMetadata, AudioFile.metadata(), and bounded, range-aware AudioFile.to_numpy() decoding to detached float64 frame-major arrays.
- Use SoundFile/libsndfile for encoded-audio parsing while preserving Vane FILE resolution, cancellation, resource limits, MIME validation, and error contracts.
- Add the optional audio dependency extra and focused SQL/Python regression coverage.

## Related issue

Refs #713. This PR implements the metadata and bounded Python decode subset; resampling and the engine decoded-audio/Arrow/distributed contract remain follow-up work.

## Documentation impact

- [ ]  A follow-up issue is required in AstroVela/vane-website.

The public contract is captured in docstrings, type stubs, and tests. README is intentionally unchanged for this focused subset.

## Validation

- scripts/format root --changed
- Ruff check on the changed Python files
- Incremental non-editable Release install with CMAKE_BUILD_PARALLEL_LEVEL=12
- Installed-package targeted suite: 73 passed across tests/fast/test_audio_file.py, the media value contract, and audio package metadata checks
- git diff --check

## Checklist

### DIFF
--- a/vane/_audio_file.py
+++ b/vane/_audio_file.py
@@ -571,7 +571,7 @@ def _decode_audio_file(
     soundfile = _load_soundfile()
     numpy = importlib.import_module("numpy")
 
-    class _AudioDecoder(soundfile.SoundFile):
+    class _AudioDecoder(soundfile.SoundFile):  # type: ignore[name-defined]
         def seekable(self) -> bool:
             # libsndfile reports SF_COUNT_MAX when a container omits its
             # total frame count. The encoded stream can still seek while
--- a/vane/_audio_file.py
+++ b/vane/_audio_file.py
@@ -600,65 +600,76 @@ def _decode_audio_file(
                     decoded_frame_limit = normalized_max_decoded // frame_bytes
                     frame_limit = min(normalized_max_frames, decoded_frame_limit)
                     if frame_limit == 0:
-                        raise AudioFileLimitError(
-                            f"one decoded audio frame requires {frame_bytes} bytes, "
-                            f"exceeding max_decoded_bytes={normalized_max_decoded}"
-                        )
+                        # A stream with an unknown total can still be empty. A
+                        # one-frame probe distinguishes EOF from a non-empty
+                        # result that cannot fit the requested output budget.
+                        probe = bytearray(frame_bytes)
+                        probe_frames = audio.buffer_read_into(probe, dtype="float64")
+                        proxy.raise_if_error()
+                        del probe
+                        if probe_frames < 0 or probe_frames > 1:
+                            raise AudioFileFormatError("audio decoder returned an invalid frame count")
+                        if probe_frames:
+                            raise AudioFileLimitError(
+                                f"one decoded audio frame requires {frame_bytes} bytes, "
+                                f"exceeding max_decoded_bytes={normalized_max_decoded}"
+                            )
+                        samples = numpy.empty((0, metadata.channels), dtype=numpy.float64)
+                    else:
+                        # Unknown-length streams cannot accumulate both a decoded
+                        # output and a same-sized scratch buffer without violating
+                        # max_decoded_bytes. Spool bounded chunks, release the
+                        # scratch allocation, then materialize the exact ndarray
+                        # directly from the temporary file.
+                        with tempfile.TemporaryFile(mode="w+b", buffering=0, prefix="vane_audio_") as decoded_file:
+                            decoded_frames = 0
+                            while decoded_frames < frame_limit:
+                                requested_frames = min(64 * 1024, frame_limit - decoded_frames)
+                                chunk = bytearray(requested_frames * frame_bytes)
+                                chunk_frames = audio.buffer_read_into(chunk, dtype="float64")
+                                proxy.raise_if_error()
+                                if chunk_frames < 0 or chunk_frames > requested_frames:
+                                    raise AudioFileFormatError("audio decoder returned an invalid frame count")
+                                chunk_bytes = chunk_frames * frame_bytes
+                                chunk_view = memoryview(chunk)[:chunk_bytes]
+                                while chunk_view:
+                                    written = decoded_file.write(chunk_view)
+                                    if written is None or written <= 0:
+                                        raise OSError("temporary audio spool made no write progress")
+                                    chunk_view = chunk_view[written:]
+                                decoded_frames += chunk_frames
+                                reached_end = chunk_frames < requested_frames
+                                del chunk_view
+                                del chunk
+                                if reached_end:
+                                    break
 
-                    # Unknown-length streams cannot accumulate both a decoded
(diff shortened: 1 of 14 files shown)


kaka11chen:
@codex review

kaka11chen:
@codex review

vane/_audio_file.py
```diff
@@ -0,0 +1,734 @@
...
+            stream.raise_if_error()
+            metadata = _metadata_from_sound_file(audio, content_type=content_type)
+            stream.raise_if_error()
+    except BaseException as error:
+        # Cancellation and interpreter-control exceptions must never be
+        # rewritten as a media or resource-limit failure.
+        if not isinstance(error, Exception):
+            raise
+        stream.raise_if_error()
+        if stream.budget_exhausted:
```
hubgeter, line 518: P2: These budget flags are checked before the current exception type, so a real decoder-side resource or internal failure is rewritten as a caller-supplied limit error. In a focused exact-wheel probe, a decoder that exhausts a 1-byte metadata budget and then raises `MemoryError` produces `AudioFileLimitError(...max_bytes=1)`, with the `MemoryError` retained only as its cause; the SQL bridge then maps it to `InvalidInputException`. That violates #713's contract that resource and internal failures propagate. Please propagate `MemoryError`, existing `AudioFileError`, and other non-SoundFile/internal exceptions before applying the budget/fetch translation, and only translate the SoundFile parse failure caused by the truncated virtual stream.

vane/_audio_file.py
```diff
@@ -0,0 +1,734 @@
...
+                        if decoded_frames != metadata.frames:
+                            raise AudioFileFormatError(
+                                f"audio decoder returned {decoded_frames} frames after reporting {metadata.frames}"
+                            )
+                        samples = numpy.frombuffer(decoded, dtype=numpy.float64).reshape(
+                            metadata.frames,
+                            metadata.channels,
+                        )
+                    extra = bytearray(frame_bytes)
+                    extra_frames = audio.buffer_read_into(extra, dtype="float64")
```
hubgeter: P2: This extra read cannot detect an underreported finite FLAC frame count with the supported SoundFile/libsndfile stack. libsndfile treats STREAMINFO's total-sample count as the decode boundary: for the added 64-frame fixture whose header reports 63, the first read returns 63 and this read returns 0, so `to_numpy()` returns a silently truncated array instead of raising. I reproduced this with the exact Python 3.12 CI wheel and SoundFile 0.14.0, and the same regression now fails in the full fast-test job. Please either validate container integrity independently or define decoder-reported frames as authoritative and remove the unsupported rejection contract/test.

hubgeter (changes requested):
Requesting changes for two reproducible correctness/error-contract issues: the finite-length decode path accepts and truncates an underreported FLAC stream, and metadata probing can rewrite decoder-side resource/internal failures as caller-supplied limit errors.

I validated head `79dd15e216d6c800d9ece9cbe9967849f909bf9f` with the exact Python 3.12 CI wheel and SoundFile 0.14.0. The targeted audio suite produced `65 passed, 1 failed`, and a focused fake-decoder probe reproduced the exception masking. The full fast-test shard now also confirms the FLAC regression; it additionally reports that the AudioFile public-member contract assertion needs to be updated before merge.

One tracking issue also needs correction: this PR uses `close: #713`, but resampling and the engine decoded-audio/Arrow/distributed contract from that issue remain unimplemented. Please use a non-closing reference or split the remaining acceptance criteria before closing the issue.

vane/_audio_file.py
kaka11chen: Fixed in 7beeee39ad. The probe now re-raises existing AudioFileError values and every non-SoundFile exception before considering budget/fetch flags. Only a SoundFile parse failure produced against the truncated virtual view is translated to AudioFileLimitError. Added focused regressions for MemoryError, an internal RuntimeError, and an existing classified media error after budget exhaustion; all preserve the original exception.

vane/_audio_file.py
kaka11chen: Fixed in 7beeee39ad by choosing the decoder-authoritative contract. The finite-length path no longer performs or promises an independent trailing-frame integrity check, and the unsupported underreported-FLAC rejection test was removed. Unknown-length streams still decode incrementally under max_frames and max_decoded_bytes limits.

kaka11chen:
@codex review

vane/_audio_file.py
```diff
@@ -0,0 +1,724 @@
...
+                                raise OSError("temporary audio spool ended before the decoded output")
+                            output_offset += read_count
+                        del output_view
+                else:
+                    frame_bytes = metadata.channels * 8
+                    if frame_bytes > normalized_max_decoded:
+                        raise AudioFileLimitError(
+                            f"one decoded audio frame requires {frame_bytes} bytes, "
+                            f"exceeding max_decoded_bytes={normalized_max_decoded}"
+                        )
```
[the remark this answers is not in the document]
kaka11chen: Fixed in 49594cf647. The finite-length path now checks only the total decoded byte count, which is zero for an empty stream and already subsumes the one-frame check for every non-empty stream. Added mono and stereo empty-WAV regressions with max_decoded_bytes=1; the full AudioFile plus media-value contract suite passes 73/73.

kaka11chen:
@codex review

kaka11chen:
@codex review

vane/_audio_file.py
```diff
@@ -0,0 +1,718 @@
...
+                    )
+                if metadata.frames is None:
+                    frame_bytes = metadata.channels * 8
+                    decoded_frame_limit = normalized_max_decoded // frame_bytes
+                    frame_limit = min(normalized_max_frames, decoded_frame_limit)
+                    if frame_limit == 0:
+                        raise AudioFileLimitError(
+                            f"one decoded audio frame requires {frame_bytes} bytes, "
+                            f"exceeding max_decoded_bytes={normalized_max_decoded}"
+                        )
```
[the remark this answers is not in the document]
kaka11chen: Fixed in 207ae12995. Unknown-length decoding now performs a bounded one-frame EOF probe when the float64 output budget cannot hold one frame: an empty stream returns a detached `(0, channels)` array, while any decoded frame still raises `AudioFileLimitError`. Added mono/stereo empty-stream regressions plus a non-empty limit regression; the complete AudioFile and File value-contract selection passes 76/76.

kaka11chen:
@codex review

kaka11chen:
@codex review

hubgeter (approved):

Issue #714, opened 2026-08-30
[VideoFile] Indexed frame seeking and observable selection cost


kaka11chen:
### Problem

At upstream/main `a322f737e8c0816bc1d0d49ba9e73afeec0194eb`, #766 and #767 provide native video frame sources and native scalar/list frame expressions. Time-window and frame-index selection still decode sequentially from the beginning, so a late interval or high frame index includes preceding decode work.

### Proposed outcome

- Explicit native indexed/keyframe selection reduces unnecessary reading and decoding for repeated selective queries. Initial index construction costs and subsequent query costs are separately observable and reproducible.
- Exact frame indices, source identity, presentation timestamps and frame metadata remain reproducible for supported variable-frame-rate, B-frame and nonzero-origin inputs. Frame indices are not approximated from average FPS.
- Scalar/list expressions and streaming frame sources expose the same selection semantics, including local and Ray execution.
- Reads stay inside the FILE logical byte window. Invalid or stale indexes, unavailable selective access, cancellation, I/O, permissions and resource failures remain distinguishable from ordinary nullable media errors.
- Python execution remains available through explicit backend selection. Compatibility layers and automatic backend or temporary-file materialization fallback are outside the accepted scope; unsupported indexed access reports an error.

### Alternatives and tradeoffs

Explicit sequential decoding remains useful for one-off requests, short clips and dense selections. Index creation requires initial work and storage; selective-access benefits should be measured separately from that cost.

### Additional context

Scope updated after #766 and #767 to reflect the accepted requirement to retain the Python backend without compatibility or automatic fallback.

Delivered stages: media FILE types (#717), video metadata (#736), Python frame iterators and provenance (#742), exact sequential frame-index access (#744), FILE-aware Worker sources (#746), native streaming frames (#766), and native frame expressions (#767).

Parent: #715. Native extension work: #757. Performance evidence: #764. Broader multi-provider/multi-scope credential governance remains tracked separately in #661.

PR #736, opened 2026-09-01
[feat](file) Add VideoFile metadata

kaka11chen:
## Summary

- Add native SQL `video_metadata(VIDEOFILE [, max_bytes])` and the Python expression facade.
- Add `VideoMetadata` and `VideoFile.metadata()`, backed by bounded, range-aware PyAV/FFmpeg inspection of the first non-attached video stream.
- Preserve exact rational time bases, nullable decoder-reported fields, FILE byte-range semantics, cancellation, and error classification.
- Reject nested external resources, redact their URLs, tolerate unrelated undecodable metadata tags, and exclude attached cover art from video selection.

## Related issue

Refs #714

This PR implements the metadata-discovery slice only. Selective frame decoding, keyframes, frame-index access, streaming sources, and distributed execution remain follow-up work.

## Documentation impact

      `AstroVela/vane-website`.

The additive public contract is recorded in docstrings, type stubs, and tests. README changes are intentionally out of scope.

## Validation

- `/home/kaka/.local/bin/ruff check vane/_video_file.py tests/fast/test_video_file.py`
- Incremental non-editable Release install with `CMAKE_BUILD_PARALLEL_LEVEL=12`
- `scripts/run_installed_pytest.sh -q tests/fast/test_video_file.py` — 41 passed
- `scripts/run_installed_pytest.sh -q tests/fast/test_file.py tests/fast/test_image_file.py tests/fast/test_audio_file.py tests/fast/test_package_metadata.py` — 229 passed, 17 skipped because optional fsspec/ADBC packages are absent
- `scripts/run_release_tests.sh` — passed, including 14 real-Ray tests
- Real FFmpeg sample containing audio plus an attached cover-art video stream — correctly rejected as having no video payload stream
- `git diff --check`

## Checklist


kaka11chen:
@codex review

kaka11chen:
@codex review

vane/_video_file.py
```diff
@@ -0,0 +1,529 @@
...
+    """Prevent a container manifest from bypassing the FILE resolver."""
+
+    def __init__(self) -> None:
+        self.error: VideoFileFormatError | None = None
+
+    def __call__(self, url: str, flags: int, options: dict[str, str]) -> Any:
+        # Do not echo a nested URL because manifests can contain signed query
+        # parameters or embedded credentials.
+        del url, flags, options
+        self.error = VideoFileFormatError("video metadata requires an external resource")
```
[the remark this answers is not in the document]
kaka11chen: Fixed in 3cbc8cb0bd. The error now says nested external resources are not permitted, and the regression test verifies that a signed nested URL is not echoed.

vane/_video_file.py
```diff
@@ -0,0 +1,529 @@
...
+        return None
+    frame_count = int(video.frames)
+    if frame_count < 0 or frame_count > _MAX_BIGINT:
+        raise VideoFileFormatError("video parser reported an out-of-range frame count")
+    # FFmpeg uses zero when the container does not carry an exact frame count.
+    return frame_count or None
+
+
+def _metadata_from_container(container: Any, content_type: str | None, av_module: Any) -> VideoMetadata:
+    video = next((stream for stream in container.streams if stream.type == "video"), None)
```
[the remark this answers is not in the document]
kaka11chen: Fixed in fd0e925e68. Stream selection now excludes PyAV's Disposition.attached_pic flag before choosing the first video stream. The tests cover both cover-before-video selection and an attachment-only container; I also verified a real FFmpeg-generated audio file with embedded cover art is rejected as non-video.

vane/_video_file.py
```diff
@@ -0,0 +1,529 @@
...
+) -> tuple[int, int, float | None, float | None, int | None, int, int]:
+    """Bounded PyAV helper called by the native SQL scalar function."""
+    av_module = _load_av()
+    stream = _VideoMetadataView(read_at, logical_size=logical_size, max_bytes=max_bytes)
+    nested_io = _NestedIOBlocker()
+    try:
+        container = av_module.open(
+            stream,
+            mode="r",
+            metadata_encoding="utf-8",
```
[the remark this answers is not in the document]
kaka11chen: Fixed in fd0e925e68. The PyAV open call now uses metadata_errors="replace" with UTF-8, and a regression test asserts the probing option. This API is present across the supported PyAV 15-17 range.

kaka11chen:
@codex review

kaka11chen:
@codex review

vane/_video_file.py
```diff
@@ -0,0 +1,535 @@
...
+    logical_size: int,
+    content_type: str | None,
+    max_bytes: int,
+) -> tuple[int, int, float | None, float | None, int | None, int, int]:
+    """Bounded PyAV helper called by the native SQL scalar function."""
+    av_module = _load_av()
+    stream = _VideoMetadataView(read_at, logical_size=logical_size, max_bytes=max_bytes)
+    nested_io = _NestedIOBlocker()
+    try:
+        container = av_module.open(
```
hubgeter, line 487: **P1: Bound decoder work before opening untrusted media**

`av.open()` is not a header-only operation. PyAV 17.1.0 [unconditionally calls `avformat_find_stream_info()`](https://github.com/PyAV-Org/PyAV/blob/v17.1.0/av/container/input.py#L61-L65) while constructing an input container, and FFmpeg documents that this function [tries to read and decode frames](https://github.com/FFmpeg/FFmpeg/blob/n8.0/libavformat/avformat.h#L112-L117). Consequently, `max_bytes` only caps compressed reads from `_VideoMetadataView`; it does not cap decoded-frame allocation or CPU. The dimension check at line 418 runs after this work, while FFmpeg's default [`max_pixels` is `INT_MAX`](https://github.com/FFmpeg/FFmpeg/blob/n8.0/libavcodec/options_table.h#L388-L391). A compact crafted `VIDEOFILE` can therefore make this SQL scalar allocate/decode a very large frame before Vane rejects its dimensions, and query cancellation may not be observed while FFmpeg works from already-buffered packets. This also contradicts the public `no frame decoding` contract at line 521.

Please enforce decoder/probe resource limits and cancellation before stream-info discovery, or use a genuinely header-only inspection path. A subprocess regression with memory and timeout limits would cover the process-level failure mode.

vane/_video_file.py
```diff
@@ -0,0 +1,535 @@
...
+    if frame_count < 0 or frame_count > _MAX_BIGINT:
+        raise VideoFileFormatError("video parser reported an out-of-range frame count")
+    # FFmpeg uses zero when the container does not carry an exact frame count.
+    return frame_count or None
+
+
+def _metadata_from_container(container: Any, content_type: str | None, av_module: Any) -> VideoMetadata:
+    attached_pic = av_module.stream.Disposition.attached_pic
+    video = next(
+        (stream for stream in container.streams if stream.type == "video" and not (stream.disposition & attached_pic)),
```
hubgeter, line 430: **P2: Reject standalone image demuxers instead of treating them as videos**

FFmpeg exposes standalone PNG/JPEG/GIF inputs as a `video` stream whose disposition is not `attached_pic`, so this predicate accepts them. `_validate_content_type()` then also accepts an unmapped image demuxer because its `compatible` set is empty. Against the CI wheel, a 1x1 PNG succeeds with `content_type=None`, `video/*`, and even forged `video/mp4`, returning `width=1`, `height=1`, `fps=25`, and `time_base=1/25` from both Python and SQL. This crosses the `IMAGEFILE`/`VIDEOFILE` contract and reports parser defaults as video metadata.

Please require positive evidence of a supported video container (or explicitly reject still-image demuxers/codecs), and add PNG/JPEG/GIF regressions with absent, generic, and forged video MIME types.

vane/_video_file.py
```diff
@@ -0,0 +1,535 @@
...
+        None,
+    )
+    if video is None:
+        raise VideoFileFormatError("logical FILE view does not contain a video stream")
+
+    width = int(video.width)
```
hubgeter, line 441: **P2: Classify an unavailable decoder as an input format error**

A stream with `type == "video"` does not guarantee that PyAV found a decoder. For example, a valid Matroska container with an unknown codec ID produces a video stream with `codec_context=None`; this attribute access then raises `AttributeError: 'VideoStream' object has no attribute 'width'`. `_probe_video_metadata()` only translates `FFmpegError`, so the Python API leaks `AttributeError` and the native wrapper turns the same user-controlled file into DuckDB `InternalException` rather than `VideoFileFormatError` / `InvalidInputException`.

Please detect the missing decoder/codec context (and wrap metadata extraction failures that represent unsupported input) before reading these attributes, then add direct-Python and SQL regressions for an unsupported codec ID.

hubgeter (changes requested):
I found one high-severity resource-boundary issue and two input-contract/error-classification issues on `fd0e925e68af5061cca7f7c6132a9b489fa4c293`. The 41 focused tests pass against the CI-built Python 3.12 wheel, but the cases below remain uncovered, so I am requesting changes.

vane/_video_file.py
kaka11chen: Fixed in ee6e49a253. The probe now configures FFmpeg before avformat_find_stream_info with per-stream max_pixels/max_samples, skip_frame=all, one decoder thread, bounded probe packets/streams/index/analyze duration, and a 5-second timeout for each probe phase. PyAV is now constrained to >=17.1 so stream_options fails closed for formats whose streams are unavailable before stream-info discovery. The native read callback also checks ClientContext cancellation before and after each source read, and the public wording no longer claims zero decoding. Added exact option assertions plus a subprocess regression with RLIMIT_AS and a parent timeout around a forged oversized PNG.

vane/_video_file.py
kaka11chen: Fixed in ee6e49a253. Container validation now requires positive evidence from the supported video-container mapping before MIME validation, so an unknown/image demuxer cannot be accepted by an absent, generic, or forged video MIME. Added direct-Python and SQL regressions for PNG, JPEG, and GIF crossed with NULL, video/*, and video/mp4 content types.

vane/_video_file.py
kaka11chen: Fixed in ee6e49a253. Metadata extraction now checks video.codec_context before accessing decoder-backed stream properties and raises VideoFileFormatError when no decoder is available. Added direct-Python and SQL regressions using a valid Matroska payload whose codec ID is changed to an unsupported equal-length ID; both paths now produce the documented input error rather than AttributeError/InternalException.

kaka11chen:
@codex review

vane/_video_file.py
```diff
@@ -0,0 +1,591 @@
...
+    if not compatible:
+        raise VideoFileFormatError(f"unsupported video container format {','.join(sorted(format_names))!r}")
+
+    if content_type is None:
+        return
+    declared = _canonical_mime_type(content_type)
+    if declared in _GENERIC_MIME_TYPES:
+        return
+    if not declared.startswith("video/"):
+        raise VideoFileFormatError(f"VIDEOFILE content_type {content_type!r} contradicts the video content")
```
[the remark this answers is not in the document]
kaka11chen: Fixed in 7b4aab60da. Video MIME canonicalization now maps application/ogg to video/ogg, matching the existing AudioFile handling while retaining detected-container compatibility checks. Added real VP8/Ogg direct-Python and SQL coverage for both application/ogg and video/ogg.

kaka11chen:
@codex review

kaka11chen:
@codex review

hubgeter (approved):

PR #742, opened 2026-09-02
[feat](file) Stream VideoFile frames

kaka11chen:
## Summary

- Add the frozen, slotted `VideoFrameData` value with exact PTS/DTS, rational time-base, duration, keyframe, and detached RGB image provenance.
- Add bounded, streaming `VideoFile.frames()` and `VideoFile.keyframes()` APIs backed by PyAV and the existing range-aware FILE reader.
- Use one unseeked sequential demux as the correctness baseline for exact time-window, keyframe, interval, and resize selection across reordered frames, distinct presentation/decode timestamp origins, and timestamp discontinuities.

## Related issue

Refs #714

This PR implements the streaming Python value API slice. Efficient indexed/keyframe seeking, exact frame-index access, FILE-aware `VideoFrameSource` distributed execution, SQL collection APIs, and a decoded IMAGE logical type remain follow-up work.

## Documentation impact

      `AstroVela/vane-website`.

The additive public contract is recorded in docstrings, native stubs, and tests. README changes are intentionally out of scope.

## Validation

- `/home/kaka/.local/bin/pre-commit run --files vane/_video_file.py tests/fast/test_video_file.py`
- Incremental non-editable Release install with `CMAKE_BUILD_PARALLEL_LEVEL=24`
- `scripts/run_installed_pytest.sh -q` for the MPEG-TS timestamp-discontinuity, distinct PTS/DTS-origin, MP4 B-frame, and combined-selection regressions — 4 passed
- `scripts/run_installed_pytest.sh -q tests/fast/test_video_file.py` — 168 passed
- `git diff --check`

## Checklist

### DIFF
--- a/vane/_video_file.py
+++ b/vane/_video_file.py
@@ -1344,6 +1344,58 @@ def _seek_video_stream(
         video = None
 
 
+def _seek_video_stream_for_time_window(
+    container: Any,
+    video: Any,
+    requested_timestamp: int,
+    stream_start: int,
+    reader: _VideoReaderProxy,
+    nested_io: _NestedIOBlocker,
+) -> None:
+    """Seek only when the selected random-access point can cover the window.
+
+    FFmpeg's seek index can be ordered by decode timestamps. For streams with
+    reordered frames, seeking to a presentation timestamp may therefore land on
+    a keyframe whose PTS is *after* the requested boundary. Probe the first
+    selected packet without decoding it and keep the optimized seek only when
+    that keyframe is early enough. Otherwise restart at the stream origin, which
+    is the correctness fallback when the container cannot prove safe preroll.
+    """
+
+    packet_and_flush: tuple[Any, bool] | None = None
+    packet: Any = None
+    try:
+        if requested_timestamp <= stream_start:
+            _seek_video_stream(container, video, stream_start, reader, nested_io)
+            return
+
+        _seek_video_stream(container, video, requested_timestamp, reader, nested_io)
+        packet_and_flush = _next_demuxed_packet(container, video, reader, nested_io)
+        seek_is_safe = False
+        if packet_and_flush is not None:
+            packet, is_flush = packet_and_flush
+            packet_and_flush = None
+            if not is_flush:
+                packet_pts = _optional_frame_integer(packet.pts, name="seek packet PTS")
+                try:
+                    is_keyframe = packet.is_keyframe
+                except AttributeError as error:
+                    raise VideoFileFormatError("video demuxer did not report seek packet keyframe state") from error
+                if not isinstance(is_keyframe, bool):
+                    raise VideoFileFormatError("video demuxer reported invalid seek packet keyframe state")
+                seek_is_safe = is_keyframe and packet_pts is not None and packet_pts <= requested_timestamp
+            packet = None
+
+        safe_timestamp = requested_timestamp if seek_is_safe else stream_start
+        _seek_video_stream(container, video, safe_timestamp, reader, nested_io)
+    finally:
+        # Probe failures must not retain native packet/container/codec owners.
+        packet = None
+        packet_and_flush = None
+        container = None
+        video = None
+
+
 def _decode_packet_frames(
     packet: Any,
     reader: _VideoReaderProxy,
@@ -1763,7 +1815,14 @@ def _iter_video_frames(
                 did_seek = options.start_time > 0
                 if did_seek:
                     seek_timestamp = _seek_timestamp(options.start_time, metadata.time_base, stream_start)
-                    _seek_video_stream(container, video, int(seek_timestamp), reader, nested_io)
+                    _seek_video_stream_for_time_window(
+                        container,
+                        video,
+                        int(seek_timestamp),
+                        stream_start,
+                        reader,
+                        nested_io,
+                    )
 
                 next_sample_time = options.start_time if options.sample_interval_seconds is not None else None
                 decoded_batches = _iter_decoded_packet_batches(
--- a/vane/_video_file.py
+++ b/vane/_video_file.py
@@ -154,9 +154,8 @@ class VideoFrameData:
 
     ``frame_time`` is elapsed presentation time from the stream's declared
     timestamp origin. ``frame_time_base`` is the exact unit for the raw PTS,
-    DTS, and duration fields. The zero-based ``frame_index`` is present only
-    when decoding starts at the beginning; after keyframe seek it remains
-    ``None`` rather than being estimated from an average frame rate.
+    DTS, and duration fields. The zero-based ``frame_index`` records the frame's
+    position in the sequential decode stream when the decoder can establish it.
     """
 
     frame_index: int | None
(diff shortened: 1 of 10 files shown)


kaka11chen:
@codex review

vane/_video_file.py
```diff
@@ -544,6 +985,953 @@ def _probe_video_metadata(
...
+    reader: _VideoReaderProxy,
+    nested_io: _NestedIOBlocker,
+) -> None:
+    """Seek at a checked native boundary and restore PyAV's reorder state."""
+
+    try:
+        reorder_depth = _video_reorder_depth(video)
+        _check_video_io(reader, nested_io)
+        try:
+            container.seek(seek_timestamp, backward=True, any_frame=False, stream=video)
```
hubgeter: [P1] Seek early enough to decode reordered frames

`seek_timestamp` is the requested presentation boundary, but passing it directly to the underlying FFmpeg seek does not guarantee enough decoder preroll for reordered frames. With a valid MP4/MPEG-4 fixture at 4 fps (`gop_size=6`, `max_b_frames=2`), full decoding contains `(0.25 s, PTS 4096)` and `(0.5 s, PTS 8192)`, while `frames(start_time=0.25, end_time=0.5)` returns no frames: the seek resumes with its first displayed frame at 0.75 s, so the end-time guard stops immediately. This silently violates the time-window contract. Please seek from a decoder-safe earlier random-access point (falling back to the stream start when that cannot be guaranteed), then apply the exact PTS filter, and add a B-frame time-window regression that compares selective decoding with the equivalent full-decode slice.

hubgeter (changes requested):
Requesting changes for one correctness issue in the new time-window decoding path. On valid video with reordered B-frames, selecting a nonzero start time can silently return an incomplete or empty stream. The inline comment includes a deterministic reproduction and the expected regression coverage.

vane/_video_file.py
kaka11chen: Fixed in 98e83ef927. The time-window path now probes the first packet selected by FFmpeg before decoding. It keeps the requested seek only when that packet is a keyframe whose PTS is at or before the requested presentation boundary; otherwise it seeks back to the declared stream start and lets the existing exact PTS filter select the window. This avoids guessing GOP size or B-frame/reorder depth from codec heuristics. Added the reported MPEG-4 fixture (`gop_size=6`, `max_b_frames=2`) and verifies the 0.25–0.5s selective result matches the equivalent full-decode slice. The regression, 332 related tests, the Release suite, and 14 real-Ray tests pass.

kaka11chen:
@codex review

kaka11chen:
@codex review

vane/_video_file.py
```diff
@@ -544,6 +985,1012 @@ def _probe_video_metadata(
...
+                try:
+                    is_keyframe = packet.is_keyframe
+                except AttributeError as error:
+                    raise VideoFileFormatError("video demuxer did not report seek packet keyframe state") from error
+                if not isinstance(is_keyframe, bool):
+                    raise VideoFileFormatError("video demuxer reported invalid seek packet keyframe state")
+                seek_is_safe = is_keyframe and packet_pts is not None and packet_pts <= requested_timestamp
+            packet = None
+
+        safe_timestamp = requested_timestamp if seek_is_safe else stream_start
```
hubgeter: [P1] Do not treat `stream_start` as a decoder-safe fallback

`AVStream.start_time` is a presentation timestamp, not necessarily the earliest decode/index timestamp, so seeking to it is not equivalent to restoring the container's initial demux position. With a valid MPEG-TS/MPEG-2 fixture at 8 fps (`gop_size=3`, `max_b_frames=1`), `start_time` is 11250 at time base 1/90000; the first keyframe is PTS 11250 / DTS 0 and the next is PTS 33750 / DTS 11250. For `frames(start_time=0.125, end_time=0.375)`, the probe correctly rejects the requested seek, but this fallback lands on the second keyframe. The full-decode slice is PTS `[22500, 33750, 45000]`; the selective call returns `[33750, 45000]`. Please restore or reopen an unseeked demux state (or otherwise establish a proven earliest decode point) rather than seeking to the presentation start, and add a regression where PTS and DTS origins differ.

hubgeter (changes requested):
Re-review of 98e83ef927: the added MP4 regression passes, but the fallback is still not decoder-safe for containers whose presentation and decode origins differ. The inline MPEG-TS reproduction shows that the requested window can still silently lose a frame.

vane/_video_file.py
kaka11chen: Fixed in d5d319b197. I removed the speculative seek path entirely instead of adding another fallback. Every frames()/keyframes() call now keeps the container in its initial, unseeked demux state, decodes sequentially, normalizes presentation time against the declared stream origin, and applies the exact PTS window afterward. Therefore AVStream.start_time is no longer used as a decode/index position, and neither GOP depth nor B-frame reorder depth is guessed. Added the reported MPEG-TS/MPEG-2 regression with gop_size=3 and max_b_frames=1; it proves the fixture has distinct packet PTS/DTS origins and verifies the 0.125–0.375s selective result matches the equivalent full-decode slice. The complete VideoFile test module passes: 167 tests.

kaka11chen:
@codex review

kaka11chen:
@codex review

vane/_video_file.py
```diff
@@ -544,6 +984,872 @@ def _probe_video_metadata(
...
+        for index, info in enumerate(batch.infos):
+            _check_video_io(reader, nested_io)
+            frame = batch.take_frame(index)
+            try:
+                if info.exact_time is not None:
+                    if info.exact_time < options.start_time:
+                        continue
+                    if options.end_time is not None and info.exact_time > options.end_time:
+                        reached_end = True
+                        break
```
hubgeter: [P1] Keep scanning across timestamp discontinuities

This turns `end_time` into a global stop condition, but MPEG-TS supports timestamp discontinuities, so presentation times are not necessarily monotonic over the decoded stream. I reproduced this at the current head by concatenating two 8-frame MPEG-TS/MPEG-2 segments produced with `mpegts_flags=resend_headers+initial_discontinuity` (8 fps, `gop_size=3`, `max_b_frames=1`). PyAV and `ffprobe -v warning` decode all 16 frames, with Vane reporting times `[0, 0.125, ..., 0.875, 0, 0.125, ..., 0.875]`. Full-decode filtering for `[0.125, 0.25]` yields frame indexes `[1, 2, 9, 10]`, while `frames(start_time=0.125, end_time=0.25)` returns only `[1, 2]`; frame 3 hits this branch and prevents the second segment from being read. Please skip an out-of-window frame and continue unless monotonicity has been established for the stream/container, and add a discontinuity regression comparing against the complete decode.

hubgeter (changes requested):
Re-review of d5d319b197: the prior B-frame and PTS/DTS-origin issue is fixed by the unseeked sequential path, and the related regressions pass. One remaining correctness issue affects supported MPEG-TS streams with timestamp discontinuities: the end-time guard still assumes globally monotonic presentation timestamps and can silently omit later in-window frames.

vane/_video_file.py
kaka11chen: Fixed in 6d581a8c07. `end_time` now filters only the current frame; it no longer terminates demux because no global timestamp monotonicity has been established. The sequential decoder therefore continues across later MPEG-TS discontinuities and can select frames whose presentation timestamps re-enter the requested window. I also carry the previous presentation time across packet batches and reset `sample_interval_seconds` targets when the timeline moves backward, so sampling does not silently discard the later segment either. Added the reported two-segment MPEG-TS/MPEG-2 regression using `resend_headers+initial_discontinuity`, `gop_size=3`, and `max_b_frames=1`; both ordinary and sampled `[0.125, 0.25]` selections match full decode at frame indexes `[1, 2, 9, 10]`. The complete VideoFile module passes: 168 tests.

kaka11chen:
@codex review

kaka11chen:
@codex review

hubgeter (approved):

PR #744, opened 2026-09-02
[feat](file) Get exact VideoFile frame by index

kaka11chen:
## Summary

- Add `VideoFile.get_frame_by_idx()` for exact zero-based presentation-order frame access.
- Reuse the range-aware FILE reader and PyAV sequential decoder as the correctness baseline, without PTS/FPS estimation or speculative seeking.
- Convert only the selected frame to a detached RGB Pillow image while preserving input, decoded-frame, and pixel resource limits.
- Cover reordered B-frames, piecewise variable frame rates, timestamp discontinuities, logical byte ranges, invalid indices, out-of-range access, and decode budgets.

## Related issue

Refs #714

This PR closes the exact Python value-access slice. FILE-aware distributed `VideoFrameSource` execution remains in #714.

## Documentation impact

      `AstroVela/vane-website`.

The additive public contract is recorded in the native docstring, type stub, and tests. Expression/SQL access remains gated on a decoded IMAGE logical type.

## Validation

- `pre-commit run --files src/vane_py/native/file.cpp tests/fast/test_video_file.py vane/_native/__init__.pyi vane/_video_file.py`
- Incremental non-editable Python 3.12 Release install with `CMAKE_BUILD_PARALLEL_LEVEL=24`
- `scripts/run_installed_pytest.sh -q tests/fast/test_video_file.py -k get_frame_by_idx` — 9 passed
- `scripts/run_installed_pytest.sh -q tests/fast/test_video_file.py` — 177 passed
- `scripts/run_release_tests.sh` — 806 passed, 5 skipped; 3 unrelated Ray diagnostic-message assertions failed in the fresh local Python 3.12.3/pybind11 3.1 build and passed 3/3 against the previously validated #742 native environment
- `git diff --check`

## Checklist

### DIFF
--- a/src/vane_py/native/file.cpp
+++ b/src/vane_py/native/file.cpp
@@ -249,6 +249,22 @@ static void BindMediaFileClass(py::handle &m, const char *class_name) {
 		    py::arg("max_input_bytes") = DEFAULT_VIDEO_MAX_INPUT_BYTES,
 		    py::arg("max_frames") = DEFAULT_VIDEO_MAX_FRAMES, py::arg("max_pixels") = DEFAULT_VIDEO_MAX_PIXELS,
 		    py::arg("connection") = py::none());
+		file.def(
+		    "get_frame_by_idx",
+		    [](const FILE_TYPE &value, const py::object &idx, const py::object &buffer_size,
+		       const py::object &max_input_bytes, const py::object &max_frames, const py::object &max_pixels,
+		       shared_ptr<DuckDBPyConnection> connection) {
+			    return py::module_::import("vane._video_file")
+			        .attr("_video_file_frame_by_idx_value")(
+			            py::cast(value, py::return_value_policy::copy), idx, buffer_size,
+			            py::arg("max_input_bytes") = max_input_bytes, py::arg("max_frames") = max_frames,
+			            py::arg("max_pixels") = max_pixels, py::arg("connection") = std::move(connection));
+		    },
+		    "Sequentially decode the exact zero-based presentation-order frame into a detached RGB Pillow image",
+		    py::arg("idx"), py::arg("buffer_size") = DEFAULT_VIDEO_BUFFER_SIZE, py::kw_only(),
+		    py::arg("max_input_bytes") = DEFAULT_VIDEO_MAX_INPUT_BYTES,
+		    py::arg("max_frames") = DEFAULT_VIDEO_MAX_FRAMES, py::arg("max_pixels") = DEFAULT_VIDEO_MAX_PIXELS,
+		    py::arg("connection") = py::none());
 	}
 }
 
--- a/vane/_video_file.py
+++ b/vane/_video_file.py
@@ -192,6 +192,7 @@ class _VideoFrameOptions:
     max_input_bytes: int
     max_frames: int
     max_pixels: int
+    target_frame_index: int | None = None
 
 
 @dataclass(frozen=True, slots=True)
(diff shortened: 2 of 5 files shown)


kaka11chen:
@codex review

vane/_video_file.py
```diff
@@ -1803,6 +1818,71 @@ def _video_file_frames_value(
...
+        if frame is not None:
+            _close_image(frame.data)
+            frame = None
+        try:
+            frames.close()
+        except BaseException:
+            pass
+        raise
+    try:
+        frames.close()
```
hubgeter: [P1] Preserve teardown errors after selecting the frame

`next(frames)` leaves `_iter_video_frames()` suspended at its `yield`, so `frames.close()` injects `GeneratorExit`. The existing `_close_container()` and `_close_video_reader()` helpers then take their exceptional-exit branches: container-close failures are suppressed, and the reader uses unchecked `close()` instead of the successful-path interruption check. At this head, I reproduced an interruption raised during normal container teardown being ignored by `get_frame_by_idx(0)`, which returned the image; a direct `RuntimeError` from `container.close()` was swallowed as well. Please let the target-index path reach normal successful teardown (for example, stop naturally after the selected yield and drive the iterator to `StopIteration`, or use a dedicated synchronous core), and add regressions for close-time cancellation/error plus ownership of the selected image when teardown fails.

src/vane_py/native/file.cpp
```diff
@@ -249,6 +249,22 @@ static void BindMediaFileClass(py::handle &m, const char *class_name) {
 		    py::arg("max_input_bytes") = DEFAULT_VIDEO_MAX_INPUT_BYTES,
 		    py::arg("max_frames") = DEFAULT_VIDEO_MAX_FRAMES, py::arg("max_pixels") = DEFAULT_VIDEO_MAX_PIXELS,
 		    py::arg("connection") = py::none());
+		file.def(
+		    "get_frame_by_idx",
```
hubgeter, line 253: [P2] Update the exact `VideoFile` method-surface contract

This adds `get_frame_by_idx` to every `VideoFile` instance, but `VIDEO_FILE_METHODS` in `tests/fast/test_file.py` still lists only `frames`, `keyframes`, and `metadata`. `test_media_file_python_value_contract` compares the complete public `dir(value)` set, so the video case now fails with an extra `get_frame_by_idx` item. I reproduced `1 failed, 2 passed` for that parametrized test after exposing the new method. Please add `get_frame_by_idx` to the expected method tuple so the required fast suite can pass.

hubgeter (changes requested):
Requesting changes for two independently reproducible issues: an in-range lookup treats successful teardown as generator cancellation and suppresses close/cancellation failures, and the existing exact `VideoFile` method-surface contract was not updated for the new public method.

vane/_video_file.py
kaka11chen: Fixed in 6b66c500d9. The target-index iterator now stops after the selected yield is resumed, and the synchronous helper advances it to StopIteration. This lets container.close() and the checked reader close run through their normal-success paths instead of GeneratorExit cleanup. A teardown cancellation or close error is propagated, and the selected image is closed before propagation; only a fully successful teardown transfers image ownership to the caller. The new parameterized regression covers successful ownership transfer, direct container-close failure, and interruption raised during container close. The complete VideoFile test module passes (180 tests).

src/vane_py/native/file.cpp
kaka11chen: Fixed in 6b66c500d9 by adding get_frame_by_idx to VIDEO_FILE_METHODS. The exact public method-surface contract now passes for all three media File value types (3 parametrized cases), and the full test_file.py run completed with 101 passed and 16 optional-fsspec skips in the local validation environment.

kaka11chen:
@codex review

vane/_video_file.py
```diff
@@ -1507,6 +1520,8 @@ def _prepare_video_packet_batch(
             _check_video_io(reader, nested_io)
             frame = batch.take_frame(index)
             try:
+                if options.target_frame_index is not None and info.frame_index != options.target_frame_index:
+                    continue
```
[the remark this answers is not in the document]
kaka11chen: Fixed in 7b4d2644b4. Indexed selection is now an explicit branch: a matching presentation-order frame bypasses the time-window, keyframe, and interval-sampling filters used only by frames(). Resource checks, provenance validation, conversion, and teardown remain shared. Added a regression with frame index 0 at exact_time=-0.5 relative to the declared stream origin; the target is converted and returned instead of being filtered. Targeted selection tests pass 13/13, the full VideoFile module passes 181/181, and pre-commit including Ruff and mypy passes.

kaka11chen:
@codex review

kaka11chen:
@codex review

hubgeter (approved):

Issue #715, opened 2026-08-30
[Tracking][Media File] Python and SQL API completeness


kaka11chen:
### Problem

The delivered File-media and Image stages do not yet cover the complete proposed public API and governed distributed execution contract. Tracking must distinguish implemented operators from provider publication, wider validation, and remaining API work.

At upstream/main `0817d580a0b5e904804611895fa03c452ab452c8` (2026-09-09), Image has a UInt8 dynamic/fixed representation, HWC ndarray materialization, UDF/Arrow/Flight transport, crop, PNG encoding, resize, color conversion and Image-to-Tensor conversion. Audio resampling returns variable-shape Tensors. Video has frame expressions, a fixed RGB Image source, reusable indexes, and complete documented Python/native selection contracts. These delivered capabilities must not remain listed as missing.

### Proposed outcome

The documented ImageFile, AudioFile, VideoFile and decoded Image APIs expose their stated result types, selection semantics and error behavior across Python, SQL, UDFs and distributed execution. File identity, logical byte windows and source/frame provenance remain intact.

Completion includes an explicit supported operation/format/mode matrix, governed multi-provider/multi-scope execution, and public installation and provider-support documentation. Accepted differences from the original proposal are documented. HDF5 is outside the accepted scope.

### Additional context

Delivered stages:

- [x] Shared File media types: #711/#717.
- [x] ImageFile metadata/decoding and IMAGE expression output: #712/#723/#748/#751.
- [x] UInt8 Image and FixedShapeImage storage, ndarray values, attributes, casts, and UDF/Arrow transport: #773.
- [x] Crop and PNG encoding, with explicit Python/native backends in the existing image extension: #781.
- [x] Image resize and color conversion with both explicit backends: #788.
- [x] Base C++ Image-to-Tensor conversion: #793.
- [x] Missing Python File/media helpers and metadata API: #754, delivered by #770.
- [x] Variable-shape Tensor foundation and audio Tensor resampling: #755, delivered by #771/#772.
- [x] Image video expressions/SQL and fixed RGB frame source: #756, delivered by #766/#767.
- [x] Reusable video indexes and selective access: #714/#768; Python/native RGB and full video contracts: #776/#778.
- [x] FILE/IMAGE Flight shuffle transport: #750/#752.
- [x] Independent native image/audio/video extensions and explicit backend selection: #765; initial image/audio profiling and validation: #769.

Remaining work:

[...]

PR #752, opened 2026-09-04
[fix](file) Preserve governed Flight types

### LINKED ISSUE #750: [bug](ray) FILE and IMAGE queries fail across Flight shuffle
### Affected area

Distributed execution or Ray

### What happened?

A Ray query that carries FILE-family or decoded IMAGE values through Flight shuffle fails before result materialization with:

```text
DuckDBError::ValueError flight exchange partition types mismatch
```

Expected behavior: the query completes and returns values with their declared FILE/media/IMAGE logical types, including governed values inside nested results.

### Reproduction

```python
import vane

vane.set_runner_ray(noop_if_initialized=True)
connection = vane.connect()
connection.sql("""
    SELECT i, file('memory://shuffle', NULL, NULL, NULL, NULL) AS value
    FROM range(2) AS values(i)
    ORDER BY i
""").fetchall()
```

The issue also affects decoded IMAGE values:

```sql
SELECT i, image(from_hex('0000ff0000ff'), 2, 1, 3, 'RGB') AS value
FROM range(2) AS values(i)
ORDER BY i;
```

### Environment

- OS: Linux
- Architecture: x86-64
- Python: 3.12.3
- Vane: `0.2.0.dev621`; audited native and relevant media sources match upstream/main `33da8668e3fffa35a39cb074448577e616b63f8a`
- Ray: enabled, local test cluster
- VANE_RUNNER: `ray`

The 2026-09-05 audit used an existing non-editable installation with additional unrelated Python-only Doris support, not a newly built clean main wheel. The issue was originally observed at `224b0271ec`.

### Additional context

The additional regression `test_default_ray_preserves_governed_types_across_flight_shuffle` from #752 failed against the audited implementation:

```text
DuckDBError::ValueError flight exchange partition types mismatch
1 failed in 10.76s
```

The existing 794 related tests and 854 base-release tests passed. The ordinary media execution tests do not cover this failing shuffle case.

Associated PR: #752, open at the time of the audit, head `4fa3b7e5207b1673367ba53c9d59ced42e833838`. Related tracking issues: #659 and #715.

kaka11chen:
## Summary

- accept only exact canonical Arrow storage when Flight transports a declared FILE-family or decoded IMAGE logical type
- validate governed FILE/IMAGE values recursively before restoring aliases with the private internal restoration cast
- preserve direct media subtypes and governed leaves nested in LIST, ARRAY/TENSOR, MAP, STRUCT, and UNION without adding a public STRUCT conversion or compatibility fallback
- share the canonical-storage matcher with distributed Python result materialization and add native plus real-Ray regression coverage

This is an internal distributed-execution correctness fix. Existing successful queries keep the same public API; governed values that previously failed at Flight schema admission now retain their declared logical types, while malformed or non-canonical transport values are rejected.

## Related issue

close: #750

Tracking: #715

## Documentation impact

      `AstroVela/vane-website`.

## Validation

- `scripts/format duckdb --changed` (using the repository's pinned formatter environment)
- `SKBUILD_BUILD_DIR="$PWD/build/python-release" SKBUILD_CMAKE_BUILD_TYPE=Release CMAKE_BUILD_PARALLEL_LEVEL=16 uv pip install . --no-build-isolation`
- `cmake --build build/python-release --target unittest --parallel 16`
- `build/python-release/duckdb/test/unittest 'test/sql/file/test_media_file_alias.test'` — 1 case, 184 assertions passed
- `build/python-release/duckdb/test/unittest '[distributed][exchange][file]'` — 2 cases, 20 assertions passed
- `PATH="$PWD/.venv/bin:$PATH" scripts/run_installed_pytest.sh tests/fast/test_distributed_result_consumers.py::test_distributed_result_normalizes_file_tensor_storage tests/fast/test_distributed_result_consumers.py::test_distributed_result_restores_decoded_image_type` — 2 passed
- `PATH="$PWD/.venv/bin:$PATH" scripts/run_installed_pytest.sh tests/fast/test_file_ray.py::test_default_ray_preserves_governed_types_across_flight_shuffle` — 1 passed
- two consecutive external `gpt-5.6-sol` reviews at max reasoning completed with no findings

## Checklist

No dependency, copied-source, model, or dataset changes are included. The Flight boundary requires exact canonical storage and validates governed values before using the private alias-restoration mode.


kaka11chen:
@codex review

tests/fast/test_file_ray.py
```diff
@@ -85,6 +85,72 @@ def test_default_ray_materializes_scalar_and_nested_file_results(monkeypatch):
...
+                video_file('memory://video/' || i::VARCHAR) AS video_file_value,
+                image(from_hex('0000ff0000ff'), 2, 1, 3, 'RGB') AS image_value,
+                struct_pack(
+                    files := [
+                        file('memory://nested/' || i::VARCHAR, NULL, NULL, NULL, NULL),
+                        NULL::FILE
+                    ],
+                    lookup := map(['preview'], [image_file('memory://preview/' || i::VARCHAR)]),
+                    image := image(from_hex('ff0000'), 1, 1, 3, 'RGB')
+                ) AS nested_value,
```
hubgeter, line 154: [P2] Exercise the advertised ARRAY/TENSOR shuffle path

This real-Ray regression covers scalar governed values plus LIST, MAP, STRUCT, and UNION, but it never sends a fixed-size ARRAY or TENSOR through the shuffle even though the PR explicitly promises both. On this exact head, the following succeeds without `ORDER BY` (two rows, type `FILE[2]`) but fails once `ORDER BY` forces Flight:

```sql
SELECT i,
       array_value(file('memory://array/' || i::VARCHAR, NULL, NULL, NULL, NULL),
                   NULL::FILE) AS files
FROM range(2) AS values(i)
ORDER BY i DESC;
```

The failure has advanced past the old schema mismatch, but the query still ends with `Cannot cast from STRUCT(...)[] to FILE[2]: FILE-family casts require an exact logical type match`. A UDF-produced `TENSOR(FILE, [2])` followed by `.order("i DESC")` fails at the same boundary. Please add real-Ray `FILE[2]` and `TENSOR(FILE, [2])` cases here and preserve the declared governed child through the downstream replay/cast boundary (or use the private restoration mode only after the same exact schema/value validation). The public STRUCT-to-FILE cast should remain closed.

hubgeter (changes requested):
The canonical-storage admission and pre-restoration validation look appropriately narrow, and the scalar/LIST/MAP/STRUCT/UNION paths pass focused exact-head testing. However, the PR currently advertises ARRAY/TENSOR preservation while fixed-size governed sequences still fail after a real Flight shuffle. The inline P2 includes an exact-head reproducer. Please cover and fix that path before merging.

tests/fast/test_file_ray.py
kaka11chen: Fixed in 4fa3b7e520. The failure was downstream of Flight: TupleDataCollection converts fixed ARRAY containers to LIST storage and ArrayType::ConvertToList rebuilt governed STRUCT aliases as plain canonical STRUCTs. The fix preserves governed leaves during that internal layout conversion and adds a private INTERNAL_ARRAY_LAYOUT cast mode that accepts only the exact source == ArrayType::ConvertToList(target) shape. Public STRUCT/LIST-to-FILE casts remain closed; no compatibility or fallback path was added. The real-Ray regression now covers FILE[2] and a UDF-produced TENSOR(FILE, [2]) through ORDER BY, and the native SQL regression covers the row-layout roundtrip. Final validation: Release incremental build passed, test_media_file_alias passed (184 assertions), and the targeted real-Ray test passed. Two consecutive gpt-5.6-sol max reviews found no issues.

hubgeter (approved):

PR #753, opened 2026-09-04
[feat](datasink) Add Apache Doris Arrow Stream Load sink

kaka11chen:
## Summary

Add `DorisStreamLoadSink` so Vane can write large scalar and vector datasets to Apache Doris 4.1 over HTTP Stream Load while keeping batches in Arrow columnar form.

- Require an explicit destination Arrow schema and safely convert scalar and nested columns before upload. Inferred `int64` values become Doris `INT`'s `int32` physical representation; floating-point narrowing rejects finite values that become infinity while preserving normal rounding.
- Accept equivalent coordinator/worker Arrow offset widths, including nested `large_string` / `large_list` representations, without relaxing logical-type or destination safety checks.
- Validate `ARRAY<FLOAT>` vector dimensions, nulls, and finite values without materializing Python rows. Reject unsupported temporal source and destination types before HTTP.
- Reject compact encoded/view inputs before a worker opens; require explicit upstream conversion/rebatching and explicit stringification for non-string/binary string destinations. Normalize null-parent and sliced-out list children before safe casting.
- Preflight aggregate destination-buffer sizes before casting and measure the complete IPC stream before allocating a fixed-size body. Bound input batches, encoded requests, responses, and HTTP body chunks. Expose concurrent workers and Doris-side load parallelism independently.
- Use internal `aiohttp>=3.14.3` and its [documented `encode_basic_auth()` API](https://docs.aiohttp.org/en/v3.14.3/client_advanced.html#authentication) with `Expect: 100-continue`, configured authentication, and one validated FE-to-BE redirect. Reject endpoint query/fragment delimiters and validate trusted host entries; normalize equivalent IPv6 literal spellings without merging distinct DNS hosts. Disable automatic retries because an interrupted upload can have an unknown commit outcome; worker batches are independent transactions.
- Use the full per-worker UUID in batch labels so large worker counts do not encounter the old 32-bit collision risk, while keeping labels within Doris's 128-character limit.
- Preserve the current batch label in interruption diagnostics while retaining cancellation identity and no-retry classification, including interruptions after upload during response validation.
- Add the `doris` package extra, usage documentation, and a required integration job using the digest-pinned official Doris 4.1.3 image. The live test exercises `local-fast`, `local`, and `ray` using a distributable SQL source, round-trips int64 input IDs, strings, and variable-length vectors through an FE redirect, and verifies temporal input leaves a real `DATETIME` table empty.

## Related issue

N/A

## Documentation impact

      `AstroVela/vane-website`.

README uses a distributable SQL example and covers input requirements, destination types, plain input representations, explicit async-caller offloading, worker Arrow representations, vector validation, numeric conversions, temporal rejection, memory limits, concurrency, redirect trust, transaction scope, and unknown outcomes. The sink targets Doris 4.1 Arrow Stream Load; no compatibility or fallback path is included.

## Validation

- Two consecutive clean local reviews after the final conversion, list-normalization, and test/documentation changes.
- `scripts/format root --changed` and all applicable pre-commit checks for the changed files.
- Non-editable Release package installation using the existing incremental build directory; the installed Doris module's SHA-256 matches the updated implementation in the checkout.
- `scripts/run_installed_pytest.sh tests/fast/test_doris_datasink.py tests/fast/test_datasink.py tests/fast/test_package_metadata.py tests/fast/test_expression_udf_map.py::test_vane_function_supports_multiple_and_nested_scalar_udfs -m 'not external_service and not real_ray' -q -q` — 317 passed. The previous CI scalar-UDF timeout was not reproduced; no UDF implementation was changed.
- `scripts/run_installed_pytest.sh tests/fast/test_doris_datasink.py tests/fast/test_datasink.py tests/fast/test_package_metadata.py -m 'not external_service and real_ray' -q -q` — 5 passed. External Doris cases are excluded from these local runs.
- `PYTEST_ADDOPTS="-q -q" scripts/run_release_tests.sh` — 1023 passed in the non-Ray process and 18 passed in the real-Ray process.
- [Required CI](https://github.com/AstroVela/vane/actions/runs/34025084369), including the real Doris integration job and all native/fast-test matrices, is not yet complete on `81a71fd53b`.
- GitHub `@codex review`: two consecutive clean rounds on the unchanged `81a71fd53b` head — [round 1](https://github.com/AstroVela/vane/pull/753#issuecomment-5558403550), [round 2](https://github.com/AstroVela/vane/pull/753#issuecomment-5558433458). Both reported no major issues; no unresolved review threads remain.

Public-API HTTP regressions independently exercise string and variable-length vector inputs on all three runners with two sink workers, 513 rows, and 64-row batches. Every uploaded IPC schema and exact row content is checked for omissions/duplicates. The live Doris cases use two sink workers and one-row batches, then read the committed scalar/vector values from Doris. The same live SQL input builder is also exercised through a real HTTP server on all three runners, without requiring an external service.

Additional regression coverage includes equivalent nested/sliced/chunked worker representations, strict logical-type checks, positive/negative floating-point overflow, regular/large/fixed-size/nested lists, encoded/view rejection, normal rounding near the float32 limit, existing nonfinite values, and hidden values beneath null/sliced lists. Invalid batches issue zero HTTP requests. URL regressions cover empty endpoint query/fragment delimiters (including EnvironmentSecret resolution), IPv6 trust matching, duplicate/invalid host entries, and refusal to forward an upload to an untrusted host.

Cancellation regressions cover four interruption types across upload, redirected upload, and response validation; current-vs-completed batch labels; bounded diagnostics without invoking exception formatting; async-body cancellation; and the public UNKNOWN error boundary with an injected retry budget. Cancellation preserves the exact exception and batch label, suppresses replay, and closes worker resources.

Worker-label regressions force two UUIDs to share their first 31 hex digits, then verify distinct same-operation batch labels with both initial and near-64-bit-limit counters. HTTP headers and result metadata retain the full UUID and stay within Doris's allowed label alphabet and length.

New allocation regressions reject a 1 MiB dictionary value referenced by one million rows before decoding, verify aggregate scalar/nested/multi-column conversion budgets, account for empty nested-list buffers and IPC metadata, and check exact-limit/one-byte-over-limit request allocation. Integer hidden-child regressions cover null, sliced, chunked, nested, regular/large/fixed-size lists.

Real-HTTP async-caller regressions pass for direct `local-fast`/`local` calls and explicit `asyncio.to_thread()` offloading on all three runners, including real Ray. Ray's existing driver deliberately rejects blocking calls on an event-loop thread; no transport thread or automatic fallback is added. The prior Python 3.10 CI failure was an over-strict async-task exception-identity assertion; the test now checks cancellation type, label, single upload, and cleanup without interpreter-version branches.

## Checklist

### DIFF
--- a/tests/fast/test_doris_datasink.py
+++ b/tests/fast/test_doris_datasink.py
@@ -6,10 +6,12 @@ from __future__ import annotations
 import builtins
 import json
 import os
+import threading
 import uuid
-from collections.abc import AsyncIterable, Callable, Mapping
+from collections.abc import AsyncIterable, Callable, Iterator, Mapping
 from dataclasses import dataclass
 from datetime import datetime
+from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
 from typing import Any, cast
 
 import cloudpickle
@@ -19,9 +21,10 @@ import pytest
 import vane
 import vane.datasink.doris as doris
 from vane import DorisStreamLoadSink, EnvironmentSecret
-from vane.datasink import BoundDataSink, WriteContext
+from vane.datasink import BoundDataSink, WriteContext, WriteOutcome
 
 _REAL_LOAD_AIOHTTP = doris._load_aiohttp
+_REAL_OPEN_HTTP_TRANSPORT = doris._open_http_transport
 
 
 class _TransportError(Exception):
@@ -252,6 +255,126 @@ def _success(call: _Call, *, status: str = "Success", **overrides: object) -> _R
     }
     payload.update(overrides)
     return _Response(200, json.dumps(payload).encode())
+
+
+@pytest.fixture
+def _stream_load_server(monkeypatch: pytest.MonkeyPatch) -> Iterator[tuple[str, list[_Call]]]:
+    pytest.importorskip("aiohttp")
+    monkeypatch.setattr(doris, "_open_http_transport", _REAL_OPEN_HTTP_TRANSPORT)
+    calls: list[_Call] = []
+    calls_lock = threading.Lock()
+
+    class Handler(BaseHTTPRequestHandler):
+        protocol_version = "HTTP/1.1"
+
+        def do_PUT(self) -> None:
+            self.connection.settimeout(10)
+            body = self.rfile.read(int(self.headers["Content-Length"]))
+            call = _Call("PUT", self.path, dict(self.headers), body)
+            response = _success(call)
+            with calls_lock:
+                calls.append(call)
+            self.send_response(response.status_code)
+            self.send_header("Content-Type", "application/json")
+            self.send_header("Content-Length", str(len(response.body)))
+            self.end_headers()
+            self.wfile.write(response.body)
+
+        def log_message(self, *_args: object) -> None:
+            pass
+
+    server = ThreadingHTTPServer(("127.0.0.1", 0), Handler)
+    thread = threading.Thread(target=server.serve_forever, daemon=True)
+    thread.start()
+    try:
+        yield f"http://127.0.0.1:{server.server_port}", calls
+    finally:
+        server.shutdown()
+        server.server_close()
+        thread.join(timeout=5)
+
+
+@pytest.fixture(params=["local-fast", "local", pytest.param("ray", marks=pytest.mark.real_ray)])
+def _doris_runner(request: pytest.FixtureRequest, monkeypatch: pytest.MonkeyPatch) -> Iterator[str]:
+    from vane import runners
+
+    runner_type = request.param
+    runner = None
+    if runner_type == "ray":
+        request.getfixturevalue("ray_local")
+        from vane.runners.ray.runner import RayRunner
+
+        runner = RayRunner(address=None, max_task_backlog=None)
+    elif runner_type == "local":
+        from vane.runners.local.runner import LocalRunner
+
+        # LocalRunner writes these settings too; retain pytest's environment cleanup.
+        monkeypatch.setenv("VANE_LOCAL_FTE_WORKERS", "2")
+        monkeypatch.setenv("VANE_LOCAL_FTE_EXECUTION_MODE", "in_process")
+        runner = LocalRunner(num_workers=2)
+    monkeypatch.setattr(runners, "get_or_infer_runner_type", lambda: runner_type)
+    if runner is not None:
+        monkeypatch.setattr(runners, "get_or_create_runner", lambda: runner)
+    try:
+        yield runner_type
+    finally:
+        if runner_type == "ray":
+            assert runner is not None
+            runner.close()
+
+
+@pytest.mark.parametrize("column_name", ["title", "embedding"])
+def test_doris_sink_runner_normalizes_worker_arrow_types(
+    _stream_load_server: tuple[str, list[_Call]], _doris_runner: str, column_name: str
+) -> None:
+    endpoint, calls = _stream_load_server
+    row_count = 513
+    if column_name == "title":
+        projection = "'item_' || i AS title"
+        column_type = pa.string()
+        values = [f"item_{i}" for i in range(row_count)]
+        vectors = {}
+    else:
+        projection = "[i::FLOAT, (i + 1)::FLOAT, (i + 2)::FLOAT] AS embedding"
+        column_type = pa.list_(pa.float32())
+        values = [[float(i), float(i + 1), float(i + 2)] for i in range(row_count)]
+        vectors = {"embedding": 3}
+    destination_schema = pa.schema(
+        [pa.field("id", pa.int32(), nullable=False), pa.field(column_name, column_type, nullable=False)]
+    )
+    with vane.connect() as connection:
+        relation = connection.sql(f"SELECT i::BIGINT AS id, {projection} FROM range({row_count}) AS t(i)")
+        summary = relation.write_datasink(
+            DorisStreamLoadSink(
+                "analytics",
+                "items",
+                endpoint=endpoint,
+                destination_schema=destination_schema,
+                vector_dimensions=vectors,
+                worker_count=2,
+                max_batch_rows=64,
+                timeout=10,
+            ),
+            operation_id=f"doris-{_doris_runner}-{column_name}",
+        )
+
+    assert summary.outcome is WriteOutcome.APPLIED
+    assert summary.rows_received == summary.rows_affected == row_count
+    assert len(calls) >= 9
+    assert len({call.headers["label"] for call in calls}) == len(calls)
+    batches = []
+    for call in calls:
+        assert call.url == "/api/analytics/items/_stream_load"
+        assert call.headers["format"] == "arrow"
+        assert call.headers["Expect"] == "100-continue"
+        assert int(call.headers["Content-Length"]) == len(call.body)
+        batch = pa.ipc.open_stream(call.body).read_all()
+        assert batch.schema == destination_schema
+        assert 0 < batch.num_rows <= 64
+        batches.append(batch)
+    received = pa.concat_tables(batches).sort_by("id")
+    expected = pa.table({"id": list(range(row_count)), column_name: values}, schema=destination_schema)
+    assert received.equals(expected)
 
 
 def test_doris_sink_is_public_without_importing_aiohttp() -> None:
@@ -518,6 +641,124 @@ def test_doris_sink_safely_casts_nested_destination_values() -> None:
 
 
 @pytest.mark.parametrize(
+    ("bound_type", "batch_type", "target_type", "values"),
+    [
+        (pa.string(), pa.large_string(), pa.string(), ["one", "二"]),
+        (pa.large_string(), pa.string(), pa.string(), ["one", "二"]),
+        (pa.binary(), pa.large_binary(), pa.string(), ["one", "two"]),
+        (pa.large_binary(), pa.binary(), pa.string(), ["one", "two"]),
+        (pa.list_(pa.int64()), pa.large_list(pa.int64()), pa.list_(pa.int32()), [[1, 2], [3]]),
+        (pa.large_list(pa.int64()), pa.list_(pa.int64()), pa.list_(pa.int32()), [[1, 2], [3]]),
+        (
+            pa.list_(pa.list_(pa.string())),
+            pa.large_list(pa.large_list(pa.large_string())),
+            pa.list_(pa.list_(pa.string())),
+            [[["one"], None], [[], ["二"]]],
+        ),
+        (
+            pa.large_list(pa.list_(pa.large_string())),
+            pa.list_(pa.large_list(pa.string())),
+            pa.list_(pa.list_(pa.string())),
+            [[["one"], None], [[], ["二"]]],
+        ),
+        (
+            pa.list_(pa.string(), 2),
+            pa.list_(pa.large_string(), 2),
+            pa.list_(pa.string()),
+            [["one", "two"], ["三", None]],
+        ),
+        (
+            pa.list_(pa.list_(pa.string(), 2)),
+            pa.large_list(pa.list_(pa.large_string(), 2)),
+            pa.list_(pa.list_(pa.string())),
+            [[["one", "two"]], [["三", None]]],
+        ),
+    ],
+)
+def test_doris_sink_normalizes_equivalent_worker_types(
+    bound_type: pa.DataType,
+    batch_type: pa.DataType,
+    target_type: pa.DataType,
+    values: list[Any],
+) -> None:
+    destination_schema = pa.schema([("value", target_type)])
+    sink = DorisStreamLoadSink(
+        "analytics", "items", endpoint="http://fe.example:8030", destination_schema=destination_schema
+    )
+    column = pa.chunked_array([values, values], type=batch_type).slice(1, 2)
+    table = pa.table({"value": column})
+    worker = sink.bind(pa.schema([("value", bound_type)])).open_worker(WriteContext("worker-offsets"))
+    _Transport.responses = [_success]
+    try:
+        result = worker.write(table)
+    finally:
+        worker.close()
+
+    assert result.rows_received == result.rows_affected == 2
+    assert result.bytes_received == table.nbytes
+    decoded = pa.ipc.open_stream(_Transport.instances[0].calls[0].body).read_all()
+    expected = pa.table({"value": [values[1], values[0]]}, schema=destination_schema)
+    assert decoded.equals(expected)
+
+
+@pytest.mark.parametrize(
+    ("bound_type", "batch_type"),
+    [
+        (pa.int64(), pa.int32()),
+        (pa.int64(), pa.uint64()),
+        (pa.float32(), pa.float64()),
+        (pa.string(), pa.large_binary()),
+        (pa.binary(), pa.large_string()),
+        (pa.int64(), pa.timestamp("us")),
+        (pa.list_(pa.float32()), pa.large_list(pa.float64())),
+        (pa.list_(pa.int64()), pa.large_list(pa.int32())),
+        (pa.list_(pa.list_(pa.int64())), pa.large_list(pa.large_list(pa.float64()))),
+        (pa.list_(pa.float32(), 3), pa.list_(pa.float32(), 2)),
+        (pa.list_(pa.float32(), 3), pa.large_list(pa.float32())),
+        (pa.list_(pa.float32()), pa.list_(pa.float32(), 3)),
+        (pa.list_(pa.field("item", pa.int64(), nullable=False)), pa.large_list(pa.int64())),
+    ],
+)
+def test_doris_sink_rejects_worker_logical_type_changes_before_http(
+    bound_type: pa.DataType, batch_type: pa.DataType
+) -> None:
+    sink = DorisStreamLoadSink(
+        "analytics", "items", endpoint="http://fe.example:8030", destination_schema=pa.schema([("value", pa.int32())])
+    )
+    worker = sink.bind(pa.schema([("value", bound_type)])).open_worker(WriteContext("worker-type-drift"))
+    try:
+        with pytest.raises(ValueError, match="bound input schema"):
+            worker.write(pa.table({"value": pa.array([], type=batch_type)}))
+    finally:
+        worker.close()
+    assert not _Transport.instances[0].calls
+
+
+@pytest.mark.parametrize(
+    ("source_type", "target_type", "values", "message"),
+    [
+        (pa.int64(), pa.int32(), [[1 << 40]], "cannot be safely cast"),
+        (pa.float64(), pa.float32(), [[1e40]], "floating-point overflow"),
+        (pa.int64(), pa.int32(), [[None]], "non-nullable"),
+    ],
+)
+def test_doris_sink_checks_destination_values_after_worker_offset_normalization(
+    source_type: pa.DataType, target_type: pa.DataType, values: list[Any], message: str
+) -> None:
+    destination_schema = pa.schema([("value", pa.list_(pa.field("item", target_type, nullable=False)))])
+    sink = DorisStreamLoadSink(
+        "analytics", "items", endpoint="http://fe.example:8030", destination_schema=destination_schema
+    )
+    worker = sink.bind(pa.schema([("value", pa.list_(source_type))])).open_worker(WriteContext("worker-safe-cast"))
+    try:
+        with pytest.raises(ValueError, match=message):
+            worker.write(pa.table({"value": pa.array(values, type=pa.large_list(source_type))}))
+    finally:
+        worker.close()
+    assert not _Transport.instances[0].calls
+
+
+@pytest.mark.parametrize(
     ("column", "target_type", "path"),
     [
         (
@@ -728,6 +969,7 @@ def test_doris_sink_rejects_secret_endpoint_delimiters_before_opening_transport(
     assert not _Transport.instances
 
 
+@pytest.mark.parametrize("batch_vector_type", [pa.list_(pa.float32()), pa.large_list(pa.float32())])
 @pytest.mark.parametrize(
     ("vectors", "message"),
     [
@@ -738,17 +980,29 @@ def test_doris_sink_rejects_secret_endpoint_delimiters_before_opening_transport(
         ([[0.1, float("inf"), 0.3], [0.4, 0.5, 0.6]], "finite float32"),
     ],
 )
-def test_doris_sink_validates_vector_values_before_http(vectors: list[list[float] | None], message: str) -> None:
+def test_doris_sink_validates_vector_values_before_http(
+    vectors: list[list[float] | None], message: str, batch_vector_type: pa.DataType
+) -> None:
     schema = _schema(vector_type=pa.list_(pa.float32()))
     worker = _worker(schema=schema)
     with pytest.raises(ValueError, match=message):
-        worker.write(_table(vectors=vectors, schema=schema))
+        worker.write(_table(vectors=vectors, schema=_schema(vector_type=batch_vector_type)))
     assert not _Transport.instances[0].calls
 
 
-def test_doris_sink_accepts_large_list_vectors_and_encodes_plain_list() -> None:
-    schema = _schema(vector_type=pa.large_list(pa.float32()))
-    table = _table(schema=schema)
+@pytest.mark.parametrize(
+    ("bound_vector_type", "batch_vector_type"),
+    [
+        (pa.large_list(pa.float32()), pa.large_list(pa.float32())),
+        (pa.list_(pa.float32()), pa.large_list(pa.float32())),
+        (pa.large_list(pa.float32()), pa.list_(pa.float32())),
+    ],
+)
+def test_doris_sink_accepts_equivalent_list_vectors_and_encodes_plain_list(
+    bound_vector_type: pa.DataType, batch_vector_type: pa.DataType
+) -> None:
+    schema = _schema(vector_type=bound_vector_type)
+    table = _table(schema=_schema(vector_type=batch_vector_type))
     _Transport.responses = [_success]
 
     _worker(schema=schema).write(table)
@@ -976,7 +1230,7 @@ def test_doris_sink_retains_transport_ownership_when_close_fails() -> None:
 
 
 @pytest.mark.external_service
-def test_doris_sink_live_arrow_stream_load() -> None:
+def test_doris_sink_live_arrow_stream_load(_doris_runner: str) -> None:
     endpoint = os.environ.get("VANE_TEST_DORIS_ENDPOINT")
     if not endpoint:
         pytest.skip("VANE_TEST_DORIS_ENDPOINT is required for the external Doris test")
@@ -1017,7 +1271,7 @@ def test_doris_sink_live_arrow_stream_load() -> None:
                 """
             )
 
-        input_table = _table()
+        input_table = _table(schema=_schema(vector_type=pa.list_(pa.float32())))
         assert input_table.schema.field("source_id").type == pa.int64()
         relation = vane.from_arrow(input_table)
         summary = relation.write_datasink(
@@ -1032,6 +1286,8 @@ def test_doris_sink_live_arrow_stream_load() -> None:
                 ),
                 field_mapping={"source_id": "id", "source_title": "title"},
                 vector_dimensions={"embedding": 3},
+                worker_count=2,
+                max_batch_rows=1,
                 trusted_redirect_hosts=(() if redirect_host is None else (redirect_host,)),
                 timeout=60,
             ),
--- a/tests/fast/test_doris_datasink.py
+++ b/tests/fast/test_doris_datasink.py
@@ -501,6 +501,127 @@ def test_doris_sink_safely_casts_nested_destination_values() -> None:
     assert decoded.column("score").to_pylist() == pytest.approx([1.25, 2.5])
 
 
+@pytest.mark.parametrize(
+    ("column", "target_type", "path"),
+    [
+        (
+            pa.chunked_array([[0.1], [None, float("inf"), 1e40]], type=pa.float64()),
+            pa.float32(),
+            "score",
+        ),
+        (
+            pa.chunked_array([[float("-inf"), -1e40]], type=pa.float64()),
+            pa.float32(),
+            "score",
+        ),
+        (
+            pa.chunked_array([[[0.1, None]], [None, [1e40]]], type=pa.list_(pa.float64())),
+            pa.list_(pa.float32()),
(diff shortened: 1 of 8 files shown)


kaka11chen:
@codex review

kaka11chen:
@codex review

kaka11chen:
@codex review

vane/datasink/doris.py
```diff
@@ -0,0 +1,770 @@
...
+            raise ValueError(f"field_mapping contains unknown input columns: {sorted(unknown_mapping)!r}")
+        unknown_vectors = set(vector_dimensions).difference(schema.names)
+        if unknown_vectors:
+            raise ValueError(f"vector_dimensions contains unknown input columns: {sorted(unknown_vectors)!r}")
+
+        fields: list[_FieldBinding] = []
+        for field in schema:
+            target_name = _doris_column("Doris target column", mapping.get(field.name, field.name))
+            vector_dimension = vector_dimensions.get(field.name)
+            if vector_dimension is not None:
```
hubgeter, line 536: [P1] Validate the Arrow types against the destination schema before upload

Only declared vectors receive type checks here; ordinary fields retain their input Arrow types without checking the Doris columns. For example, `pa.table({'value': [1, 2]})` produces int64, which this sink accepts for a destination `value INT`. In the targeted Doris 4.1.3, [the load planner uses the destination column type for Arrow slots](https://github.com/apache/doris/blob/4.1.3/fe/fe-core/src/main/java/org/apache/doris/nereids/load/NereidsLoadScanProvider.java#L370-L375), and [the numeric decoder reinterprets the value buffer as that destination C++ type](https://github.com/apache/doris/blob/4.1.3/be/src/core/data_type_serde/data_type_number_serde.cpp#L320-L339), without an Arrow type check or numeric cast. Applying that decoder to the exact wheel's IPC payload yields INT values `[1, 0]`; float64 `[1.0, 2.0]` sent to FLOAT similarly yields `[0.0, 1.875]`. Both still contain two loaded rows, so `strict_mode` and the response row-count checks cannot detect the corruption. Resolve or require a destination schema and safely cast or reject incompatible fields before sending any batch, including nested element types. Add a live test with inferred Python integers and a Doris INT column.

vane/datasink/doris.py
```diff
@@ -0,0 +1,770 @@
...
+        return pa.chunked_array(chunks, type=pa.list_(pa.float32()))
+
+    def _wire_table(self, table: pa.Table) -> pa.Table:
+        arrays: list[pa.ChunkedArray] = []
+        fields: list[pa.Field] = []
+        for index, binding in enumerate(self._fields):
+            column = table.column(index)
+            if binding.vector_dimension is not None:
+                column = self._vector_column(column, binding)
+            arrays.append(column)
```
hubgeter: [P1] Preserve naive timestamp values when encoding Doris DATETIME columns

This path also copies timezone-naive Arrow timestamps unchanged. Even when the types otherwise match (`timestamp[us]` into Doris `DATETIME(6)`), the targeted Doris 4.1.3 [Arrow reader initializes its timezone from `default_time_zone`](https://github.com/apache/doris/blob/4.1.3/be/src/format/arrow/arrow_stream_reader.cpp#L44-L57), which [is fixed to `+08:00`](https://github.com/apache/doris/blob/4.1.3/be/src/util/timezone_utils.cpp#L56), and [the DATETIME decoder interprets the timestamp as an epoch in that timezone](https://github.com/apache/doris/blob/4.1.3/be/src/core/data_type_serde/data_type_datetimev2_serde.cpp#L430-L445), without checking whether the Arrow timestamp is naive. The exact wheel emits `datetime(2026, 9, 5, 0, 0)` unchanged as a naive timestamp; applying this decoder yields `2026-09-05 08:00:00`. Row counts still match, so the sink accepts the terminal success. Define and enforce the timestamp encoding contract for the supported Doris version: preserve naive wall-clock values through normalization, or reject unsupported temporal inputs before upload. Add a real DATETIME round-trip regression; setting only a Stream Load timezone header would not change this reader's fixed timezone.

hubgeter (changes requested):
Reviewed commit 4941f91841332a322e48a9f0f64c6dadd9e778c4. The sink adds distributed, columnar Arrow IPC writes to Doris with independent batch transactions, guarded FE-to-BE redirects, and disabled automatic retries.

I found two data-integrity issues in the Arrow-to-Doris type contract: unvalidated target numeric types and timezone-naive timestamp handling. Both can preserve the reported row counts while changing the values. Please address the inline findings before merging.

Validation:
- Installed the wheel from this commit's CI run in an isolated environment; verified the changed Python files against the checkout and native identity (v1.5.5-vane.224b0271ec / d73f75edfe). Used aiohttp 3.14.3.
- Doris sink, package metadata, and DataSink tests: 193 passed, 1 deselected.
- Release suite: 899 passed in the non-Ray process and 14 passed in the Ray process.
- Additional real HTTP probes passed: persistent FE/BE requests with 800 KiB bodies, no upload before 100-continue, configured authentication after redirect, a public two-worker write of 513 rows across nine batches, no retry after an uploaded request loses its response, and rejection of an untrusted redirect.
- The current required CI and existing Doris 4.1.3 live integration test are green.

Evidence boundary: the two findings are supported by the exact wheel's emitted IPC data and the corresponding Doris 4.1.3 planner/decoder source. I reproduced the resulting decoder operations locally, but did not execute these two new cases against a running Doris server because this environment cannot access the Docker socket. The existing live test only covers matching BIGINT / ARRAY<FLOAT> / VARCHAR columns.

vane/datasink/doris.py
kaka11chen: Addressed in 2b0b14eca1 with the explicit-rejection contract. The sink rejects temporal types recursively in both source and destination schemas during construction/bind, before a worker or HTTP request exists, rather than relying on Doris timezone behavior. The live Doris 4.1.3 regression creates a real DATETIME table, attempts a timestamp write, verifies the unsupported-temporal error, and then asserts that the table still contains zero rows.

vane/datasink/doris.py
kaka11chen: Addressed in 2b0b14eca1. DorisStreamLoadSink now requires destination_schema, validates every supported scalar and nested type, snapshots the schema at bind time, safely casts every batch column before opening HTTP, and rejects overflow, incompatible values, and top-level/nested nullability violations locally. The live Doris 4.1.3 regression now starts from inferred Arrow int64 IDs, loads them into a Doris INT column through FE-to-BE Stream Load, and asserts exact values 1 and 2.

kaka11chen:
@codex review

kaka11chen:
@codex review

vane/datasink/doris.py
```diff
@@ -0,0 +1,894 @@
...
+        chunks = array.chunks if isinstance(array, pa.ChunkedArray) else (array,)
+        for chunk in chunks:
+            self._validate_nullability(pc.list_flatten(chunk), child_field, f"{path}[]")
+
+    def _destination_column(self, column: pa.ChunkedArray, binding: _FieldBinding) -> pa.ChunkedArray:
+        if binding.vector_dimension is not None:
+            column = self._vector_column(column, binding)
+        if column.type != binding.target_type:
+            try:
+                column = pc.cast(column, binding.target_type, safe=True)
```
hubgeter: [P2] Reject floating-point overflow before uploading the cast batch

`safe=True` does not reject floating-point narrowing overflow: with PyArrow 25, casting `pa.array([1e40], type=pa.float64())` to `pa.float32()` returns `[inf]` without an exception. The same occurs for `list<float64>` values `[[1e40]]` cast to `list<float32>`. Running both through the current worker confirms that the IPC payload containing infinity is sent and a successful mock response reports the row as written. Thus a finite input can silently change to infinity before Doris sees it, despite the documented guarantee that overflow fails locally. Detect finite-to-nonfinite results (or validate the destination floating-point range) before opening HTTP, including nested floating-point elements, and add regression tests asserting that these inputs fail with zero requests. Ordinary rounding during double-to-float conversion need not be rejected.

hubgeter (changes requested):
Quick re-review of 2b0b14eca1de0f94a5d13b8914d0b6e5725319fe. The two original P1 findings are addressed by requiring an explicit destination schema and recursively rejecting temporal types before upload. One P2 remains in the new conversion path: Arrow's safe cast can turn a finite double into float infinity without raising, so the documented local overflow rejection is incomplete.

Validation: 68 focused tests passed (the external Doris test was excluded), using the current Doris Python module with native dependencies from the previous review wheel. A capture-transport reproduction confirmed that both scalar and nested overflowing values reach the HTTP payload and a successful mock response is accepted. This was not a rebuild of the current native code or a live Doris reproduction; the current CI run is still in progress.

vane/datasink/doris.py
kaka11chen: Addressed in a92163e73c. After Arrow's safe cast, the sink now recursively checks floating-point leaves for finite source values that became nonfinite float32 results, before serialization or any HTTP request. Existing NaNs/infinities in ordinary float columns and normal rounding (including near the float32 maximum) are preserved.

Regression tests cover positive and negative overflow, chunked columns, list/large_list/fixed_size_list/nested lists, dictionary-encoded scalars, and null/sliced list buffers. Each overflow case asserts zero transport requests. Unsupported dictionary-of-list casts continue to fail locally without a fallback.

Validation after two consecutive clean local reviews: non-editable Release installation; 211 focused tests passed (1 external-service test deselected); release suite passed with 917 non-Ray tests and 14 real-Ray tests; formatting, Ruff, and mypy passed. Required CI has restarted. The first fresh GitHub Codex review is requested on the updated commit; a second will follow after a clean result.

kaka11chen:
@codex review

kaka11chen:
@codex review

vane/datasink/doris.py
```diff
@@ -0,0 +1,920 @@
...
+        try:
+            self._session = self._loop.run_until_complete(self._open(user, password, timeout))
+        except BaseException:
+            self._loop.close()
+            raise
+
+    async def _open(self, user: str, password: str, timeout: int) -> Any:
+        session = self._aiohttp.ClientSession(
+            headers={
+                "Authorization": self._aiohttp.encode_basic_auth(user, password, encoding="utf-8"),
```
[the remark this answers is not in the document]
kaka11chen: This finding does not apply to the declared dependency range, `aiohttp>=3.14.3,<4`. `aiohttp.encode_basic_auth()` is a public API added in 3.14; the versioned [3.14.3 client reference](https://docs.aiohttp.org/en/v3.14.3/client_reference.html#aiohttp.encode_basic_auth) documents it, and the [3.14.3 authentication guide](https://docs.aiohttp.org/en/v3.14.3/client_advanced.html#authentication) recommends it in place of the deprecated `BasicAuth` constructor.

Verified against the real dependency, not the test double: aiohttp 3.14.3 exports the function, and the installed Doris module (SHA-256 identical to this checkout) successfully initializes and closes its real ClientSession with the Authorization header and automatic retries disabled.

The [live Doris integration job on a92163e73c](https://github.com/AstroVela/vane/actions/runs/33947653982/job/101261761122) also passed. Its log explicitly installs aiohttp 3.14.3 and reports `1 passed, 77 deselected`; this exercises the real transport and FE-to-BE Stream Load against Doris 4.1.3. All Required CI checks on that commit are green. No code change is needed for this comment.

kaka11chen:
@codex review

kaka11chen:
@codex review

vane/datasink/doris.py
```diff
@@ -0,0 +1,920 @@
...
+    endpoint = _name("endpoint", value)
+    try:
+        parsed = urlsplit(endpoint)
+        hostname = parsed.hostname
+        _ = parsed.port
+    except ValueError as error:
+        raise ValueError("endpoint must be a valid HTTP or HTTPS Doris endpoint") from error
+    if parsed.scheme not in {"http", "https"} or hostname is None:
+        raise ValueError("endpoint must be an absolute HTTP or HTTPS Doris endpoint")
+    if parsed.username is not None or parsed.password is not None or parsed.query or parsed.fragment:
```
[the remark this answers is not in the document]
kaka11chen: Addressed in 49c7ac7e49. Endpoint validation now rejects the query and fragment delimiters themselves, including empty `?`/`#` after the authority or root slash, so the Stream Load path cannot be appended into a query or fragment. The same check runs when an EnvironmentSecret endpoint is resolved, before opening the transport. Added constructor and secret-resolution regressions.

Validation after two consecutive clean local reviews: non-editable Release install; 231 focused tests passed (1 external-service test deselected); release suite passed with 937 non-Ray tests and 14 real-Ray tests; formatting, Ruff, and mypy passed.

vane/datasink/doris.py
```diff
@@ -0,0 +1,920 @@
...
+def _trusted_redirect_hosts(value: object) -> tuple[str, ...]:
+    if isinstance(value, str) or not isinstance(value, Sequence):
+        raise TypeError("trusted_redirect_hosts must be a sequence of host names")
+    normalized: list[str] = []
+    for item in value:
+        host = _name("trusted redirect host", item)
+        try:
+            parsed = urlsplit(f"//{host}")
+            parsed_host = parsed.hostname
+            parsed_port = parsed.port
```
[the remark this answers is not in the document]
kaka11chen: Addressed in 49c7ac7e49. Trusted redirect entries now parse bare and bracketed IPv6 literals explicitly. The allowlist, endpoint host, and redirect host share IPv6 canonicalization, so compressed/expanded spellings match and equivalent duplicate entries are rejected. Host entries with ports, URL delimiters, or malformed addresses are rejected. DNS normalization also avoids Unicode case folding that could merge distinct IDNs.

Added regressions for bare/bracketed/expanded IPv6 entries, implicit trust of an equivalent endpoint address, invalid ports, duplicate IPv6 spellings, and rejection of untrusted IPv6 and distinct case-folded DNS hosts before a second request. README documents IPv6 forms and the no-port requirement.

Validation: 231 focused tests passed; release suite passed with 937 non-Ray and 14 real-Ray tests; formatting, Ruff, and mypy passed.

kaka11chen:
@codex review

kaka11chen:
@codex review

vane/datasink/doris.py
```diff
@@ -0,0 +1,929 @@
...
+        return label
+
+    def _validate_table(self, table: pa.Table) -> tuple[int, int]:
+        if not isinstance(table, pa.Table):
+            raise TypeError(f"DorisStreamLoadSink expected pyarrow.Table, got {type(table).__name__}")
+        if len(table.schema) != len(self._schema) or any(
+            batch_field.name != bound_field.name or batch_field.type != bound_field.type
+            for batch_field, bound_field in zip(table.schema, self._schema, strict=True)
+        ):
+            raise ValueError("Doris Stream Load batch schema does not match the bound input schema")
```
hubgeter, line 687: [P1] Accept equivalent Arrow representations from execution workers

The coordinator records `string` and regular `list` types by default, but both [Ray workers](https://github.com/AstroVela/vane/blob/49c7ac7e49c076c625a351ae97ad16c011bb2c69/vane/runners/ray/worker.py#L950) and the [local runner](https://github.com/AstroVela/vane/blob/49c7ac7e49c076c625a351ae97ad16c011bb2c69/vane/runners/local/runner.py#L537) set `arrow_large_buffer_size=true`. Their batches therefore contain `large_string` / `large_list`, and this strict type comparison rejects them before `_destination_column()` can safely convert them to the declared destination schema.

For example, writing `SELECT i::BIGINT AS id, 'item_' || i AS title FROM range(4) AS t(i)` to an explicit `id: int32, title: string` destination schema fails in both `ray` and `local` with `Doris Stream Load batch schema does not match the bound input schema`. An ordinary `FLOAT[]` vector has the same problem. With two workers, both cases emitted zero HTTP requests; integer-only and fixed-size vector controls passed, and all cases passed in `local-fast`.

Validate and normalize compatible Arrow representations while retaining column-name, logical-type, and destination safe-cast checks. Add real-Ray regressions for string and variable-length vector columns so the sink's distributed path is exercised.

hubgeter (changes requested):
Reviewed commit `49c7ac7e49c076c625a351ae97ad16c011bb2c69`.

Writes containing string or variable-length vector columns fail in both `ray` and `local` before upload because the worker's Arrow representation differs from the schema captured during binding. This blocks ordinary distributed use of the new sink. Please address the P1 inline finding and add real-Ray coverage for these inputs.

Validation:

- Used the Python 3.11 wheel from this PR's CI run, built from merge commit `3f8c8357ef748701847ea547d41a6da6f6d0fa49`. The installed Doris module and DataSink framework match the reviewed head byte for byte; native identity was checked against the matching merge checkout.
- Focused Doris, DataSink, and package metadata tests: 230 passed, 1 skipped, 1 external-service test deselected.
- Release suite: 936 non-Ray tests and 14 real-Ray tests passed; the optional ADBC test was skipped.
- Real HTTP probes passed for FE-to-BE redirects, configured authentication, `100-continue`, and no replay after a persistent connection lost its response. A `local-fast` write delivered 513 rows across nine batches without missing or duplicate rows.
- Public-API probes with two workers reproduced the schema failure in both `ray` and `local`: integer-only and fixed-size vector inputs succeeded; string and variable-length vector inputs failed with zero HTTP requests. All four inputs succeeded in `local-fast`.

The probes used a local HTTP test server and real workers. A live Doris database was not exercised locally. The passing standard suites do not cover the additional failing distributed cases above.

vane/datasink/doris.py
kaka11chen: Fixed in `12e105a39efff11c18022a1ca4e1a3037d968ddd`.

The strict physical-type comparison was wrong at the coordinator/worker boundary. Batch validation now accepts equivalent string/binary offset widths and recursively equivalent list offsets, then normalizes directly through the existing destination safe cast. Column names/order, scalar logical types, fixed-list dimensions, nested nullability, destination overflow checks, and vector validation remain enforced; there is no fallback path.

Added public-API regressions using a real HTTP server with `local-fast`, `local`, and real `ray` workers. String and variable-length FLOAT vector inputs are tested separately, with two sink workers, 513 rows, and 64-row batches. Every uploaded IPC schema and all row values are checked, including no missing/duplicate rows. Unit regressions cover nested/sliced/chunked equivalent representations and ensure logical-type drift, unsafe narrowing, invalid vectors, and destination nulls still fail before HTTP.

The live Doris test now runs all three runners with a variable-length vector, a string column, two workers, and one-row batches, then queries Doris to verify exact values. That expanded live test is running in [the new CI run](https://github.com/AstroVela/vane/actions/runs/34013441170); the previous head's entire CI, including live Doris, was green.

Validation after two consecutive clean local reviews:

- Formatting and all pre-commit checks, including Ruff and mypy: passed.
- Non-editable incremental Release installation; installed Doris source SHA-256 matches the checkout.
- Affected tests: 267 non-Ray + 3 real-Ray passed.
- Release gate: 974 non-Ray + 16 real-Ray passed.

Restarting the two consecutive GitHub `@codex review` rounds on this new head.

kaka11chen:
@codex review

kaka11chen:
@codex review

tests/fast/test_doris_datasink.py
```diff
@@ -0,0 +1,1346 @@
...
+                ) ENGINE=OLAP
+                DUPLICATE KEY(`id`)
+                DISTRIBUTED BY HASH(`id`) BUCKETS 1
+                PROPERTIES ("replication_num" = "1")
+                """
+            )
+
+        input_table = _table(schema=_schema(vector_type=pa.list_(pa.float32())))
+        assert input_table.schema.field("source_id").type == pa.int64()
+        relation = vane.from_arrow(input_table)
```
hubgeter: [P2] Use a distributable input relation for the live runner cases

This test now runs with `_doris_runner` set to `local` and `ray`, but it still constructs its input with `vane.from_arrow(input_table)`. That produces an `arrow_scan` which the local distributed plan translator cannot copy. On the current head, [the required Doris integration job](https://github.com/AstroVela/vane/actions/runs/34013441170/job/101440456699) fails the local parameter with `Distributed execution does not support table function "arrow_scan": Copy not supported for TableFunctionData`, before it can exercise Stream Load.

Use a supported distributable SQL/file relation, as the new HTTP regressions already do, while retaining the string/vector columns, two sink workers, small batches, and Doris readback assertions. Otherwise the newly added distributed live coverage cannot validate the fix and keeps the required CI gate red.

hubgeter (changes requested):
Re-reviewed `12e105a39efff11c18022a1ca4e1a3037d968ddd`, focusing on the changes since `49c7ac7e`.

The previous P1 is addressed: equivalent string/binary offset widths and nested list representations are accepted, while logical-type, destination safe-cast, overflow, nullability, and vector checks remain enforced. The new public-API string and variable-length vector regressions passed locally on `local-fast`, `local`, and real `ray` workers.

One P2 remains in the expanded live integration test: its `from_arrow()` input cannot be dispatched by the local distributed runner, so the required Doris job fails before that case can validate Stream Load. Please use a supported distributable input and rerun the live gate.

Validation used the Python 3.11 wheel from [this head's CI run](https://github.com/AstroVela/vane/actions/runs/34013441170), built from merge commit `7d7a63a2ca890c0dd6633deb487b7619e77bd495`. The installed Doris module matches the reviewed head byte for byte; native identity matches the artifact's source manifests.

- Focused non-Ray tests: 238 passed, including the scalar-UDF case that timed out in CI.
- Focused real-Ray tests: 3 passed.
- Release gate: 973 non-Ray tests and 16 real-Ray tests passed; the optional ADBC test was skipped because its dependency was unavailable.
- Current CI's live Doris result is 1 passed and 2 failed: the local case reports unsupported `arrow_scan`, and the Ray case has a `SIGSEGV`/`ActorDiedError`. The separate fast-test scalar-UDF timeout was not reproduced locally.

No live Doris database was exercised locally.

tests/fast/test_doris_datasink.py
kaka11chen: Fixed in `79540fe96b77fe56a027728c404bf9ff2f0c44bc`.

The live test now creates its input from a distributable `range()` SQL relation, not an in-memory `arrow_scan`. The int64 source ID, string column, variable-length FLOAT vector, field mapping, two sink workers, one-row batches, and exact Doris readback assertions are preserved. The temporal negative input also uses SQL.

The live input builder is shared with a new real-HTTP regression on `local-fast`, `local`, and real `ray`. It verifies two distinct uploads, the destination IPC schema, and every scalar/vector value without requiring a Doris service, so a future unsupported-input change is covered by the regular runner tests. The README example had the same input issue; it now uses SQL and documents the distributed sink input requirement.

Validation after the final two consecutive clean local reviews:

- Applicable formatting/pre-commit checks and a non-editable incremental Release installation passed.
- Affected tests: 270 non-Ray + 4 real-Ray passed.
- The non-Ray invocation also includes `test_vane_function_supports_multiple_and_nested_scalar_udfs`; the prior CI timeout was not reproduced. No UDF implementation was changed.
- Release gate: 976 non-Ray + 17 real-Ray passed.

[New CI run](https://github.com/AstroVela/vane/actions/runs/34020360785) is running on this head, including the real Doris three-runner gate. Restarting the two GitHub `@codex review` rounds on this commit.

kaka11chen:
@codex review

vane/datasink/doris.py
```diff
@@ -0,0 +1,962 @@
...
+            raise RuntimeError("Doris Stream Load returned invalid JSON") from error
+        if not isinstance(payload, Mapping):
+            raise RuntimeError("Doris Stream Load returned a non-object JSON response")
+        return payload
+
+    def _put(self, label: str, body: pa.Buffer) -> Mapping[str, Any]:
+        headers = self._headers(label, body.size)
+        try:
+            return self._response_payload(self._put_once(headers, body))
+        except Exception as error:
```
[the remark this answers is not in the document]
kaka11chen: Fixed in f23bed7e7d45052b5ceb5a863ddd5f0078a40e98.

The batch-level guard now spans the HTTP upload, FE-to-BE redirect, JSON decoding, and terminal-result validation. Interruptions escaping that guard retain the original exception object/type and cause chain, with the current batch label and inspect-before-retry guidance attached to `args`. This is deliberate: DataSink's bounded public diagnostics read `args`, not exception notes. Normal exceptions still retain their cause and are reported as labeled failures. Diagnostic extraction does not invoke exception formatting, and oversized messages are bounded before encoding.

Added 18 regression cases covering `KeyboardInterrupt`, `asyncio.CancelledError`, `SystemExit`, and `GeneratorExit`; upload/redirect/result-validation interruption; distinction from the previous completed batch's label; empty, non-string, oversized, and unprintable diagnostics; cancellation after the async body iterator is consumed; and the public DataSink error boundary with a fault-injected retry budget. The public error remains UNKNOWN, contains the exact label, and is not retried; cleanup still runs.

Validation after two consecutive clean local reviews: all applicable pre-commit checks passed; non-editable installation passed; 288 affected non-Ray tests and 4 real-Ray tests passed; the release gate passed 994 non-Ray tests and 17 real-Ray tests. Real Doris CI and two consecutive GitHub review rounds are being rerun on this new head.

kaka11chen:
@codex review

vane/datasink/doris.py
```diff
@@ -0,0 +1,962 @@
...
+        self._schema = schema
+        self._destination_schema = destination_schema
+        self._fields = fields
+        self._endpoint_scheme = endpoint_parts.scheme
+        self._trusted_hosts = frozenset(trusted_hosts)
+        self._load_path = f"/api/{quote(sink.database, safe='')}/{quote(sink.table, safe='')}/_stream_load"
+        self._url = f"{endpoint}{self._load_path}"
+        self._transport: _AioHttpTransport | None = _open_http_transport(sink.user, password, sink.timeout)
+        operation_digest = hashlib.sha256(context.operation_id.encode("utf-8")).hexdigest()[:16]
+        self._label_stem = f"{_LABEL_PREFIX}_{operation_digest}_{uuid.uuid4().hex[:8]}"
```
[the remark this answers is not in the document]
kaka11chen: Fixed in 4ecab976bf3fde6969cce0b2223c056985656a6a. Worker label stems now retain the entire UUID instead of truncating it to eight hex digits.

Added deterministic collision regressions with two UUIDs that share their first 31 hex digits and differ only in the last digit. Both workers use the same operation ID and emit two batches; all four labels must remain distinct and include the complete worker UUID. The tests cover both initial batches and counters near the signed 64-bit limit, verify HTTP headers and result metadata agree, and enforce Doris's label-character and 128-character limits.

After two consecutive clean local reviews: all applicable pre-commit checks and the non-editable installation passed; 290 affected non-Ray tests and 4 real-Ray tests passed; the release gate passed 996 non-Ray tests and 17 real-Ray tests. Real Doris CI and two consecutive GitHub review rounds are being rerun on this head.

kaka11chen:
@codex review

vane/datasink/doris.py
```diff
@@ -0,0 +1,962 @@
...
+        chunks = array.chunks if isinstance(array, pa.ChunkedArray) else (array,)
+        for chunk in chunks:
+            self._validate_nullability(pc.list_flatten(chunk), child_field, f"{path}[]")
+
+    def _destination_column(self, column: pa.ChunkedArray, binding: _FieldBinding) -> pa.ChunkedArray:
+        if binding.vector_dimension is not None:
+            column = self._vector_column(column, binding)
+        if column.type != binding.target_type:
+            try:
+                converted = pc.cast(column, binding.target_type, safe=True)
```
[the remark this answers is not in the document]
kaka11chen: Fixed in 81a71fd53be5762cbb08842cc65d8b5672845ef1.

Binding now accepts only plain scalar and standard list representations. Dictionary, run-end encoded, view, and other unsupported representations are rejected recursively before a worker/transport opens; callers must explicitly convert and rebatch them first. String destinations require string/binary input so implicit stringification cannot bypass size accounting. There is no compatibility or fallback path.

For accepted inputs, destination-buffer estimates are summed across columns before any destination cast. The complete IPC stream is then sized with Arrow's non-buffering sizing stream, including schema and chunk metadata, before allocating one fixed-size request buffer. Both conversion and encoded-body limits fail locally before HTTP.

Regressions cover the exact 1 MiB dictionary value referenced by one million rows (with casting forbidden), recursively encoded/view types, scalar/nested/multi-column expansion, null-list child buffers, IPC metadata overhead, and exact-limit/one-byte-over-limit allocation. README documents the supported input contract and memory accounting; the PR validation section records the completed checks.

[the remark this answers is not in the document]
kaka11chen: Fixed in 81a71fd53be5762cbb08842cc65d8b5672845ef1.

List inputs are normalized with Arrow's logical `list_flatten` before safe casting: sliced-out and null-parent child ranges are removed, offsets are rebuilt, and parent null masks are retained. This happens recursively, so the same visible values reach both the safe cast and the floating-point overflow/nullability checks; no Python row materialization is introduced.

Added integer-overflow regressions for regular, large, and fixed-size lists, including nested, sliced, and chunked arrays with out-of-range payloads beneath null parents and outside the slice. These batches now preserve their logical null/value contents. Existing regressions still reject visible integer/float overflow and visible nulls for non-nullable destinations.

vane/datasink/doris.py
```diff
@@ -0,0 +1,962 @@
...
+    """Synchronous worker facade over aiohttp's real 100-continue support."""
+
+    def __init__(self, user: str, password: str, timeout: int) -> None:
+        aiohttp = _load_aiohttp()
+        self._aiohttp = aiohttp
+        self._loop = asyncio.new_event_loop()
+        self._session: Any | None = None
+        self._used_https = False
+        try:
+            self._session = self._loop.run_until_complete(self._open(user, password, timeout))
```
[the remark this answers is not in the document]
kaka11chen: Verified the reported public-call scenario in 81a71fd53be5762cbb08842cc65d8b5672845ef1; the worker does not run in the caller's event-loop thread.

`DataSinkWorker` is owned by an actor process. `DataSinkExecutionOptions.map_batches_kwargs()` selects `subprocess_actor` for `local-fast`/`local` and `ray_actor` for Ray. Consequently, opening the Doris worker through `relation.write_datasink()` does not nest its HTTP loop inside the caller's loop. New real-HTTP regressions invoke the public write directly inside `asyncio.run()` with a running-loop assertion and pass on both `local-fast` and `local`, with two workers and two distinct uploaded batch labels.

Ray deliberately rejects blocking ObjectRef resolution on an event-loop thread in its driver, before a Doris worker opens. The supported async-caller pattern is to offload the complete synchronous connection/relation/write operation with `asyncio.to_thread()`; additional real-HTTP regressions pass on all three runners, including real Ray. README now documents that pattern. No extra transport thread, version branch, or fallback was added.

kaka11chen:
@codex review

kaka11chen:
@codex review

hubgeter (approved):

PR #767, opened 2026-09-05
[feat](video) Add Image frame expressions for both backends

### LINKED ISSUE #756: [VideoFile] Missing native Image frame Expression and SQL APIs
### Problem

At upstream/main `33da8668e3fffa35a39cb074448577e616b63f8a`, Python VideoFile frame iterators and a FILE-aware Worker data source are available, but the proposed typed video expression and SQL surface is incomplete.

The following calls are absent:

- Python/SQL `video_keyframes`, `video_frames`, and `get_video_frame_by_idx`.
- Expression methods `expr.video_keyframes()` and `expr.video_frames()`.
- Python `read_video_frames` and its SQL table function.

The current data source returns `frame: TENSOR(UTINYINT, [H, W, 3])` with `file: VIDEOFILE` and temporal metadata. The proposal instead specifies native Image frame values, including a fixed-shape RGB `data` column for the streaming source.

### Proposed outcome

Users can express keyframe, time-window, interval, resize, and exact frame-index selection in Python expressions and SQL and receive native Image values with reproducible source/frame metadata.

`video_keyframes` returns a List of Images; `video_frames` returns a List of frame records; `get_video_frame_by_idx` returns one Image. `read_video_frames` returns one row per frame with the documented fixed-shape RGB Image column.

Matching methods and functions have consistent arguments. Scalar-list and streaming calls have defined output limits, NULL/error behavior, cancellation, and source association.

_No response_

### Additional context

Parent: #715. Selective access and fallback behavior: #714. Shared distributed issues: #661 and #750/#752.

Delivered stages include #736, #742, #744, and #746. Native IMAGE and ImageFile expression decoding are available through #748/#751; their presence does not make these video APIs available.

- [Video value methods](https://github.com/AstroVela/vane/blob/33da8668e3fffa35a39cb074448577e616b63f8a/src/vane_py/native/file.cpp#L221)
- [Current video SQL registration](https://github.com/AstroVela/vane/blob/33da8668e3fffa35a39cb074448577e616b63f8a/src/vane_py/video_file_functions.cpp#L268)
- [Current data-source schema](https://github.com/AstroVela/vane/blob/33da8668e3fffa35a39cb074448577e616b63f8a/vane/datasource/video_reader.py#L935)

kaka11chen:
## Summary

Complete the video expression surface with Python/SQL `video_frames`, `video_keyframes`, and `get_video_frame_by_idx`, plus matching frame/keyframe Expression methods. Frame lists contain the original VIDEOFILE view, presentation-order index and timing metadata, and native RGB IMAGE values; keyframes and index lookup return IMAGE values directly.

The explicit `video_backend` setting selects execution at bind time. Python execution uses a C++ scalar bridge and PyAV with the current query's FILE context; native execution runs FFmpeg directly in the independently loadable DuckDB video extension. SQL macros supply named/default arguments to those scalar functions. FILE byte windows, Worker credentials, cancellation, and error categories are preserved. No compatibility layer or automatic backend/materialization fallback is introduced.

Explicit IMAGE casts over fields retain their validation during column pruning. Unnested frames can be constrained to a fixed layout, while MAP key conversions reject collisions before CTAS or INSERT writes them. Field extraction can still be pushed down.

Scalar output has per-row byte/frame limits and a per-chunk byte ceiling, including allocations retained after a suppressed format failure. Exact-index lookup uses the existing sequential presentation-order baseline. `read_video_frames` provides the streaming path for results that exceed scalar limits.

## Related issue

Closes #756. Builds on the merged #766, which supplies public streaming and fixed IMAGE types. Indexed seeking remains #714.

## Documentation impact

      `AstroVela/vane-website`.

VIDEO_FRAME_API.md documents the result schemas, function/method options, limits, NULL/error behavior, provenance, and backend execution. NATIVE_MEDIA_EXTENSIONS.md lists the new video extension operators.

## Validation

- Head `62fe3cfd23` passed two consecutive requested GitHub reviews after the column-pruning fix: [round 1](https://github.com/AstroVela/vane/pull/767#issuecomment-5556691994), [round 2](https://github.com/AstroVela/vane/pull/767#issuecomment-5556715988).
- Two consecutive clean local static reviews before each native build; pre-commit, mypy, `scripts/format workspace --from-ref upstream/main --check`, and `git diff --check` passed.
- Noneditable incremental Release installation (vane-ai 0.2.0.dev626, head `62fe3cfd23`) with `VANE_LOADABLE_EXTENSIONS=image;audio;video`, the CI test signing key, and `BUILD_UNITTESTS=ON`. `cmake --build build/python-release --target vane_loadable_extensions unittest --parallel 8` passed.
- `scripts/run_installed_pytest.sh tests/fast/test_video_expressions.py tests/fast/test_native_media_extensions.py tests/fast/test_read_video_frames.py`: 169 passed, including all 61 new expression cases for Python/native selection, FILE windows, NULL/error categories, output budgets, query-context lifetime, and multi-chunk reuse.
- 459 cases passed (12 new column-pruning regressions and 447 existing cases) across `test_fixed_shape_image.py`, `test_image.py`, `test_file_udf.py`, `test_video_file.py`, and `test_video_optional_deps.py`.
- `scripts/run_installed_pytest.sh -m real_ray tests/fast/test_ray_video_expressions.py tests/fast/test_ray_native_media_extensions.py tests/fast/test_ray_read_video_frames.py`: 15 passed with matching signed provider wheels; includes both expression backends through UDF/Flight, bounded FILE views, and fixed IMAGE casts over unnested frames.
- C++ checks: `[distributed][exchange][file]`, `test/sql/file/test_image_type.test`, struct pushdown validity, struct projection pushdown, unused columns after joins, CTE cast pushdown, and ROLLUP column pruning: 8 cases, 214 assertions passed.
- `scripts/run_release_tests.sh`: 839 non-Ray and 14 Ray passed; one optional ADBC test skipped because `adbc_driver_manager` is absent.

A local CPU measurement of the unchanged frame operators (vane-ai 0.2.0.dev632) used Linux x86-64, Intel Xeon E5-2686 v4, 36 exposed CPUs, Python 3.12, PyAV 17.1.0, `VANE_RUNNER=local-fast`, and connection `threads=1`. Each query consumed 32 rows referencing the same locally generated 51,168-byte MPEG-4 file (120 frames, 320x240, 24 FPS, GOP setting 12, max B-frames 2). After one warmup, medians over five executions were:

| Operation and selection | Python | Native | Python/native time |
| --- | ---: | ---: | ---: |
| `video_frames`: 0–1 s, interval 0.2 s, 160x120 | 769.4 ms | 424.9 ms | 1.81x |
| `video_keyframes`: 160x120 | 890.9 ms | 507.0 ms | 1.76x |
| `get_video_frame_by_idx`: index 90, source size | 548.3 ms | 302.6 ms | 1.81x |

The queries aggregate list length or IMAGE byte length; both backends produced the same results (192 frames, 512 keyframes, and 7,372,800 pixel bytes respectively). This measures the current local execution paths on one small synthetic fixture, without network or distributed execution.

## Checklist

      licenses and are recorded where required. This uses the existing optional
      FFmpeg/PyAV dependencies and locally generated video fixtures.


kaka11chen:
@codex review

Please review the full PR at 69777234ba707836b2410f8f6f39acf1b2e08a76. This is the first of two consecutive clean reviews requested after local static reviews, native builds, affected tests, Ray tests, and the release suite passed. The PR depends on #766; the expression changes are in 6eca23e456 and c782d8dc0b.

kaka11chen:
@codex review

Please review the full current PR diff at `77d10836648ec724377db27b21054a72b173746c`. This is round 1 of two sequential reviews on the final head after incorporating the reviewed IMAGE cast fixes from #766. Local static reviews, incremental Release build, affected C++/Python/Ray tests and the release suite have passed.

kaka11chen:
@codex review

Please perform a second full review of the unchanged final head `77d10836648ec724377db27b21054a72b173746c`. The first requested review completed without findings: https://github.com/AstroVela/vane/pull/767#issuecomment-5554729028 . No code changes have been made since that review.

external/duckdb/src/function/cast/map_cast.cpp
```diff
@@ -145,11 +146,55 @@ static bool MapToVarcharCast(Vector &source, Vector &result, idx_t count, CastPa
...
+	return success;
+}
+
 BoundCastInfo DefaultCasts::MapCastSwitch(BindCastInput &input, const LogicalType &source, const LogicalType &target) {
 	switch (target.id()) {
-	case LogicalTypeId::MAP:
-		return BoundCastInfo(ListCast::ListToListCast, ListBoundCastData::BindListToListCast(input, source, target),
+	case LogicalTypeId::MAP: {
+		auto cast = ListCast::ListToListCast;
+		if (input.file_cast_mode == FileCastMode::EXPLICIT_IMAGE_LAYOUT &&
```
hubgeter, line 190: [P1] Preserve explicit IMAGE cast mode through column pruning

Selecting `MapImageKeyCast` only in `EXPLICIT_IMAGE_LAYOUT` mode leaves its duplicate-key validation vulnerable to the existing column-pruning rewrite. `RemoveUnusedColumns::RewriteExpressions` reconstructs casts with `BoundCastExpression::AddCastToType` (`src/optimizer/remove_unused_columns.cpp:625`), preserving the target type but losing the explicit mode. An IMAGE-bearing MAP cast then uses `ListToListCast`, allowing keys that collide after conversion to be stored.

This reproduces on the reviewed head with the default optimizer:

```sql
CREATE TABLE stored AS
SELECT CAST(row_value.m AS MAP(STRUCT(image IMAGE, label INTEGER), INTEGER)) AS m
FROM (
    SELECT unnest([{'m': MAP([
        {'image': CAST(image('abc'::BLOB, 1, 1, 3, 'RGB') AS IMAGE('RGB', 1, 1)),
         'label': '01'},
        {'image': CAST(image('abc'::BLOB, 1, 1, 3, 'RGB') AS IMAGE('RGB', 1, 1)),
         'label': '1'}
    ], [10, 20])}]) AS row_value
);

SELECT cardinality(m),
       list_transform(map_keys(m), key -> key.label),
       map_values(m)
FROM stored;
-- Returns (2, [1, 1], [10, 20]): duplicate keys persisted.
```

On a fresh connection, running `SET disabled_optimizers='unused_columns'` before the same CTAS correctly raises `Map keys must be unique.` The same loss of mode also makes a matching `CAST(frame.data AS IMAGE('RGB', 1, 1))` over an unnested generic IMAGE fail with the governed-type exact-match binder error; disabling that optimizer makes it succeed.

Please preserve the original explicit cast semantics when rebuilding/pushing down these casts, or prevent this rewrite for casts that require the IMAGE validation mode. Regression coverage should verify both rejection of colliding MAP keys before storage and successful matching IMAGE casts through UNNEST with the default optimizer enabled.

hubgeter (changes requested):
Requesting changes for one reproducible correctness issue at `77d10836648ec724377db27b21054a72b173746c`: column pruning drops the new explicit IMAGE cast mode. This can persist duplicate MAP keys and also reject an otherwise valid fixed-shape IMAGE cast. The inline finding includes a standalone SQL reproducer and the optimizer-dependent behavior.

Validation:

- Reproduced both failures with the non-editable CI wheel `vane-ai 0.2.0.dev635`; verified `PRAGMA version` reports `v1.5.5-vane.77d1083664` / SourceID `c0f4e4dbfe`.
- Focused tests: 515 passed, 54 skipped. Affected Ray tests: 3 passed, 2 skipped. These skips require the optional native video provider artifact, which was unavailable locally.
- Required release suite: 840 non-Ray tests and 14 Ray tests passed.

The passing suites do not cover the optimizer rewrite exercised by the reproducer.

external/duckdb/src/function/cast/map_cast.cpp
kaka11chen: Fixed in `62fe3cfd235fa6ab875e028b952a704978309e94`. Column pruning now keeps mode-bearing IMAGE casts and engine formatting casts above the field extraction, preserving the original bound validation while still allowing extraction pushdown. This protects both projection reconstruction and table-scan extraction.

The 12 new default-optimizer regressions cover both UNNEST and stored inputs: colliding MAP keys fail before CTAS/INSERT writes them; matching fixed IMAGE casts succeed; mismatches and TRY_CAST retain their error/NULL behavior. Both Python/native video backends also pass the new real-Ray UNNEST/fixed-IMAGE tests. Affected C++/Python/video tests and the required release suite passed; full counts are in the PR validation section. The branch is now based on main with #766 merged. Please re-review the updated head.

kaka11chen:
@codex review

Please review the full current PR diff at `62fe3cfd235fa6ab875e028b952a704978309e94` against main. This is round 1 of two sequential reviews after fixing the P1 loss of explicit IMAGE cast semantics during column pruning. Two clean local static reviews, the noneditable incremental Release build, 12 new default-optimizer regressions, affected C++/Python/native/Ray tests, and the release suite have passed.

kaka11chen:
@codex review

Please perform a second full review of the unchanged head `62fe3cfd235fa6ab875e028b952a704978309e94`. The first requested review after the column-pruning fix completed without findings: https://github.com/AstroVela/vane/pull/767#issuecomment-5556691994 . No code changes have been made since that review.

hubgeter (approved):
Focused re-review of `62fe3cfd235fa6ab875e028b952a704978309e94`, compared with the previously reviewed `77d10836648ec724377db27b21054a72b173746c`.

No new blocking findings. The previous P1 is addressed in the code: `TryGetCastChild` now leaves validation-bearing IMAGE casts and internal formatting casts intact. The visitor still reaches and pushes down the child extraction, while both projection reconstruction and table-scan extraction avoid replacing the original bound cast with an ordinary cast.

The 12 added default-optimizer cases cover UNNEST and stored inputs, duplicate MAP keys before CTAS/INSERT, matching and mismatching IMAGE layouts, and TRY_CAST NULL behavior. The additional Ray test covers fixed IMAGE casts over unnested video frames for both backends. The delta from the prior review is limited to this fix, its tests, and documentation.

Validation boundaries:

- Ran the 12 new regression cases against the verified old-head wheel (`v1.5.5-vane.77d1083664`, SourceID `c0f4e4dbfe`) as a negative control: 6 expected failures and 6 passes. The failures exercise the reported defects; these are not results from the new head.
- `git diff --check` passed.
- The new-head CI native build is still running and has not published a wheel. I have not independently executed the new-head native or Ray tests in this quick re-review.

Approving the reviewed code changes. Required CI still needs to complete successfully before merging.

PR #876, opened 2026-09-21
[feat](ai) Add video embeddings and Cosmos text-video retrieval

kaka11chen:
## Summary

Add `vane.ai.embed_video`, the Relation method, and SQL `ai_embed_video` for one vector per ordered clip. Inputs reuse decoded IMAGE frame records with original indices/timestamps and preserve the existing Arrow/Ray execution path.

Introduce provider-neutral video contracts and a Transformers adapter for Cosmos-Embed1-224p, with paired 256-dimensional text/video encoders. Require eight RGB frames, explicit CUDA precision and revision-pinned custom code; reject unsupported inputs/configuration without fallback.

Handle all-NULL clip batches in shared Arrow transport by preserving a typed empty child slice. Python and SQL calls now return NULL vectors without initializing a model, including NULLs from `video_frames(..., on_error => 'null')`. Empty clips remain distinct from NULL, and subsequent populated batches retain their vectors.

Clip field lookup follows DuckDB's case-insensitive naming in both the native binder and Arrow reader, including reordered fields; ambiguous Arrow names are rejected. The SQL prompt helper separately declares its supported `text`, `blob`, `blob_list`, `file`, and `file_list` input kinds.

## Related issue

N/A

## Documentation impact

The standalone user guide is omitted at the maintainer’s request. Required optional dependency/model/media license notices remain in `THIRD_PARTY.md`.

## Validation

- Completed two consecutive clean self-review passes before an incremental native Release rebuild and non-editable installation. Verified that the installed native binary matches the rebuilt binary and that the installed Python modules match this checkout.
- Added regression coverage for uppercase, mixed-case and reordered clip fields through Python/SQL and local/default Ray execution, sliced/chunked Arrow inputs, ambiguous Arrow field names, and the SQL prompt helper's five supported input kinds. Existing tests also cover all-NULL batches, decode-error NULLs, and populated batches following NULL-only batches.
- Affected installed-package tests: **766 passed** (750 non-Ray and 16 real-Ray). Installed-package typing checks for both AI prompt and embedding APIs passed.
- `scripts/run_release_tests.sh`: **1,776 passed** in the required isolated processes (1,722 non-Ray, 52 shared-Ray, 2 test-owned-Ray); 9 externally provisioned service tests were excluded.
- Configured package mypy hook, Ruff/root formatting, `git diff --check`, and the source/native license checks passed.
- Earlier real-model validation of the feature: **2 passed** with actual Cosmos weights, a decoded javelin video, eight sampled frames and six text candidates. RTX 2080 Ti 22 GB; default Ray; FP16 and FP32; CUDA PyTorch 2.7.0+cu126, Transformers 4.51.3, Ray 2.58.0. Finite normalized 256-dimensional vectors and the correct top-ranked caption were verified offline. CUDA model inference was not rerun for the field-lookup and prompt-typing fixes.

Dependency limitation: the Cosmos extra pins the validated Transformers 4.51.3 release. An additional isolated probe with Transformers 5.17.0 failed while importing the official model code because `find_pruneable_heads_and_indices` was removed from `transformers.pytorch_utils`. No compatibility shim or fallback was added.

Real-model validation used a single GPU node. Multi-host execution requires dependencies and model/source assets accessible on the selected workers, before executing queries. No external paid model endpoints were called.

## Checklist


kaka11chen:
@codex review

QuakeWang (approved):
+1

